PR #182, opened 2026-04-20
feat: add jbct-parser JMH benchmark module

siy:
## Summary

New `jbct/jbct-parser-benchmark` module that began as a JMH harness for `Java25Parser` and grew into a full perf-exploration workspace. The findings from this module directly informed the java-peglib 0.2.2 performance rework adopted in PR #183.

### Contents

- **`Java25ParserBenchmark.java`** — JMH benchmark with **6 `@Benchmark` methods**:
  - `parseCst` / `parseAst` / `parseWithDiagnostics` — run against the stock (dependency) `Java25Parser`
  - `parseCstExp` / `parseAstExp` / `parseWithDiagnosticsExp` — run against the local experimental `Java25ParserExp` for A/B comparison
- **`Java25ParserExp.java`** — experimental hand-edited copy of the generated parser (~26k LOC, single-package rename). Seven focused optimizations applied over several iterations, each profile-guided:
  - `trackFailure` early-exit (skip `SourceLocation` alloc when position is behind furthest-seen failure)
  - `matchLiteralCst` message-string hoisting
  - `matchLiteralCst` match-loop split by `caseInsensitive` (JIT specialization)
  - `matchLiteralCst` bulk-advance on no-newline text
  - Per-instance failure caches for `matchLiteralCst` and `matchCharClassCst`
  - `skipWhitespace` fast-path (shared empty list when no trivia starter at current char)
  - Success-path `endLoc` reuse (one `SourceLocation` allocation instead of two)
  - Measured end-to-end win: ~1.9× over the pre-rework generator output (offline, under JMH)
- **`QuickTime.java`** — simple timing harness used during iteration (fewer warmups than JMH, faster feedback — explicitly NOT a replacement for JMH)
- **`FactoryClassGenerator.java.txt`** — 1,900-LOC real-world Java fixture (copied into resources so results stay comparable across commits)
- **`java25.peg`** — reference copy of the grammar used to generate both the stock parser and the corresponding Exp prototype
- **`run-benchmark.sh`** + **`README.md`** — launcher + usage docs
- **`pom.xml`** — excluded from `deploy` / `install` / `gpg`, and skips `central-publishing-maven-plugin` so the module builds locally but is never published to Maven Central

### Downstream impact

The optimizations validated here were written up as a detailed spec for the java-peglib maintainer, who implemented them as configurable generator flags in peglib 0.2.2. Adopting that release (PR #183) lands the same wins — plus a structural `choiceDispatch` optimization that only makes sense at the generator level — into the production `Java25Parser`. JMH on the new default config shows **~22× end-to-end speedup** on the same fixture vs the pre-rework baseline.

## JMH config

- JMH 1.37
- Modes: `Throughput` + `AverageTime`, unit ms
- Warmup: 3 iterations × 2 s
- Measurement: 5 iterations × 2 s
- Forks: 2
- Fresh parser instance per invocation (matches real usage — parsers are single-use)

## Test plan

- [x] `mvn -pl jbct/jbct-parser-benchmark -am clean package -DskipTests` succeeds
- [x] Shaded `benchmarks.jar` produced (~4 MB)
- [x] `java --enable-preview -jar benchmarks.jar -l` lists all 6 benchmark methods
- [x] Full measurement run: `./jbct/jbct-parser-benchmark/run-benchmark.sh` — all 6 benchmarks complete and produce sensible numbers
- [x] `central-publishing-maven-plugin` configured with `<skipPublishing>true</skipPublishing>` (module won't push to Maven Central on release)


siy:
Not required for production — investigation branch

PR #187, opened 2026-04-23
feat: stream-addressing foundation types for #165

siy:
## Summary

PR0 foundation for #165 — three-component stream addressing wired end-to-end behind a feature flag (default disabled).

Depends-on: #186 (design spec).

## What's in

**Value types and validation (`aether/slice-api`):**
- `StreamAddress`, `StreamVersion` — parsed/validated three-component addressing
- `StreamVersionSpec` — sealed `Exact | Latest` for the `resources.toml` boundary
- `StreamResource` — sealed `Owned | External` for declared resources

**Namespace derivation (`aether/slice`):**
- `BlueprintNamespace.deriveNamespace(groupId, artifactId)` → `groupId + "." + artifactId`
- Blueprint identity comes from Maven classifier `blueprint`, not an artifactId suffix (spec updated in #186)

**Resources.toml parsing:**
- `StreamConfigParser.parseResources(toml)` returns `Map<String, StreamResource>`
- `version` and `source` fields are mutually exclusive; absence = parse error

**Registry and lifecycle:**
- `StreamRegistryEntry` — captures address, retention, registered-at, registered-by, refcount
- `StreamRegistry` interface + `InMemoryStreamRegistry` — register/lookup/resolve/acquire/release
- `latest` resolution picks the highest-semver entry per `(namespace, stream)`
- Reference counting: increment on acquire, decrement on release, removal at zero

**System streams:**
- `SystemStreams.CLUSTER_EVENTS` constant for `system:cluster-events:1.0.0`
- `SystemStreamBootstrap.bootstrap()` registers all `SystemStreams.ALL` idempotently with framework reference

**Feature flag and facade:**
- `StreamNamespacesConfig` — `DISABLED | ENABLED`, default disabled
- `StreamNamespacesService` — flag-gated facade: when disabled, snapshot is empty and lookups return `NOT_FOUND`

**HTTP routes:**
- `GET /api/stream-namespaces` — snapshot of registered streams
- `GET /api/stream-namespaces/{namespace}/{stream}/{version}` — exact lookup
- Both registered via `StreamNamespacesRoutes` and wired into `ManagementServer`
- `ManageableNode.streamNamespacesService()` — default returns disabled service

**Circular-dep prevention:**
- `StreamLifecycleEventPolicy.shouldEmit(address)` — `false` for `system:*`, `true` otherwise
- Prevents the future `STREAM_REGISTERED`/`STREAM_DELETED` events from looping back into `system:cluster-events`

**Spec consistency:**
[...]

Issue #215, opened 2026-05-10
Stream metadata replication via consensus (StreamConfigKey)


siy:
## Problem

`StreamPartitionManager.streamInfo()` and `listStreams()` read from a per-instance `ConcurrentHashMap<String, StreamEntry> streams` populated only when `createStream(StreamConfig)` runs locally. There is no consensus/KV write, no notification handler, and no cluster-uniform registry. `GET /api/streams/<name>` therefore returns 404 (empty body via `_api_call`) on any node that did not locally execute `createStream`.

Discovered during RC1 cloud validation (2026-05-10c session). The 04-streaming `test_read_from_non_governor_node` fails on cloud-jvm and cloud-container because the assertion correctly identifies that non-governor nodes have no metadata. Docker-remote masks this because (a) host port-range maps `5151`→a different container, AND (b) the data-path `replicationManager.replicateEvent` lazily seeds a `StreamEntry` on receivers — both accidents of layout, not design.

## Proposed solution

Add `StreamConfigKey(String streamName) → StreamConfigValue(StreamConfig config)` to `AetherKey.java`. Write on `createStream` (governor path) via the consensus batch. Have every node hydrate a metadata-only view of `streams` on the corresponding `ValuePut` notification.

`streamInfo` / `listStreams` then become consistent across nodes regardless of where partition data physically lives. Read-routing for partition data (`StreamReadRouter`) can stay governor-pinned.

## Acceptance

- [ ] `StreamConfigKey` / `StreamConfigValue` added to `AetherKey.java` / `AetherValue.java` — same shape as existing key types.
- [ ] `StreamPartitionManager.createStream` writes the key via `clusterNode.apply` instead of (or in addition to) the local map insert.
- [ ] A `KVStoreNotification.ValuePut<StreamConfigKey, StreamConfigValue>` handler hydrates the local `streams` map on every node.
- [ ] `04-streaming/test-stream-replication.sh:test_read_from_non_governor_node` passes on `--env cloud` once the test-side endpoint fix in this branch is in place AND the product change lands.

## Files

- `aether/aether-stream/src/main/java/org/pragmatica/aether/stream/StreamPartitionManager.java`
- `aether/node/src/main/java/org/pragmatica/aether/api/routes/StreamRoutes.java`
- `aether/slice/src/main/java/org/pragmatica/aether/slice/kvstore/AetherKey.java`
[...]

Issue #212, opened 2026-05-06
SSE + WebSocket tail subscription for /api/streams/{ns}/{stream}/{version}/tail (deferred from PR #187)


siy:
## Context

Event-stream-namespaces spec §16 (RC1 acceptance checklist) requires "SSE + WebSocket tail subscription protocols" for the `/api/streams/{ns}/{stream}/{version}/tail` route. PR #187 ships polling-based tail (CLI polls `GET /api/streams/.../{version}` with `?fromOffset=N&maxEvents=K`), with the SSE/WebSocket implementation deferred to this ticket.

## Why deferred

The current management-API route framework has no streaming primitives. Every route terminator goes through `.asJson()` / `.asText()` — single-shot responses. There is no chunked-transfer-encoding support, no long-lived-connection mechanism, no fan-out infrastructure for SSE/WebSocket beyond the existing `EventWebSocketPublisher` (which broadcasts cluster events, not stream tail).

Implementing SSE + WebSocket tail in PR #187 was scoped as a Wave 4A sub-task but blocked on this missing infrastructure. The pragmatic call: ship polling for RC1; defer streaming to RC2 with proper infrastructure.

## Scope when picked up

- **Streaming primitives in the route framework** — chunked encoding, response-writer streaming, keep-alive, backpressure.
- **SSE handler** for `Accept: text/event-stream` clients on `/api/streams/{ns}/{stream}/{version}/tail`.
- **WebSocket handler** for `Upgrade: websocket` clients on the same route. Reuse `EventWebSocketPublisher` patterns or extend.
- **CLI `aether stream tail`** preference: prefer SSE when server supports it; fall back to polling on legacy servers.
- **Frozen-at-connect semantics** per spec §9.3: version is resolved at connect; reconnect picks up newer version.
- Spec §16 acceptance checklist amended back to "SSE + WebSocket tail subscription protocols" once landed.

## Marker

`rc2` — RC1 ships polling-based tail; streaming protocols are post-RC1.

## References

- Spec: `aether/docs/specs/event-stream-namespaces-spec.md` §9.3 (frozen-at-connect), §12.4 (protocol selection), §16 (acceptance checklist)
- PR #187 (event-stream-namespaces implementation)
- Stub site: `aether/node/src/main/java/org/pragmatica/aether/api/routes/StreamApiRoutes.java` `tailNotImplemented` (will be replaced by the polling impl in PR #187)

Issue #210, opened 2026-05-05
12-network on cloud: SWIM NODE_LEFT/NODE_FAILED events not detected after kill_node


siy:
## Symptom

After `kill_node` of a non-leader on Hetzner Cloud:
- aether-node container is killed (verified via SSH).
- Cluster auto-heals — 02-chaos and 03-scaling pass on cloud, confirming SWIM internal detection works.
- BUT `wait_for_node_departure` polling /api/events on each node times out: no NODE_LEFT or NODE_FAILED events for the killed node ID, even with TIMEOUT_SCALE=3 (180s).

Reproducible with:
```
./run-tests.sh --env cloud --suites 12 --skip-build
```

12-network: 1 passed (gossip-encryption), 2 failed (test-quic-connectivity, test-swim-detection).

## What's been ruled out

- **Test fixture / port-hop**: `topology_events_since` was Docker-only (port-hopping `${TARGET_HOST}`). Fixed in `ebfe4409f` to enumerate per-VM IPs on cloud. Verified curl `/api/events` returns events from each node.
- **Restart policy**: containers run with `--restart unless-stopped`, so `docker kill` was being auto-restarted before SWIM detection. Fixed in `584d89748` to `docker update --restart=no` before kill.
- **Timeout**: `wait_for_node_departure` now scales 60s → 180s on cloud (`d71f2469e`). Failure occurs after the full 180s wait — the events simply never appear.
- **Event matching**: regex correctly handles `[^}]*` spanning into the `details` object containing `nodeId`. Verified against actual event JSON shape.

## Hypotheses to investigate

1. **NODE_LEFT/NODE_FAILED never get emitted on cloud** — events on the leader stream show NODE_JOINED, CONNECTION_ESTABLISHED, LEADER_ELECTED, DEPLOYMENT_STARTED. Cluster recovers from kill (verified by 02-chaos passing) but the corresponding NODE_LEFT/NODE_FAILED is either never persisted to the events buffer or persisted with a different type name on cloud.
2. **Per-node events are sparse** — handover note says "each node maintains its own per-node event buffer (notifications are local)". Maybe only certain nodes emit failure events, and the test polls all 5 but the actual emitter has a different ID format.
3. **Restart-policy update fails silently** — `docker update --restart=no aether-node` via `cloud_ssh` may fail due to SSH context or permission issues, and the kill still gets auto-restarted before SWIM detects.

## Reproduction harness

After bootstrap of cluster B on cloud:
```
[...]

PR #239, opened 2026-06-05
feat: stream + pub/sub production-readiness — shared resource naming, active replication, replicated cluster-events (supersedes #187)

siy:
## Summary

Brings streams AND pub/sub to production-readiness on rc1 under a **single shared resource-naming scheme**, with active replication for all streams and a replicated cluster-event stream. Supersedes #187.

## 1. Shared resource naming (`ResourceAddress`)

One naming model — `namespace:name:version` — reused identically by streams and pub/sub:
- **`ResourceAddress(Namespace, ResourceName, ResourceVersion)`** — each component a validated value object (parse-don't-validate; factories return `Result`); validation routed through core `Verify`. `system:*` reserved.
- Replaced the earlier per-domain `StreamAddress`/`TopicAddress`/`*Version` wrappers — the stream-vs-topic distinction lives in the **key types** (`StreamRegistryKey` vs `TopicSubscriptionKey`), not the address. No duplicated grammar; `asString()`/wire form preserved.

## 2. Pub/sub namespaces (identical model)

- `TopicConfig`/`TopicSubscriptionKey`/`TopicSubscriptionRegistry`/`SliceTopology` carry a `ResourceAddress`; topics declared via `@Notify` resolve to `namespace:topic:version` (namespace derived from blueprint coords, version defaulted) with **back-compat** for bare names.
- `PubSubValidator` rejects `system:*` for app topics. Topology graph keys/matches on the canonical address. Topics have no HTTP/CLI surface today — none invented (documented).

## 3. Stream reliability foundation (active RF for all streams)

- **HRW (rendezvous) placement** + a per-node **`ReplicaSetController`** that reconciles the replica set into the previously-inert `ReplicaRegistry` on every `MembershipDecision` — deterministic, no coordinator, automatic owner-failover.
- **Catch-up backfill** (new replicas reach `CAUGHT_UP` before serving reads), **activated app-stream replication** + owner-routed publish/write-forwarding.

## 4. Replicated cluster-events (#210) + durable metadata (#215)

- Sealed `ClusterEvent` `@Codec`; `system:cluster-events` moved off the node-local buffer onto the **replicated off-heap partition transport** (RF=`max(3,N−2)`), **owner-gated emit**, **production retention** (count+bytes+age, config-overridable, byte OOM cap).
- Durable cross-node metadata: `StreamConfigKey` (flat config) + `StreamRegistryKey` (namespaced refcount); 405 gate on `system:*` writes; **C1** symmetric snapshot/restore round-trip.

## Verification

- **Full reactor compiles (134 modules)**; touched-module unit tests green (slice-api 302, slice 600, node 532, aether-stream 372/383, deployment 450, cli 470, …) + new tests for placement, controller, catch-up/durability, codec, retention, `ResourceAddress`/`Namespace`/`ResourceName`/`ResourceVersion`, pub/sub namespaces.
- A JBCT/correctness review of the replication subsystem found and **fixed real durability bugs** (false-ready CAUGHT_UP, partial-apply loss, unstable partitioning, owner-gate dup/drop).

## Known follow-ups

- App primary-publisher ownership is governor-based while the replica *set* is HRW-maintained (reliable; unifying is a follow-up).
- `STREAM_REGISTERED/DELETED` lifecycle emission deferred (RC2); SSE/WS tail (#212); RBAC (#205); read-forwarding (#136/#137/#135).
- **04-streaming integration suite** (non-governor reads, kill→event-cluster-wide, RF-heals, snapshot round-trip) must run in CI/Docker.


siy (commented):
## Review — stream namespaces + cluster-events-over-stream (#239)

Reviewed the full diff (5 commits, +8251/−1922, 93 files): the cluster-events hard-replace core, KV grafting, the 405 gate, all read routes, the FSM refcount path, and JBCT sampling of the largest new files. Base is rc1 HEAD `37e4e257b`, so it cleanly carries the latest membership/auto-heal/transport work.

**Verdict: right feature, built the right way — but not merge-ready yet. Two blockers (C1, C3) + one behavioral change to document.**

### Strengths (verified, not taken on faith)
- **Retention buffer correct.** `ClusterEventStreamBuffer` is genuinely bounded (drop-oldest), `synchronized`, and the read window (`FETCH_BATCH = MAX_RETAINED_EVENTS = 10_000`, `ClusterEventAggregator.java:99-101`) covers the full retained suffix — the "no truncation" claim holds. Closes #187's never-enforced-retention gap.
- **405 system-write gate is exactly as described** — `ManagementServer.java:679` runs `rejectSystemStreamWrite` **before** `validateManagementSecurity` (`:686`), role-independent, covering URL-encoded `system%3A`.
- **Refcount-tied-to-slice-ACTIVE is correctly wired** (`NodeDeploymentState`): ACQUIRE consensus-batched onto the `SliceNodeValue` ACTIVE put, RELEASE on DEACTIVATING/UNLOADING/force-cleanup, via the command-builder path (not the racy `acquireReference` wrappers) — correct consensus ordering.
- **Emit-site migration is complete** — zero live-code references to the deleted `ClusterEventLogPublisher`/`Sweeper`/`AggregatorConfig`/`ClusterEventValue`; all read routes re-sourced to the new aggregator.
- No envelope-version bump needed (`jbct/**` untouched — invariant #3 correctly N/A).

### 🔴 C1 (BLOCKER) — KVStoreSerializer round-trip is asymmetric → snapshot/backup restore fails
`StreamRegistryKey`/`BlueprintStreamBindingsKey` have **serialize** arms (`KVStoreSerializer.java:171-172, 240-256`) but **no matching parse arm** in `parseKeyValue` (the section switch ~`:401-435` has no `case "stream-registry"` / `case "blueprint-stream-bindings"`), so they fall through to `UnknownKeyType`. These keys are persisted via `cluster.apply(...)` (`KvBackedStreamRegistry`), so any TOML snapshot/backup containing them **cannot be restored**. Contrast the pre-existing `StreamConfigKey`, which has both arms (serialize `:154/:223` + parse `:430`). Live Rabia replication uses a separate reflection codec and is unaffected — the break is the snapshot/DR path. **Fix:** add the two parse arms + a round-trip test mirroring `KVStoreSerializerTest`.

### 🔴 C3 (BLOCKER) — Triad invariant #1: zero doc updates
13 new `/api/streams/*` + 2 `/api/stream-namespaces/*` routes and a full `aether stream` CLI were added with **no** changes to `management-api.md`, `cli.md`, `feature-catalog.md`, or `CHANGELOG.md`. Per the project triad rule the feature is "incomplete" until those land.

### ⚠️ Behavioral change to document loudly — observability goes node-local
`/api/events`, `/api/alerts`, `/api/traces` now serve **per-node** history (the buffer is node-local; both SPI halves close over one buffer). I traced the consumers: this is **observability-only — no control path (deploy/membership/drain/alert-dedup) reads cluster events**, so there is no correctness regression. Two notes:
1. `AlertManager.activeAlertsAsList()` previously did a cross-node UNION + `alertId` dedup for injected alerts; now node-local, so an alert/trace injected on node B is invisible on node A. Broader than "events" — call it out in CHANGELOG/release notes (the #215 deferral).
2. **Membership events** (`NODE_FAILED`/`NODE_LEFT`/`NODE_JOINED`) are emitted **per-node** (`TopologyObserver NodeRemoved → ClusterEventAggregator`, `AetherNode.java:490`), so basic detection still works on any backend. **But** the integration harness's `topology.sh` helpers were written for the replicated model (they note the LB "picks a random backend… only sees one" and reference "the replicated commit") — order/`clusterSize`-sequence and injected-alert/trace assertions need a harness pass. **Recommend running the full integration suite against this branch before merge** — a green 134-module compile won't catch this.

### Cleanups (non-blocking)
- **C2** — `AetherKey.java:1322-1324` comment claims `ClusterEventLogKey` is "added alongside the retained … so rc1's path is untouched," but the diff *renames/replaces* it (`-ClusterEventLogKey` → `+StreamRegistryKey`); the key no longer exists. Misleading comment — fix to avoid future confusion (no wire-ordinal risk; the switch is sealed-exhaustive).
- **C4** — `StreamLifecycleEventPolicy.shouldEmit` is referenced only in comments (`ClusterEventAggregator.java:81`, `AetherNode.java:1272-1273`), never called; `ClusterEvent.StreamRegistered`/`StreamDeleted` are never emitted. Staging these is fine, but `AetherNode.java:1272-1273` asserts the self-loop "is prevented by StreamLifecycleEventPolicy.shouldEmit" — false (nothing emits them). Tighten the comment or wire the guard.
- **C5** — `StatusRoutes.java:109-116` silently remaps the legacy `?sinceSeq=` cursor as epoch-milli into the new `Instant`-based `eventsSince`; existing callers get different behavior. Belongs in CHANGELOG.
- **C6 (JBCT)** — try/catch where the idiom is `Result` (`StreamVersion.java:63-69` `NumberFormatException`; `StreamTailPoller.java:169-170` `Thread.sleep`), and pervasive FQCN-in-method-body (`AetherNode.java:1265-1268`, `ClusterEventAggregator.java:198`, `StatusRoutes.java:115`, `AlertRoutes.java:48-49,102`, `ManagementServer.java:1127,1131`). Confirm `jbct:check` passes; sweep these.

### Bottom line
Architecture is sound and it doesn't disturb the membership/auto-heal substrate. Fix **C1** (+ round-trip test) and land the **C3** docs/CHANGELOG (including the node-local observability change), correct the C2/C4 comments, address the C6 JBCT items, and run the full integration suite — then it's merge-ready.

siy:
## Integration validation findings — full suite vs `da91b059a`

Running the full 15-suite against this branch on the remote cluster (build green, 67 modules; rc1 fixes + harness intact). Cluster A + the streaming analysis are complete below; **cluster B (membership / `NODE_FAILED`-over-replicated-events) is still running — follow-up comment to come.** Issues are split by lane: 🐞 = real-bug candidate (code), 🧪 = stale test / test-infra (validation lane).

Context that shaped the triage: pub/sub streams are **not** namespaced (plain names + flat API are correct for them); slice `@Notify`/`resources.toml` streams **are** namespaced and need the full `(namespace, stream, version)` address externally (spec §3.1/§4.1/§9).

---

### 🐞 Issue 1 — Sustained pub/sub publish fails 100% (owner-route/auto-create race under load)
- **Test:** `04-streaming/test-stream-under-load.sh::test_sustained_stream_publish` (`load-test-stream`, 10 RPS × 30s).
- **Observed:** `total=174, success=0, failure=174` (100% error); `GET /api/streams/load-test-stream` → `500 "Stream not found"`.
- **Why it's a real bug, not staleness:** `load-test-stream` is a plain **pub/sub** stream on the flat API — which is correct for pub/sub. Small/slow pub/sub auto-creates fine (`integration-pubsub` 25/25; `test-events` single + batch-of-50 pass). Only **sustained-rate publish to a brand-new auto-creating stream fails completely and never recovers**.
- **Hypothesis:** race in the new owner-routed/replicated publish + auto-create (A6 / B5b) — rapid publishes arrive before HRW owner-placement/registration settles; the stream never registers, so every publish 500s and reads 404.
- **Contract:** breaks 04 charter C4 (≥95% strict-2xx under sustained load).
- **Action:** exercise the owner-routed publish path with concurrent first-publishes to a new stream; confirm auto-create + owner-placement is atomic/awaited before serving.

### 🐞 Issue 2 — Cross-node replicated event delivery gap
- **Test:** `11-observability/All_nodes_agree_on_order` — "Node 0 returned no marker-bearing events".
- **Observed:** a published marker event is absent from node 0's `/api/events` view.
- **Significance:** this is the replicated `cluster-events` path (post `node-local→replicated`, B5b) — **the same path membership `NODE_FAILED` rides.** If events published on one node don't reach all replicas, cross-node observability (and possibly membership-event detection) regresses. Cluster B will confirm whether membership detection is affected (linked to this).
- **Action:** verify replicated cluster-events delivery reaches every replica (catch-up + steady-state), not just the owner/emitter.

### 🐞 Issue 3 — Rolling-promote deployment quiesce timeout
- **Test:** `06-deployment/Rolling_promote` — "did not quiesce within 30s".
- **Context:** #239 modifies `NodeDeploymentState`'s activation chain (refcount-tied-to-ACTIVE ACQUIRE/RELEASE, blueprint stream-bindings at deploy). `06-deployment` was green on rc1.
- **Action:** check whether the consensus-batched refcount writes / stream-binding writes add latency to the rolling-promote quiesce path.

### 🐞 Issue 4 (low confidence — possibly transient) — No-leader 503 on a leader-bound route
- **Test:** `07-cluster-mgmt/Config_visible_on_all_nodes` — node-3 probe got `503 "No leader elected for leader-bound management route"` on `GET /api/config/overrides`; cluster was healthy seconds later.
- **Context:** cluster A was 10/10 green on rc1. Possibly a wider no-leader window during formation from #239's boot-time replication wiring (`ReplicaSetController` on membership change), or just variance.
- **Action:** low priority; investigate only if leader-bound 503s recur across runs.

---

### 🧪 Issue 5 — `@Notify` namespaced stream accessed via the flat pub/sub API (stale test)
- **Test:** `08-resources/test-streaming-resources.sh` (`notifications`).
- **Observed:** publish 0/20; `GET /api/streams/notifications` → `500 "Stream not found"`.
- **Root:** `notifications` is a slice `@Notify` resource = **namespaced** (slice's Maven-derived namespace), but the test publishes/reads it via the plain pub/sub flat API. Per the spec, namespaced streams require the full external address; plain names are slice-internal only. **Not a code bug** — the test predates namespacing.
- **Action (validation lane):** migrate the test to drive the slice SDK with the internal plain name and read back via the full external address.

### 🧪 Issue 6 — Stream harness helpers may target a removed/legacy API shape
- **Files:** `aether/tests/integration/lib/cluster.sh:2291-2303` (`stream_publish`/`stream_info`/`stream_list`) use flat single-name paths (`/api/streams/publish/{name}`, `/api/streams/{name}`). `StreamApiRoutes` doc (lines 38-40) states the legacy flat `?streamName=` query routes were removed Wave 4A; path-based addressing exclusively.
- **Open question for the author:** is the flat **path-segment** form `/api/streams/publish/{name}` still the intended **pub/sub** external API (it works for pub/sub today), or should all external access be full-address-only? Confirming the intended external contract decides how much of the harness/tests I migrate.
- **Action (validation lane, pending the contract answer):** align the helpers + namespaced tests to the path-based addressing.

---

**Net so far:** the namespace rework itself looks faithful to the design — several "failures" are stale tests against the new addressing (my lane). The genuine code-regression candidates are narrower: **owner-routed pub/sub publish under sustained load (Issue 1)**, **cross-node replicated-event delivery (Issue 2)**, and **deployment-quiesce timing (Issue 3)**. Cluster-B results (membership `NODE_FAILED` detection over the replicated event path — directly linked to Issue 2) to follow.

siy:
## Integration validation — cluster B (follow-up) — vs `da91b059a`

🚨 **RC-BLOCKER found on cluster B. This supersedes the streaming issues above in priority.**

### 🐞🔴 Issue 7 (BLOCKER) — Consensus leader-election wedges after a leader kill (no re-election, ever)
- **Tests (all in `02-chaos`):** `Kill_leader_and_re-elect` ("no new leader observed within 150s"; "leader changed … timed out after 300s"), `Cluster_has_quorum` ("expected ≥4, got 0"), `Health_with_4_nodes` ("got ''"), `Auto-heal_restores_to_5` ("current=0", timed out 360s).
- **Observed:** kill the current leader → **no new leader is ever elected.** Survivors continuously submit leader proposals with a **monotonically climbing epoch** and **never commit**:
  ```
  LeaderElectionState.sendProposal: candidate=<replacement>, viewSequence=57, epoch=57
  ... viewSequence=58 ... 101 ... 102 ... 103   (≈9s/round, climbing indefinitely, ZERO LeaderElected/commit)
  ```
  Result: permanent no-leader → every leader-bound route (`/api/events`, `/api/cluster/topology`, `/api/cluster/generation`) returns `503 "No leader elected"` → no CTM → cluster cannot recover (`current=0`).
- **Conditions:** 4 alive, quorum-capable nodes (3 seeds + 1 replacement); **`0 "Network is unreachable"` across all nodes — NOT an env partition.**
- **Attribution: #239 regression.** `02-chaos/Kill_leader_and_re-elect` **passes on rc1** (re-election is near-instant there). So something in #239's additions is preventing the Rabia consensus layer from **committing** a leader proposal after a leader change — proposals are *submitted* (election logic runs) but never reach consensus. Prime suspects given the diff: the new owner-routed/replicated emit path or stream-replication wiring contending with / blocking the consensus apply path, or a consensus-membership/quorum miscount after the membership change. (`aether-deployment`/`ntt`/`LeaderReconciler`/`MembershipFsm` are untouched by #239, so this is on the consensus/event-emit/replication side.)
- **Impact:** RC-blocking. Also **cascade-contaminates the rest of cluster B** (03/05/12/13 run against a leaderless, dead cluster), so their results in this run are not trustworthy.
- **Action:** reproduce in isolation (`02-chaos` kill-leader against this branch), trace why submitted leader proposals never commit under Rabia with the new replication active. This is the must-fix before any further cluster-B validation is meaningful.

### 🐞 Issue 8 — `/api/events` (+ topology/generation) is now leader-bound → 503 during any leader churn
- **Observed:** `GET /api/events` on a follower returns `503 "No leader elected for leader-bound management route"` (and the same for topology/generation) whenever there is no leader.
- **Regression vs rc1:** on rc1 `/api/events` was readable on any node (node-local ring / replicated replay). Routing cluster-event **reads** through the leader means **event/membership observability is unavailable during exactly the windows you most need it** (leader churn, failure detection). Compounds Issue 7 and is relevant to Issue 2 (the membership `NODE_FAILED` path now can't be read mid-re-election).
- **Action:** confirm whether leader-binding `/api/events` is intended; for a replicated stream, reads should be servable from any in-sync replica, not gated on the leader.

---

**Net:** the streaming-API failures (Issues 1, 5, 6) are mostly stale tests against the new addressing, as discussed — but **cluster B surfaced a genuine, severe consensus regression (Issue 7)** that gates everything else. I'd treat Issue 7 as the top priority; until it's fixed, cluster-B destructive validation can't produce a clean tally. Happy to re-run the full suite (and the isolated `02-chaos` repro) once it's addressed.

Issue #277, opened 2026-06-11
Interceptors: runtime aspect switching not wired (Aspect seam always identity; frozen at processing-time)


siy:
## Finding R22 (resource-surface assessment 2026-06-11)

The aspect-based observability design (every injection point runtime-switchable into metrics/logging/tracing combinations) has its **seam structurally present** but the **runtime-switching mechanism is not wired** at HEAD:

1. `SliceFactory.invokeFactory` always passes `Aspect.identity()` (`aether/slice/.../dependency/SliceFactory.java:143`) — no caller ever supplies a non-identity Aspect.
2. Interceptors are wired at **annotation-processing time** and frozen into an immutable generated wrapper record at slice creation (`FactoryClassGenerator.java:630,656`).
3. No `DynamicAspectKey`/`DynamicAspectRegistry` exists (zero hits outside docs); no aspect entry in `AetherKey`; no management route or KV subscription re-provisions interceptors.
4. The only runtime lever is the slf4j logger threshold for `LoggingMethodInterceptor`.

So changing any interceptor config requires a slice redeploy. The static interceptor path works; the runtime-toggle premise does not exist yet.

## Acceptance
- A mechanism to toggle/reconfigure aspects per injection point at runtime (KV-driven config + management route), honoring the Aspect seam — OR an explicit decision to document aspects as deploy-time-static for RC1 and defer runtime switching.
- If implemented: changing metrics/logging/tracing on a method takes effect without redeploy; zero-cost when all-off preserved.

Design intent per the observability philosophy; this ticket is about realizing the runtime half.

siy:
## ⏸️ Postponed — deeper codegen design required (2026-06-25)

Per-injection-point granularity needs a generated-factory change that warrants a deeper design pass than was feasible in-thread. Pausing here with full state captured.

**Decisions locked:**
1. Granularity = **per-injection-point × facet** (spec §7.2), not per-slice (repo-owner confirmed).
2. **No envelope bump** — GA envelope format = 1000 (overrides project-invariant #3 for GA).
3. **Memory verified negligible** — one holder per injection point is fine (typical ~80/node ≈ 10 KB, high ≈ 67 KB; created once at slice load, no GC churn). No per-slice-array optimization needed.
[...]

Issue #268, opened 2026-06-11
Resource SPI: context-overload provide() bypasses cache/refcount/close → resource leaks + use-after-close + cached failures


siy:
## Findings R1+R2+R3 (resource-surface assessment 2026-06-11)

The two `SpiResourceProvider.provide()` overloads implement incompatible lifecycle models, and every slice-generated call is upgraded to the **context** overload (`CompositeAwareResourceProvider`), which bypasses the lifecycle machinery:

1. **Leak on every unload (R1):** context overload registers the consumer then calls `createResourceWithContext` directly (`SpiResourceProvider.java:90-98`); only the plain path caches (`:83-87`). `releaseAll` closes only cached entries (`:117-129`) → context-provisioned resources (DB pools, SMTP/Netty event loops) are **never closed**.
2. **Cross-slice use-after-close (R1):** only context-path callers register in `consumers` (`:93-95`); last-consumer unload closes a cached resource a plain-path caller still holds.
3. **Wrong factory closes (R1):** `releaseAll` uses `factoryList.getFirst()` (`:122-127`), not the factory whose `supports()` matched (matters for DB async/R2DBC/JDBC sharing `SqlConnector`).
4. **Failed promises cached forever (R2):** plain path `computeIfAbsent` (`:86`) caches a failed promise → an `Intermittent` failure classified for FSM retry-with-backoff returns the same poisoned promise on every retry. The transient/fatal classification only works today because the context path happens to be uncached.
5. **Races (R3):** release/provide check-then-act over `consumers` (`:109-131`); consumer registered before provisioning succeeds (`:93`); partial-provision leak (param 1 provisions uncloseable, param 2 fails).

**Root cause behind resource leaks across DB, @Notify (#TBD), and interceptors.**

## Acceptance
- One lifecycle path: cache + refcount + close cover the context overload; failed promises evicted (not cached).
- Close uses the provisioning factory, not getFirst().
- Tests for releaseAll / refcount / close-on-last-consumer / transient-eviction (currently zero).

Part of the resource-surface assessment; highest-leverage resource fix.

siy:
Re-verified on disk 2026-09-05 (CTO, session 9), while speccing the three-form resource provisioning design. **This ticket's substance holds exactly. Two updates.**

## 1. The line citations have drifted; the mechanism has not

Current code, at `release-1.0.0-rc4` `b16ba5688`:

[...]

PR #356, opened 2026-06-24
feat: #277 PR1 — system observability aspect seam + write-side registry (foundation)

siy:
## #277 PR1 — Foundation: always-on system-observability aspect seam + write-side registry

First of three PRs for #277 (runtime-switchable system observability aspects). Spec: `aether/docs/specs/runtime-observability-aspects-spec.md`.

**This PR is foundation/plumbing only — it changes no runtime behavior.** The system observability aspect is now sourced at the slice seam and is KV-swappable, but `apply()` is a passthrough; the per-call read/emit (the rehome) is PR2.

### What's here (items 1–3)
- **`AspectObservabilityConfig`** (slice-api) — immutable per-injection-point snapshot `(logging, metrics, spans, tracing, depth)` + `OFF` + `allOff()` hot-path check.
- **`ObservabilityAspect<T>`** (slice-api) — holds a `volatile AspectObservabilityConfig`; `apply()` passthrough for now; `swap()` for the registry.
- **Slice source-swap** — `SliceFactory:143` now passes a per-slice system aspect instead of `Aspect.identity()`, threaded node→slice via `SliceCreationContext.observabilityAspect()` (default `identity()` keeps every other context/test untouched) and a `Fn1<Aspect<?>,ArtifactBase>` factory built at `AetherNode` over the registry.
- **Write-side registry** — `ObservabilityConfigKey(artifactBase)` + `ObservabilityConfigValue` (`AetherKey`/`AetherValue`, KV serializer wired in all three switches), and `ObservabilityConfigRegistry`: holds live aspect instances per `artifactBase`, swaps their `volatile` on a KV put via the `KVNotificationRouter` chain (wired identically to `ObservabilityDepthRegistry`), register-on-load / deregister-on-unload (via a release callback in `SliceStore.cleanup`).

### Tests + build
- 16 new unit tests (key round-trip, registry register/swap/deregister/no-leak, serializer round-trip). Full reactor green.

### Design decisions worth a look
1. **Mechanism = G-3 volatile-push** (per-instance volatile, swapped on event), chosen over the simpler per-call registry lookup that the shipped `ObservabilityDepth` uses. This was an explicit call — the volatile model needs the node→slice threading you see here, which `ObservabilityDepth` doesn't have.
2. **Per-slice granularity** (one volatile per slice aspect). True per-*method* facet toggling can't coexist with "one volatile, no per-call lookup" without generated per-method fields (excluded — no codegen change / no envelope bump). The existing per-method `ObservabilityDepth` registry is untouched. **Flag if per-method facet control is required** — it reopens the mechanism choice.
3. **STEP-0 baseline deferred to PR2.** PR1 adds zero per-call cost (passthrough), so there's nothing to baseline; the regression gate belongs with the PR2 rehome that introduces the volatile read. No aether JMH harness exists yet — I'll build it there.
4. **Load/put race closed** in `createAndRegister` — register the instance *before* seeding from `configs`, so a concurrent KV put always converges on the latest snapshot.

### Notable conventions discovered (FYI)
- `Fn1<R,T1>` is **return-first** (`R apply(T1)`).
- `@MessageReceiver` is `@Retention(SOURCE)` / doc-only — dispatch is the `KVNotificationRouter` chain, so I mirrored that wiring.
- Erasure forced the key construct-factory to be named `forArtifactBase(String)` (parse keeps `observabilityConfigKey(String)→Result`).

### ⚠️ Blocker for PR2
**PR2 (resource wrap, item 4) is blocked on #268.** `SpiResourceProvider.provide(type, section, context)` bypasses the cache/refcount path, so wrapping resources can't compose with lifecycle until #268 lands. Per spec Open-Q1 + your brief gotcha #2, flagging before I start item 4.

### Not in this PR
- Hot-path read / rehome of logging+metrics+trace onto the seam → **PR2** (+ resources, blocked on #268).
- Management triad (REST+CLI+docs) → **PR3**. (`setConfig`/`getConfig`/`allConfigs` already exist on the registry, unused, ready for it.)

Per our flow: **please review + merge — I won't self-merge.**


siy:
## 🔴 REQUEST CHANGES — do not merge; per-injection-point rework required

Reviewed — and to be clear up front: this is strong work, and you did **exactly** the right thing flagging the granularity question rather than silently shipping it. The miss is in **my brief**, not your implementation.

## Decision: per-injection-point (spec §7.2), not per-slice
Confirmed with the repo owner. §7.2 mandates **per-injection-point × facet** and explicitly rejects per-slice (*"can't isolate one hot method; floods"*). The incident-microscope ability — *"turn on tracing for `OrderService.placeOrder` only"* — and visibility into the **entire invocation tree** (each method / cross-slice / resource boundary independently switchable) is the whole point. So the per-slice granularity needs to change.

## The no-codegen constraint (brief gotcha #4) is LIFTED
It conflicted with the spec. Per-injection-point × one-volatile-no-lookup requires a **distinct aspect instance per injection point**, which means the **generated factory must emit/capture a per-injection-point aspect at each `apply()` site**. That's a codegen change **+ an envelope bump 1005→1006** (project invariant #3) — both now in scope and required. (The research that fed my brief was simply wrong on "no codegen.")

## Concrete direction
- **`FactoryClassGenerator`** — the `aspect.apply(...)` sites (slice-entry method, cross-slice call, and resource call — the addressable injection points per §4.3) must thread a **per-injection-point aspect instance**, not the single shared param-0 `Aspect<T>`. Each injection point gets its own `ObservabilityAspect` holding its own `volatile AspectObservabilityConfig`.
- The wrapper reads `this.observabilityConfig` (its own instance) — **no per-call lookup** (§6.1). All-off stays a nop: one volatile load + a predicted branch (the owner confirmed perf is a non-issue — a disabled aspect is basically a nop).
- **Registry keys by injection-point**, mirroring the existing per-injection-point `ObservabilityDepthKey` / `ObservabilityDepthRegistry` granularity (§7.3 — that depth registry becomes the depth dimension of the new config).
- Bump `ENVELOPE_FORMAT_VERSION` 1005→1006 in `ManifestGenerator` + add 1006 to `SliceManifest.SUPPORTED_ENVELOPE_VERSIONS` (the generated factory output structure changes).

## Keep most of PR1
`AspectObservabilityConfig`, `ObservabilityAspect` (volatile + `swap`), the registry skeleton + `ObservabilityConfigKey`/value + serializer wiring, register-on-load / deregister-on-unload, and the load/put race fix all carry over. The slice-seam swap at `SliceFactory:143` stays — but the aspect threaded becomes **per-injection-point**, resolved/captured per site in the generated factory. So the delta is: granularity → per-injection-point instances + the codegen to thread per-site aspects + the envelope bump.

## STEP-0 / perf
PR2's rehome introduces the first per-call volatile read, so the all-off micro-bench baseline belongs there (your call to defer it stands). PR2 (#268-blocked) and PR3 (triad) build on this corrected foundation, unaffected otherwise.

Nice catch on the load/put race — and especially on surfacing the granularity conflict instead of burying it. Re-request review when the per-injection-point version is up.

siy:
## ⏸️ On hold — do not merge

Review correctly steered this to **per-injection-point** granularity; the per-slice foundation in this PR is superseded. Reusable parts (config snapshot, `ObservabilityAspect`, registry + key/value + serializer wiring, register/deregister lifecycle, and the **load/put race fix**) carry into the rework; the delta is granularity → per-injection-point, the key → `(artifactBase, methodName)`, and an **`AspectFactory`** codegen change (always-generate-wrapper + per-method weave). No envelope bump (GA format = 1000).

Postponed for a deeper codegen design pass. Full state, locked decisions, and open questions are in the #277 comment and `aether/docs/internal/progress/277-observability-design-state-2026-06-25.md` (this branch). I'll re-request review when the per-injection-point version is up.

siy:
Superseded by the dispatch-seam design shipped on release-1.0.0-rc2 (#277 closed — see the reconciliation there). This PR's reusable parts (config snapshot, registry skeleton + load/put race fix, KV serializer shape) were carried into increment 1 (c1224d720), re-keyed per-injection-point; the per-slice seam approach it prototyped was replaced by strategy cells at the dispatch seams, and the factory Aspect seam it threaded was deleted in increment 6 (fd1649ad0). Thanks — the race-fix and registry lifecycle survived to production intact.

Issue #336, opened 2026-06-18
Reconciler-under-load: leader-pinned CTM provisioning (scale-up/auto-heal) stalls on a degraded/leadership-churning cluster; fine when healthy


siy:
## Summary

**Reconciler-under-load class.** When 2 core nodes are deleted simultaneously (5→3), the cluster stalls at `actualCoreCount=3, state=RECONCILING` and never self-heals across two 720s windows. The deficit-follow-up loop fires and keeps calling `provisionReplacement`, but the underlying `lifecycleManager.provisionNode` attempts **fail (~49× observed)** → provisioning circuit breaker trips → cluster stays wedged at 3/5.

Single-node deficits heal in seconds (cloud-proven). The cluster **survives** the double-kill with quorum + a committed leader (no data loss) — this is a self-heal completeness gap, not an availability outage.

## Severity

**Non-RC-blocking.** Quorum and leadership both survive a simultaneous double core kill; single-node auto-heal works. This is deferred hardening, not a release blocker.

## Confirmed

- **NOT quota.** Kills use `hcloud server delete` (frees Hetzner quota); the breaker is reset before each subtest; still reaches `priorFailureCount=49`. Quota exhaustion is refuted as the cause.
- The deficit-follow-up loop **does** fire and **does** keep invoking `provisionReplacement`.
- The actual `provisionNode` attempts genuinely fail ~49× → breaker opens → stuck.
- A contributing defect — suppressed-success masking the deficit (a no-boot deferral stamping a phantom in-flight placeholder) — was **already fixed** in `4fe6f6fdc` (`ProvisionDisposition`). That made any transient failure burst *permanent*; with it fixed, a burst should no longer wedge **if** the underlying provision attempts eventually succeed.

## NOT yet pinned (the open question)

**Why does `lifecycleManager.provisionNode` fail ~49×** under a 2-node burst? Candidates not yet distinguished: capacity/rate-limit at the provider, user-data rendering under churn, or a replacement-led-leader path. Single-node provisioning on the same path succeeds, so it is something specific to the concurrent/burst case.

## Blocked on — observability (do this first)

Root-causing is gated on being able to read the leader's `provisionNode` failure cause on the destructive (cluster-B) test cluster. Today that is impossible:

[...]

PR #361, opened 2026-06-26
fix: #336 SWIM node-add eviction — OBSERVED birth state (validated on real cluster)

siy:
## #336 — SWIM evicts freshly-added nodes before their first probe-ack

**Root cause:** SWIM born a freshly seeded/announced member as `SUSPECT` with its death timer armed at birth, while the first probe is gated ~10s behind `startupDelay` and the only protecting grace covered the observation stream but not the `onMemberFaulty` death-path. So a live-but-not-yet-probed joiner was FAULTY'd before confirmation — a regression from the 2026-06-13 failure-detector rework (the original SWIM seeded joiners ALIVE). Node-type-agnostic: hits core scale-up (#336) and worker join (#241) identically. Full diagnosis: `aether/docs/internal/progress/336-swim-node-add-eviction-diagnosis-2026-06-25.md`.

**Fix:** a new **local-only `OBSERVED`** birth state — known + probe-eligible, but not alive and not death-timer-armed, until a real probe-ack (or gossiped `Alive`) promotes it; a sustained probe-timeout past the join deadline escalates to SUSPECT (normal co-confirmed death path). `OBSERVED` is the weakest gossip-merge priority (gossiped `Alive` promotes it; gossiped SUSPECT/FAULTY is ignored — own probing decides), never serialized on the wire, never counted toward quorum. Subsumed born-SUSPECT band-aids removed; #231 / #126 / #94 invariants preserved.

### Commits
- `00be33514` OBSERVED birth state + wire guard + band-aid removal
- `a15b0e1f9` gossip-merge correction (Alive promotes, SUSPECT/FAULTY ignored, gossip-SUSPECT-of-unknown → OBSERVED)
- `08ca58349` `CommunitySizing` config (#241; defaults unchanged)
- plus the tracked-red repro (`CommunityFormationProbeTest`) + diagnosis/handover docs

### Validation
- `mvn -pl integrations/swim test` → **170/170**; line-by-line review + full `jbct-reviewer` pass
- **Real Docker cluster** (`run-tests.sh --env remote`): scale-up 5→7 in **25s, zero evictions**; suite-02 kill-leader/node/multiple/under-load + **joining-window-kill** + suite-13 worker-join all PASS; full-cluster total-restart (graceful + abrupt) re-forms cleanly
- The single-JVM **Forge harness cannot validate node-add at 6-8 nodes** (transport/probe-ack contention — harness artifact, not the fix); the two probes stay tracked-red against that limit

### Notes for review
[...]

Issue #378, opened 2026-06-29
Stream sync replication unsatisfiable: effective RF is one peer short (minSyncReplicas>=1 always fails NOT_ENOUGH_REPLICAS)


siy:
## Summary
With `minSyncReplicas >= 1`, **every synchronous stream publish fails `NOT_ENOUGH_REPLICAS`**. The effective replication factor counts the owner, but the await requires `minSyncReplicas` *distinct non-self* acks — so the replica set is always one peer short.

## Evidence (HEAD `e320881f0`)
- `ReplicaPlacement.replicationFactor(APP, minSyncReplicas, N) = clamp(minSyncReplicas, 1, N)` — `ReplicaPlacement.java:108`, set at `ReplicaSetController.java:301`. The HRW replica set is owner-first, so RF **includes** the owner.
- Both publish paths await `minAcks = minSyncReplicas`: `PartitionedStreamAccess.java:527`, `DefaultStreamPublisher.java:226`.
- `awaitReplication` counts only DISTINCT NON-SELF acks and prechecks `targets.size() < minAcks -> NOT_ENOUGH_REPLICAS` — `DefaultReplicationManager.java:162,235`.
- Therefore peers = RF - 1 = `minSyncReplicas - 1` < `minSyncReplicas` ⇒ always short, for every `minSyncReplicas >= 1`.
- The existing unit test `selfInReplicaSet_isNotCounted` (`ReplicationManagerTest`) **asserts this shortfall as expected** — it encodes the bug rather than catching it.

## Impact
Synchronous-durability publish (`minSyncReplicas > 0`) is unusable end-to-end. The default RF=1 eventual path is unaffected. `feature-catalog.md` row 139 ("Sync replication ack — Complete") is inaccurate.

## Suggested fix
Set the effective APP replication factor to `minSyncReplicas + 1` (owner + N synchronous peers), clamped to cluster size, and fail clearly when `minSyncReplicas + 1 > N` (don't silently under-provision). Update `selfInReplicaSet_isNotCounted` to reflect the corrected provisioning.

## Related
#262 (CLOSED) reworked ack identity / self-count / ack-before-register race — **adjacent**, did not fix the placement count.

---
Found by the consistency-lens guarantees audit, 2026-06-29. See `aether/docs/reference/guarantees.md` §7 (C1) and `guarantees-corrections-needed.md`.

siy:
[...]

Issue #337, opened 2026-06-21
Migration: dialect-aware SQL statement splitting + execution strategy (PG/MySQL/MariaDB/Oracle/DB2/SQL Server)


siy:
## Why (pre-GA, data-layer)
Migrations are a foundational data-layer operation, and **Aether is multi-DB by design** — `DatabaseType` enumerates PostgreSQL, MySQL, MariaDB, Oracle, DB2, SQL Server, H2, SQLite, CockroachDB, and the schema-management design doc mandates an *"engine-agnostic migration engine, engine-specific scripts."*

The current runtime executor does not honor that. `AetherSchemaManager.executeStatements` (`aether/aether-deployment/.../schema/AetherSchemaManager.java:269-281`) splits with `sql.split("\n")` (strip `--` lines) then **`split(";")`**. That is not engine-agnostic — it is **broken for every engine**:
- Any `;` inside a string literal, a `$$…$$` / `$func$…$func$` body, an `E'…'`, or `COPY … FROM STDIN` data **mis-splits the migration**.
- For non-PG dialects the boundary often **isn't a bare `;`** at all (MySQL `DELIMITER`, SQL Server `GO`, Oracle `/`).
- The runtime splitter has **zero tests**.

(Separately, the PEG grammar `postgres.peg` used by build/lint tooling has its own `$$`-only `DollarString` + `;`-blind `RestOfStatement` gaps — lint-path only, reconciled as a secondary item below.)

## Design — one pure lexer engine + per-dialect descriptors (data, not code)
Statement splitting is a **lexical** problem and is cleanly parameterizable. Build **one state-machine splitter** + small `DialectSpec` descriptors. The splitter **models zero statement semantics** — statements come out verbatim and execute verbatim; an unmodelable function body still splits and runs. This is the only cross-DB-viable approach (a PEG grammar would need one parser per dialect).

`StatementSplitter.split(sql, dialect) : Result<List<Statement>>` — pure, deterministic.

**Two axes per descriptor:**

*Lexical table* (spans to skip when seeking a boundary): string quoting (`'…'`, `''`, backslash escapes, `E'…'`, `N'…'`, `q'X…X'`), identifier quoting (`"…"`, `` `…` ``, `[…]`), line comments (`--`, `#`), block comments (`/*…*/`, nested where the dialect nests).

*Boundary-mechanism primitives* (composable; each dialect enables a subset):

| Primitive | Triggered by | Dialects |
|---|---|---|
[...]

PR #391, opened 2026-06-30
fix: #336 — cluster-sync reachability evidence (false-eviction of established peers under churn; complements #361)

siy:
## What
Closes the **complementary #336 case that #361 (`OBSERVED` birth state) leaves open**. #361 protects a *freshly-added* node at join; this fixes an *established* HEALTHY peer transiently SUSPECTED under load/jitter.

Two synergistic commits (do not split):
1. **`9156887a8` — option-1 reachability evidence.** `ClusterSyncContext.emitPingTimeoutIfExceeded` no longer does a destructive `network.disconnect` + eviction-hint broadcast on transient SWIM-SUSPECT; it feeds `recordTransportHint(PeerUnreachable(PING_TIMEOUT))` into SWIM (the same pipeline the QUIC `onPeerLeft` listener uses), which is **refuted** when pongs resume. Removes the eviction-hint broadcast that is an S20 self-drain-cascade trigger.
2. **`e26c2f809` — SWIM refutation hardening.** At-risk proactive self-refutation (own incarnation advances on inbound-reachability silence, out-ranking an unheard `Suspect(self)`); a co-confirmation kill-gate (≥2 distinct accusers OR transport-unreachable hint) before terminally departing an ever-HEALTHY peer on one prober's first-hand FAULTY; `log(N)` suspect-window scaling. (The kill-gate's transport-veto arm is what option-1's hint trips — mutually dependent.)

## Reconciliation with release
All changes are **additive** to `release-1.0.0-rc2` (release has no equivalent of any); verified by reconciliation against #361 + the 53 intervening commits. `ClusterSyncContext` on release still has the destructive disconnect — the gap is real and open.

## Validation
- **Units:** 170 green (`integrations/swim` + `aether/aether-metrics`), incl. Wave-6 / Tombstone / DeathPathCoConfirmation / `ClusterSyncFsmTest`; clean compile.
- **Real infra:** `run-tests.sh --env remote --suites 02` kill suite — a killed node is removed from membership in ~6s and the cluster re-elects/fails-over; the dropped-broadcast convergence risk did not materialize.
- ⚠️ The in-JVM `CommunityFormationProbeTest` is **not a valid gate** — single-JVM CPU starvation (8 nodes/8 cores) makes nodes genuinely unreachable (LHM sawtooth = late acks, not absent). Validate on real infra.

## Follow-up
Delete the now-inert `evictionHints`/`processEvictionHints` plumbing (broadcast's sole writer removed). See `aether/docs/internal/progress/336-reachability-evidence-fix-2026-06-30.md`.


siy:
## Phase-0 reply (aether-main)

**(a) #391 MERGED** — merge commit `ee5b36b50` on `release-1.0.0-rc2`. Used a merge commit (not squash) deliberately: the in-PR progress doc references the fix SHAs `e26c2f809`/`9156887a8`, which squash would dangle. Design stream: rebase onto `ee5b36b50`. Review verdict: scope confirmed SWIM/membership-only (the `AetherNode` hunks are SWIM wiring + `PING_TIMEOUT` cause; nttDisconnectTap change is comment-only); the three fixes compose correctly (option-1's hint is what trips Fix-2's transport-veto arm, so genuine deaths aren't slowed by the kill-gate); Fix-1 rate-limit preserves anti-oscillation; Fix-3 canonical + clamped. Real gates green (build-and-test, CodeRabbit); forge-tests red = known `/data` disk-tier env-flake. **Two non-blocking follow-ups:** (1) `withinJoinGrace` widened to `public` with no in-PR caller — dead API until the cluster-sync guard that uses it lands; (2) the now-inert `evictionHints`/`currentEvictionHints`/`processEvictionHints` plumbing sweep (your own note). Suggest folding both into the #262 stack or a small follow-up.

**(b) No local unpushed stream-subsystem work.** Working tree clean, HEAD == origin throughout. Today I closed **#337/#338/#333 as reconciliation-only closures** — zero new code, zero commits under `aether/aether-stream/**` (the #333 fix/sensors were already on origin: `d2dd76021`/`20eb0ac0d`/`dfed0822d`). The two rc3 follow-ups I filed (#408 PEG lint path, #409 H2/SQLITE dialect) live in `pg-tools`/`pg-parser` — outside the freeze scope. Nothing on my side needs to enter the reconciliation. Heads-up only: #333's closure was validated against the pre-#262 sensor surface; since `51875e69a` rewrites `test-stream-replica-failover.sh` and strengthens the convergence predicate, I'll re-confirm the closure rationale still holds during Phase-2 review (the in-JVM `WriteIdleResidualReverify` suite + the fix itself are untouched by your stack, so I expect it stands).

**(c) No objection to the #378 supersession.** Two-knob (independent `replicas` + `min-sync-replicas`) is strictly more expressive and Kafka-proven; my single-knob derivation was the minimal fix and should yield. Three expectations for the #262 PR (review focus, not conditions): (1) the #378 regression tests are **migrated, not deleted** — they encode the off-by-one; (2) guarantees §7 + streaming-spec §10.5 state the guarantee **per-operation**: publish ack = quorum-durable on `minSync` replicas before ack, remaining `RF − minSync` eventual; make `minSync=1` (owner-only ack, no barrier since `minAcks=0`) and the `minSync=0` edge semantics explicit; (3) my `45e8c9688` §7 retirement entry for #378 is rewritten as "resolved by two-knob" (already in your plan).

**Freeze acknowledged and in effect on my side** — no commits to `aether/aether-stream/**`, stream parts of `aether/node`, `slice-api` StreamConfig, or `StreamConfigParser` until the #262 PR lands. The migration cluster is already closed (no held #333 work), so no interaction. My next work avoids the frozen paths.

Issue #493, opened 2026-07-19
aether corpus debt from #451 rule batch (~160 sites) — BND-01 disposition gates its ERROR restore


siy:
## Context

The #451 easy-tier rule batch (e50e712d6) was corpus-validated before landing (the #489 protocol). Rules with corpus findings ship at WARNING; this ticket tracks the burn-down and the severity restores. All counts deduped file:line from the 67-module sweep + sample audit (0% rule-FP on both audited heavy rules; RET-08's noisy null-compare arm was dropped from the rule instead of burning down — see #451).

## Debt inventory (~160 deduped sites)

1. **JBCT-BND-01 (3 sites, gates ERROR restore)** — `AwsLoadBalancerProvider.java:119,:125` (Optional; **aether-main's cloud front** — needs their disposition: fix to Option or excludePackages-scope the aws adapter) + `SliceStore.java:13` (CompletableFuture import). Restore ERROR when dispositioned.
2. **JBCT-SEAL-02 (~50 sites)** — 100% zero-component record Cause variants → enum (audit: 8/8 true positive; incl. the `unused()` placeholder-filler idiom ×3). Mechanical-to-light-judgment. Top: aether-config 10, cli 9, slice-api 8, forge 6.
3. **JBCT-RET-08 null-arg (~90-120 sites)** — literal null call arguments (audit: mostly REAL — orElse(null) bridges, null in Map.entry, ctor nulls; a few JMX/JDK-imposed BOUNDARY → justified suppressions). Top: cli, pg-tools, pg-codegen, node.
4. **JBCT-STY-09 (4 sites)** — real nested ternaries (AetherSchemaManager:456, ConfigurableLoadRunner:801, DdlAnalyzer:730, SchemaBuilder:118).
5. **JBCT-MUT-01 (2 sites)** — JwtSignatureVerifier:97 (`pos`), MavenLocalRepoLocator:95 (`path`).

## Follow-ups (rule-side, jbct/)

- Maven check goal lints MAIN sources only — NAM-05 (test naming) never runs through the plugin gate; add opt-in test-source linting to the plugin (then corpus-measure NAM-05 before enforcing).
- Flow-aware null-compare successor for the dropped RET-08 arm → #455 hard tier.

## Acceptance

Sites fixed or justification-suppressed per the #489 conventions; BND-01 restored to ERROR after item 1; sweep clean for all five rules at final severities.

siy:
[...]

Issue #451, opened 2026-07-15
JBCT lint: easy-tier rule batch — 8 unchecked book rules implementable with existing CST infra


siy:
## Context

Cross-referencing the JBCT book (74 prescriptive rules) against the linter (41 implemented rules) found ~43 unchecked rules, of which 8 are **easy tier**: pure syntax, single-file, no type resolution — each implementable in the existing `CstLintRule` framework (self-driven DFS over the PEG CST) in a day or less. All are independent; batch or split freely.

Adding a rule = new `CstLintRule` class + append to `CstLinter.defaultRules()` (`jbct/jbct-lint/.../cst/CstLinter.java:88-187`) + severity in `LintConfig.DEFAULT` + fixtures (positive/negative — required, see #454).

## Rules

1. **`JBCT-BND-01` — forbidden boundary types in business logic** (ERROR). Flag `java.util.Optional`, `CompletableFuture`, `Mono`/`Flux`, `ResponseEntity` in imports or type positions. The book marks this HARD (four-return-types; systematic-application) and it's a plain text/import match — the most surprising gap in current coverage. Adapter-boundary code is exempt via the existing package-glob config (`excludePackages` or the layering classification from #452 once that lands).
2. **`JBCT-STY-09` — nested ternaries** (WARNING). Ternary containing a ternary anywhere — `JBCT-LAM-03` only covers ternary-in-lambda. Book: "extract each decision to a named function" (basic-patterns).
3. **`JBCT-NAM-03` — `*State` suffix discipline** (WARNING). `*State` reserved for lifecycle sum types (sealed interfaces); variants stay bare (`Free`, `Held` — never `HeldState`). Book: from-process-to-patterns; glossary.
4. **`JBCT-NAM-04` — local records lowercase camelCase** (WARNING). Records declared inside method bodies use `generationCache`-style naming, not PascalCase. Book: systematic CP8.
5. **`JBCT-NAM-05` — test method naming** (WARNING). `methodName_scenario_expectation()` for `@Test` methods. Needs test-file detection (path or `@Test` presence). **Decision required first:** the book states the 3-part order two ways — `methodName_scenario_expectation` (examples, systematic-application) vs `methodName_outcome_condition` (basic-patterns). The project standard (CLAUDE.md) is `methodName_scenario_expectation`; normalize the book's basic-patterns chapter to match (book repo change, out of this repo's scope — track in the book).
[...]

PR #500, opened 2026-07-23
#493: JBCT lint corpus burn-down (5 rules → 0) — SEAL-02/RET-08 FP-reduction rule refinements + aether fixes

siy:
## #493 — JBCT lint corpus-debt burn-down (all 5 rules → 0)

Closes the #451-batch corpus debt. All internal — **no behavioral or API change**; error messages, HTTP statuses, and wire formats are unchanged.

### Disposition per rule

| Rule | Sites | What changed |
|------|------:|--------------|
| **MUT-01** | 2 | Parameter reassignment → locals (`JwtSignatureVerifier.writeInteger`, `MavenLocalRepoLocator.expandPath`) |
| **STY-09** | 4 | Nested ternaries de-nested (guard+ternary / extracted method / early-`yield`) |
| **SEAL-02** | 50 | 20 real named fixed-message causes → per-cause `enum Foo implements X { INSTANCE; … }`; 30 `record unused()` placeholders **rule-exempted** |
| **BND-01** | 3 | Already dispositioned (sites clear, severity ERROR) — confirmed, no change |
| **RET-08** | 110 | Rule exemption extended + 44 fixes + ~29 justified suppressions |

### Two rule changes — both FP-reductions (safe)

Both are pure **exemptions** — they make `jbct:check` emit **strictly fewer** diagnostics, never more, so they cannot break a build:

- **JBCT-SEAL-02** now exempts the `record unused()` sealed-interface placeholder-filler idiom (a permitted-subtype stub, not a fixed-message cause — a 136-site repo-wide convention). Named-cause detection is unchanged.
- **JBCT-RET-08** exemption extended from just `.or(null)` to the distinctive JDK boundary adapters `orElse` / `compareAndSet` / `getAndSet` (an `Optional` bridge and `java.util.concurrent.atomic` empty-sentinels — not `Option`-wrappable). Common names (`set`/`init`/`load`/`invoke`) are deliberately **not** exempted; their nulls carry explicit `@SuppressWarnings` instead.

Unit-tested: SEAL-02 8/8 (incl. 2 FP-guard fixtures), RET-08 9/9 (incl. 3 new).

### RET-08 corpus (110 sites) — how they were cleared

~85% were JDK/framework-boundary or a legitimate data table (the same "correct JDK-boundary" pattern that got RET-08's sibling null-*comparison* arm dropped):

- **Fixed (44):** `TypeMapper` PG-type→Java-type table Option-ified (40 `null` → `Option.empty()`/`present`, factory now takes `Option<String>`); `OutputFormatter.printQuery` given a private `Option<TableSpec>` core (no external caller churn).
- **Justified-suppressed (~29):** `AtomicReference` clear-sentinels, `SSLContext.init`/`KeyStore.load`, reflective static `invoke`, JMX `addNotificationListener`, Jackson view-DTO absent fields (`AlertView` — GA wire contract), cloud provider-request DTOs. Each `@SuppressWarnings("JBCT-RET-08")` carries a one-line justification.

### Verification

- **Compile:** full 78-module reactor `BUILD SUCCESS` (the `printQuery` overloads and `TypeMapper` `Option` types all type-check).
- **Lint:** `mvn jbct:check` sweeps to **0** for RET-08 and SEAL-02 across every affected module; MUT-01/STY-09 clean in their files; BND-01 ERROR live.
- **Format:** `jbct:format` clean on the two signature-changed files.

### Merge notes for aether-main

- **Atomic:** the jbct rule changes and the aether corpus fixes are coupled — the corpus only sweeps clean with **both**. Merge as one unit.
- **Requires a jbct rebuild:** the new rules must be installed before the aether lint reflects them. The standard `./build.sh` bootstrap (build jbct → install → lint aether) handles this; a plain `mvn jbct:check` against a stale `~/.m2` jbct would not see the exemptions.
- No files here overlap aether-main's recent commits (rebased conflict-free onto `437ca28a4`).


siy:
**Pre-merge coordination note (shared `~/.m2`):** my local `~/.m2` currently has the jbct build from this PR installed (the SEAL-02 + RET-08 exemptions), so it is *ahead* of what's on `release-1.0.0-rc3` until this merges. Both rule changes are FP-reductions (strictly fewer diagnostics — they cannot fail a build). Caveat: if you rebuild jbct from a pre-merge branch you'll install the OLD rules and the exemptions disappear until #500 lands. Recommended order: merge this, then rebuild jbct, to keep `~/.m2` aligned with the branch. (This is the documented shared-`~/.m2` cross-stream swap hazard; the MAILBOX entry in this PR's diff documents the rule changes in full.)

Issue #533, opened 2026-07-26
Lint score denominator is a placeholder — ScoreCalculator checkpoints TODO makes absolute scores uninterpretable


siy:
## Problem

The lint score's denominator is a placeholder. `ScoreCalculator.java:105` carries:

```java
// TODO: Get actual checkpoint counts from linter
```

The formula is `category_score = 100 * (1 - weighted_violations / checkpoints)`, where `weighted_violations = SUM(count[severity] * multiplier)` with ERROR x2.5, WARNING x1.0, INFO x0.3. With `checkpoints` a proxy rather than a real count of checked sites, the absolute category scores — and the weighted `overall` derived from them — do not mean what they appear to mean.

## Consequences

- Absolute scores are not interpretable. "Overall 33" cannot be read as "33% compliant" against anything real.
- Cross-project and cross-module comparison is unsound, since the proxy does not scale with the amount of code actually checked.
- `ScoreMojo`'s `jbct.score.baseline` gate (`ScoreMojo.java:24-25`, VERIFY-phase failure at `:49-50`) keys on `overall`. Anyone who configures a baseline is gating CI on a number with a fabricated denominator. The gate is dormant today — no pom in the repo sets a baseline — which is the only reason this has not bitten yet.

## What a real denominator would need

`checkpoints` should reflect the number of sites a rule actually examined, which the linter knows at analysis time but does not currently report. That likely means threading a per-rule (or per-category) checked-site count out of the rule execution path alongside the diagnostics.

## Acceptance

- `checkpoints` derives from counts reported by the linter, not a proxy.
- The TODO is gone.
- A test pins the relationship: for a fixed corpus, changing only the amount of *unviolating* checked code moves the score in the expected direction.
- If a real count turns out to be infeasible for some rule families, that is documented explicitly at the call site and the score output says what it is relative to — rather than presenting an absolute-looking number.

## Scope note

Filed separately by owner ruling so it stays independent of the STYLE score-category work (adding a weight-0 advisory bucket for the 14 previously-uncategorized style/naming/zone rules). That change is neutral with respect to this one: it does not touch the denominator, and it does not depend on it being fixed. Both, however, bear on whether the reported score can be trusted.

KaustAbhinand:
Hi! Can I work on this?

siy:
Investigated. The issue as I filed it understated the problem, and the fix is not the one it proposed.

## It is not a placeholder proxy — it is self-referential

[...]

Issue #449, opened 2026-07-15
JBCT score: ScoreCalculator/RuleCategoryMapping keyed to retired rule IDs — all diagnostics fall to default bucket


siy:
## Problem

The lint scoring subsystem is keyed to a **retired rule-ID scheme** — none of the 41 active rules emit the IDs it maps.

`RuleCategoryMapping.MAPPING` (`jbct/jbct-core/src/main/java/org/pragmatica/jbct/core/score/RuleCategoryMapping.java`) maps `JBCT-NULL-01/02`, `JBCT-EXC-01/02`, `JBCT-CHAIN-01`, `JBCT-NEST-02`, `JBCT-RES-01`, `JBCT-VO-03/04`, `JBCT-ERR-01`, `JBCT-LAM-04`, `JBCT-NAME-01/02`, `JBCT-IMPORT-01`, `JBCT-FQN-01`, `JBCT-DOMAIN-01`, `JBCT-LOOP-01`, `JBCT-PARSE-01` — while the live registry (`CstLinter.defaultRules()`, `jbct/jbct-lint/.../cst/CstLinter.java:88-187`) emits `JBCT-RET-*`, `JBCT-EX-*`, `JBCT-SEQ-01`, `JBCT-NAM-*`, `JBCT-STY-*`, `JBCT-MIX-01`, `JBCT-PAT-*`, etc.

Consequence: `ScoreCalculator.categoryFor()` falls through to its default for virtually every diagnostic, so nearly all findings land in PATTERN_PURITY and the 7-category weighted score (Return Types 25 / Null Safety 20 / Exception Hygiene 20 / Pattern Purity 15 / Factory 10 / Lambda 10 / …) is meaningless. `ScoreMojo` output is currently not trustworthy.

## Scope

1. Rebuild `RuleCategoryMapping` against the current 41 rule IDs (source of truth: `defaultRules()` + `LintConfig.DEFAULT`, `jbct/jbct-core/.../lint/LintConfig.java:25-125`).
2. Make the default bucket loud: unknown rule ID in scoring ⇒ warning (or build-time check that every registered rule ID has a category), so the mapping can never silently rot again.
3. Re-validate the category weights once mapped — the current weights were tuned against the old scheme.
4. Follow-up (non-blocking, after #448): base the Pattern Purity category on classifier shape results instead of rule-ID proxies.

## Acceptance

[...]

PR #536, opened 2026-07-26
feat: STYLE advisory score category (weight 0) surfaces 14 previously-invisible rules

siy:
## Summary

Adds a `STYLE` score category with **weight 0** — measured and reported like any other category, excluded from the weighted `overall`. This ends the silent exclusion of 14 registered rules.

## The problem

`RuleCategoryMapping` partitioned the 65-rule registry into 51 categorized rules plus an explicit `UNCATEGORIZED` set of 14: `STY-03/04/06/07/08/09`, `ORD-01`, `STATIC-01`, `LOG-01/02`, `NAM-05`, `ZONE-01/02/03`. `categoryFor` returned `Option.none()` for those and `ScoreCalculator.isInCategory` dropped them from every category.

Unlike genuinely *unknown* rule IDs — which at least hit `warnUnknownRule` once per run — these 14 produced **no signal anywhere**. Measured over the full source trees:

- aether: **6317 of 13184 diagnostics (47.9%)** invisible in the reported score
- jbct-self: **1642 of 4498 diagnostics (36.5%)** invisible

Dominated by `STATIC-01` (2272 aether / 715 jbct-self), `STY-06`, `ZONE-02/03`, `NAM-05`. A user saw a number and could not tell that nearly half the findings had not been counted. #449 already established that silent miscategorization in this scorer is a real bug rather than a style nit.

## What changed

All 14 rules now map to `ScoreCategory.STYLE`, and the `UNCATEGORIZED` set is **gone** — the mapping is a total function on the registry. `advisory()` is derived from `weight == 0.0` rather than a parallel flag that could drift out of sync. `overall` is summed over `weightedCategories()` explicitly rather than relying on a zero multiplier to do it implicitly. The unknown-rule warn path is untouched and still fires for unregistered IDs.

**A 7th number that silently doesn't count would be the same defect in a new place**, so both output surfaces mark advisory categories explicitly:

- **Terminal** — advisory rows sit below a divider, tagged `STYLE (advisory)`, with a closing legend: `(advisory) — weight 0: reported for visibility, NOT counted in the score above`.
- **JSON (breaking, pre-GA)** — each `breakdown` entry is now an object rather than a bare number: `"style": {"score": 9, "weight": 0.0, "advisory": true}`. Every category self-describes its weight, so no consumer has to guess which numbers form the total. This also fixes a latent locale bug — the `%.1f` weight is now formatted with `Locale.ROOT`, so a comma-decimal locale can no longer emit invalid JSON.
- **Badge** — renders `overall` only and never showed a per-category number; unchanged.

The terminal renderer was **duplicated verbatim** between `ScoreCommand` and `ScoreMojo` (box drawing plus `createProgressBar`). It moved into one shared `ScoreReport` in jbct-core, so the CLI and the `jbct:score` goal cannot drift apart on how advisory categories are presented — the same duplication-removal rationale as the `PackageGlob` extraction in #531.

## Verification

- **650/650 tests pass** (jbct-parser + jbct-core + jbct-lint, after a clean `target` wipe). jbct-maven-plugin and jbct-cli `test-compile` clean from scratch.
- **Registry bijection**: `CstLinter.defaultRules()` emits exactly 65 rules; `MAPPING` has exactly 65 entries. 51 previously mapped + 14 moved = 65, nothing dropped.
- **Weight-0 neutrality proven empirically**, not just argued: the pre-change jar was built in an isolated `git worktree` and both jars run over a byte-identical tree. Full-tree `overall`: aether 4→4, jbct-self 4→4, with STYLE reporting 9% and 10% alongside the unchanged six. Also confirmed per-module: aether/node 42→42, aether/cli 51→51, aether-stream 73→73, aether-deployment 65→65, jbct-lint 21→21.
- **Old invariants re-expressed, not weakened.** `mapping_partition_coversRegistryExactly` is now *tighter* (previously allowed a rule to satisfy it by living in either MAPPING or UNCATEGORIZED; now requires the keyset to equal the live registry exactly). `categoryFor_zoneRule_isUncategorized` became `_mapsToStyle`, asserting `Option.some(STYLE)` rather than `none()`.
- JSON validated through `python3 -m json.tool` on both corpora; box alignment holds at 52 chars per row with the longer `STYLE (advisory)` label.

## Pre-existing issues found while verifying (not introduced here, not fixed here)

1. **No test coverage on the `ScoreCommand` / `ScoreMojo` output wiring** — repo-wide search found no golden or snapshot test for either call site. `ScoreReportTest` covers the new shared renderer well, but the surfaces that invoke it have none. Worth a ticket.
2. **`score` on a large tree throws `StackOverflowError`** in `java.util.regex.Pattern` during file collection at default `-Xss`; works with `-Xss8m`. Reproduces on the pre-change jar, so unrelated to this diff. Worth a ticket.
3. `FileCollector` interleaves `Skipping...` lines into stdout, which makes raw `score --format json` output not directly pipeable to `jq`. Pre-existing.

## Context

Implements the pending "style score category" ruling: a distinct bucket for the uncategorized style/naming/zone rules. The weight-0 advisory form was chosen over a weighted 7th category so that no arbitrary weight has to be invented and every historical score stays comparable — the data this now surfaces is what would justify a real weight later, if one is wanted.

The score's denominator remains a separate known problem, filed as #533: `ScoreCalculator`'s `checkpoints` is a placeholder, so absolute scores are not interpretable. This change is neutral with respect to it.


siy:
Correction to this PR's own description, found after it was opened.

The description cites STYLE reporting **9%** on aether and **10%** on jbct-self as though those describe style quality. They do not. They are the scoring formula's floor.

`ScoreCalculator.countCheckpoints` derives the denominator from the violation count itself:

```java
checkpointMap.put(category, (int)(violations * 1.1 + 10));
```

So for a WARNING-dominated category the ratio tends to `v / (1.1v + 10) → 1/1.1`, and the score asymptotes to `100 × (1 − 1/1.1)` = **9.09**:

| WARNING-only violations | Reported score |
|---|---|
| 100 | 17 |
| 1,055 | 10 |
| 2,272 | 9 |
| 10,000 | 9 |
| 1,000,000 | 9 |

The measured values match the formula exactly: 6317 aether diagnostics → 9, and 1642 jbct-self diagnostics → 10. Any sufficiently large warning count yields the same pair of numbers. (At the other end, an ERROR-only category saturates to 0 at just 10 violations, since the multiplier is 2.5.)

**This does not change the case for merging.** The defect is pre-existing and applies identically to all six existing categories — this PR neither introduces nor worsens it. What the PR actually fixes is real and independent: 14 registered rules produced no signal anywhere, and 42–48% of diagnostics were invisible in the report. The weight-0 neutrality proof also stands, since `overall` is unchanged by construction.

What is *not* true is that STYLE's number tells you anything about style. It tells you "this category has a lot of warnings", which the diagnostic count already said more honestly.

Tracked in #533, which turned out to be a semantic problem rather than a plumbing one: no definition of "checkpoint" exists anywhere in the codebase, specs, or the book (the book uses the word for an unrelated manual review checklist). The agreed direction there is to drop the 0–100 percentage in favour of violation density, which needs no invented population and is comparable across codebases. When that lands, every category's number changes meaning — STYLE included.

Also relevant to reviewers: this same degeneracy explains why the same corpus scored 33 in one measurement and 4 in another. Different scan scopes produce different violation counts, and the formula turns that into arbitrarily different scores.

PR #600, opened 2026-08-16
fix: jbct on peglib 0.7.2 — restore 1962 silently-dropped lint findings, plus three linter defects (supersedes #602)

siy:
Supersedes **#602**, which is closed in favour of this PR. Rebuilt on top of `release-1.0.0-rc3` **after #618 merged**, so all three streams of jbct work land together and #618's fixes are preserved.

## What's in here

**1. peglib 0.6.2 → 0.7.2** (published to Central today; `v0.7.2` at `6fd769b`, signed).

The substance is not the version. peglib 0.7.1 changed how the grammar spells type members — interfaces and records lost the `Member` wrapper that classes keep — and **the linter silently stopped reporting 1962 of 13274 findings** on the aether corpus. `CstNodes.findAllMethods` keyed on `MEMBER`, so it returned empty for every interface, which is JBCT's primary subject matter. 48 call sites inherited it. Fixed by reconciling the shapes in `CstNodes` and migrating ~20 rule sites.

**2. Three linter defects** (was #602):

- A **trailing comment after a bare annotation disabled it**. Not line-text matching — a node's SPAN reaches past its last real token into trailing trivia, so `text(qualifiedName)` returned `"Contract  // reason"`. It ran the other way too: `@Test  // comment` stopped registering as a test, so **JBCT-NAM-05 silently skipped those methods** — a false negative no corpus diff can surface. Fixed with `CstNodes.tokenText` (non-trivia tokens).
- **JBCT-BND-01 matched boundary types by simple name**, so a domain `Expression.Optional` was reported as `java.util.Optional`. Now matched by origin. **See the known gap below.**
- **JBCT-STY-03 reported qualified names inside string literals and comments** — worst case javadoc, where `[org.pragmatica.lang.Cause]` cross-references *must* be qualified, so the rule penalised correct documentation.

## Verification

```
jbct reactor      1475 tests, 0 failures
pg-tools           813 tests, 0 failures (2 Testcontainers errors need Docker)
lint over aether  2161 files, 14,304 findings — runs clean end to end
```

Both PRs' own suites pass in the combined tree: `BoundaryAndAnnotationTriviaTest` 10/10, `CstBoundaryTypeRuleTest` 8/8, `MapperSafetyTest` 11/11, `MemberShapeRegressionTest` 12/12, and **#618's `AnnotationCommentPreservationTest` 5/5**.

Two of the three linter fixes verified **end to end through the CLI**, not just via unit tests:

- STY-03: a file with a qualified name in javadoc, a line comment, a string literal, and one in real code yields **exactly 1 finding** — the code one.
- NAM-05 / annotation trivia: `@Test  // comment` is still seen as a test method.

## Merge, not rebase — and why

Both branches were 53 commits behind and overlap in 5 files, and #600 rewrites `FlowPrinter` (+211/−64) — the same file #618 fixed. Replaying 22 commits would force that conflict repeatedly, and each replay is a chance to silently drop #618. Merging resolves once, verifiably.

Only two conflicts, both resolved keeping **both** sides:

- `CstFullyQualifiedNameRule` — `MapperSafety.blankNonCode(memberDeclText(method))`. `memberDeclText` handles the 0.7.x member shape (#600); `blankNonCode` masks comments and literals (#602). Same composition `MethodShapeClassifier` already uses.
- `jbct/CHANGELOG.md` — both PRs' entries kept.

Checked explicitly that the merge preserved #618 (`flushInSpanOwnLineComments`, `SpecialFuncExpr`, the DML corpus, `CorpusParseTest`) and that every overlapping file carries both PRs' markers rather than one.

## Known gap — filed, not fixed here

**BND-01's origin fix does not take effect through the CLI.** #602's own fixture asserts 0 findings and its unit test passes, but the same file linted via the CLI reports 2. Verified this reproduces on **#602 alone**, so it is pre-existing and not introduced by the merge. It is also **not a regression** — before #602, simple-name matching flagged this case too. Tracked separately; the other two linter fixes are unaffected and verified working.

## Follow-up

**JBCT-EX-02 burn-down.** With the member shape repaired, anything gating on a clean lint run must reckon with the restored findings. Worth noting the current corpus reports EX-02 at 4 of 57 `.orElseThrow` calls — the earlier "53 violations" figure is not reproducible from this build and should be re-derived rather than assumed.


siy:
Re-pinned to **peglib 0.7.2** (`85c4c595d`), now that it is in the local repo. The property was also renamed `peglib.maven.plugin.version` → `peglib.version`: it always governed both the runtime dependency and the generator plugin, which ship together, while its name said otherwise.

Upstream 0.7.2 is `%import` composition, plugin import resolution and a single-flight parser cache — none of which touches the Java grammar. Verified rather than taken on trust:

- `Java25ParserV6` and `Java25Visitor` regenerate **byte-identical**.
- `Java25Lexer` differs by 104 lines that are a **pure permutation of the 54 keyword-map entries** — sorted `r0.put(...)` sets identical, and **not one non-`r0.put` line changed**. The raw line count looks alarming; the permutation check is what settles it.
- `regen_rulekind.py` delta zero, and regeneration is **deterministic** across runs — worth checking now that 0.7.2 made generation fork-join and this repo commits generated sources.
- Both sensors clean: formatter goldens green, and the corpus differential over the same 2135 aether files **identical — 14236 findings, 0 lost, 0 new, no per-rule change**.

One trap worth recording for the next bump: my first corpus comparison showed 392 lost / 420 new. That was **not** 0.7.2 — the baseline predated this branch's merge from `release-1.0.0-rc3`, so the aether tree itself had grown. Rebuilding both sides against the same tree (worktree at the pre-change commit) gives the identical result above.

Full reactor green on 0.7.2.

PR #602, opened 2026-08-16
fix: a comment after a bare annotation disabled it, and BND-01 flagged domain types by simple name

siy:
Fixes both defects from the jbct-maven-plugin bug report. Branched off `release-1.0.0-rc3`, independent of #600 (peglib 0.7.1) — either can merge first.

Both reproduce exactly as reported; both are fixed with regression tests, and the aether corpus is **byte-identical before and after** (14183 findings, BND-01's 4 genuine findings preserved).

## Issue 1 — a trailing comment defeated a bare annotation

**The reported cause is wrong, and the real one is worse.** Detection was already CST-based, not line-text matching. The defect is that a node's **span can reach past its last real token into trailing trivia**:

```
annotation 'Contract  // JVM entry-point contract: main must return void.'  span=[124,191)
```

So `text(qualifiedName).trim()` returned `"Contract  // JVM entry-point…"`, matched no known annotation name, and the annotation was skipped entirely. `@SuppressWarnings` was immune only because its `(...)` terminates the name span before a comment can be absorbed — which is why the report saw the two behave differently.

That makes **every bare annotation** vulnerable: `@Contract`, `@TerminalOperation`, `@NullReturn`.

**It also runs the other way, which the report did not catch.** The same one-line pattern in `CstTestMethodNamingRule` means an annotated test method stops being recognised as a test:

```java
@Test  // trailing comment
void badnamewithcomment() {}      // NOT flagged  <- JBCT-NAM-05 silently skipped it

@Test
void badnamenocomment() {}        // flagged
```

A false negative like this is invisible to any corpus differential, because it shows up as nothing.

**Fix:** `CstNodes.tokenText(node)` returns the node's non-trivia tokens; every comparison of node text against a *name* goes through it.

## Issue 2 — BND-01 matched by simple name

`headName()` reduced every type to its last dotted segment, so `Expression.Optional` and `java.util.Optional` both became `Optional`. The rule's own javadoc accepted this FP surface; 18 false positives across 9 files in a real project is the refutation.

**Fix — match by origin, with no cross-file type resolution needed.** The key fact is that **none of these types live in `java.lang`**:

| use | verdict | why |
|---|---|---|
| `Expression.Optional` | clean | qualified use names its own origin |
| `Optional`, no import | clean | cannot be `java.util.Optional` without one |
| `import com.acme.Optional;` + `Optional` | clean | imported from elsewhere |
| local `record Optional` + `import java.util.*` | clean | local declaration shadows the star import |
| `import java.util.Optional;` + `Optional` | **FLAG** | |
| `java.util.Optional<String>` | **FLAG** | |
| `import java.util.*;` + `Optional` | **FLAG** | |
| `import java.util.Optional;` | **FLAG** | the import itself |

An import is now flagged only when its fully qualified name is a boundary type, so `import com.acme.Optional;` is no longer a violation either.

## One change worth your attention

Four existing `CstBoundaryTypeRuleTest` cases asserted the old behaviour with fixtures that reference `CompletableFuture` / `Mono` / `ResponseEntity` / `Optional` with **no import at all** — code that does not compile, and the only shape under which simple-name matching was load-bearing. I added the imports the scenario requires rather than reverting the fix. If you would rather keep bare-simple-name detection as a deliberate heuristic, say so and I will restore it behind an explicit opt-in — but as written it cannot distinguish a domain type from the JDK one.

## Verification

- `mvn -f jbct/pom.xml test` — all 11 modules green.
- 10 new regression tests pin both directions: the true positives above, the two reported false positives, the `@Test` false negative, and `@Contract` with and without a trailing comment.
- Corpus differential against the pre-fix binary over 2135 aether files: **identical**, 14183 findings, no per-rule change, BND-01 still exactly its 4 genuine findings (`AetherSchemaManagerResumeTest`, `StreamFanoutConsumerTest` ×2, `StreamPublishReshuffleTest`).

## Not addressed

The report's `Future` suggestion is out of scope — `Future` is not in the rule's forbidden set today, and adding it would expand what the rule reports rather than fix a defect. Worth a separate decision.


siy:
Follow-up: linting my own diff surfaced a seventh defect and some self-inflicted findings. All fixed in `74387f6cf`, `37a618a46`, `e75aa630f`.

**JBCT-STY-03 reported qualified names inside string literals, text blocks and comments.** Found because this PR's own tests feed Java source to the linter — every `import` line inside a fixture text block was flagged. The rule scanned raw method text while 10+ sibling rules already mask non-code through `MapperSafety.blankNonCode`; it now does the same.

The worst case is javadoc: a `/// … [org.pragmatica.lang.Cause] …` cross-reference **must** be fully qualified to resolve, so the rule was penalising correct documentation.

Corpus effect on aether: **1130 → 957**. I classified all 173 removals rather than sampling — 25 in comments (mostly javadoc links), 148 inside string literals and text blocks, **zero genuine findings lost**.

**Self-inflicted findings from the BND-01 rewrite**, measured by linting the same files before and after my edit with the same binary (13 introduced): factory `NameScope.of` → `nameScope` (NAM-01), static-import `Option.some/none` (STATIC-01), a 6-step chain in `importedName` replaced with two precompiled patterns (SEQ-01), and the raw loop in `tokenText` (PAT-01). The new test file is now clean; its import order and a nested lambda are fixed too.

**Left deliberately:** 7 `JBCT-UTIL-02` advisories ("Consider using Verify.Is::…"), consistent with 4 already in that file. `Verify.Is` predicates are built for `Verify.ensure(value, predicate)` validation pipelines, and `Verify.Is.contains` takes `(String, CharSequence)` — so the suggestion against a `Set.contains` would not even typecheck. UTIL-02 has 917 corpus findings and its own suggestion-quality problem; narrowing it needs its own corpus validation and is not this PR's job.

Full reactor green; the BND-01 corpus result is unchanged (14183 findings, its 4 genuine ones intact).

siy:
Closing in favour of **#600**, which now contains all three of this PR's fixes.

#600 has been rebuilt on top of `release-1.0.0-rc3` after #618 merged, and this branch was merged into it. Two conflicts, both resolved keeping **both** sides rather than picking one:

- `CstFullyQualifiedNameRule` — `MapperSafety.blankNonCode(memberDeclText(method))`. `memberDeclText` is #600's 0.7.x member-shape handling; `blankNonCode` is this PR's masking of comments and string literals. Same composition `MethodShapeClassifier` already uses.
- `jbct/CHANGELOG.md` — both PRs' entries kept.

This PR's suites pass in the combined tree: `BoundaryAndAnnotationTriviaTest` 10/10, `CstBoundaryTypeRuleTest` 8/8, `MapperSafetyTest` 11/11. STY-03 and the annotation-trivia fix were additionally verified end to end through the CLI, not only via unit tests.

One gap carried forward and noted on #600: **BND-01's origin fix does not take effect through the CLI** — this PR's own fixture asserts 0 findings and its unit test passes, but the same file linted via the CLI reports 2. Confirmed to reproduce on this branch **alone**, so it is pre-existing rather than a merge artifact, and it is not a regression (simple-name matching flagged the same case before). The other two fixes are unaffected.

No work is lost by closing this.

PR #618, opened 2026-08-19
feat: migrate pg-tools to peglib 0.7.2 and close the SELECT-side corpus gap

siy:
Migrates `aether/pg-tools` from peglib 0.6.0 to **0.7.2** (published to Maven Central today, `v0.7.2` at `6fd769b`), and closes the coverage gap that let the migration's worst bugs stay invisible.

## Result

| Module | Tests | Failures |
|---|---|---|
| pg-parser | 311 | 0 |
| pg-schema | 112 | 0 |
| pg-codegen | 218 | 0 |
| sql-splitter | 172 | 0 (+2 Docker) |
| **Total** | **813** | **0** |

The 2 errors are `StatementSplitter*DiffTest` needing Docker — the same two excluded from the 0.6.0 baseline of 807.

Verified **clean-room**: `~/.m2/repository/org/pragmatica-lite/peglib*` deleted and rebuilt, so this consumes the published artifact (`_remote.repositories` reads `peglib-0.7.2.jar>central=`), not the local install that stood in during development.

## What 0.7.2 changes for a consumer

peglib 0.7.x is lex-then-parse. Three consequences reach the CST, and every consumer fix here follows from them:

1. **An identifier can arrive three ways.** `id` → `Token ColId`; `public` → `Token PublicKW` (a named rule spells that literal and claims the kind); `name` → **anonymous `Terminal`** (collides with an inline literal). So `findAll("ColId")` drops an arbitrary subset — silently, because a missing identifier reads as "no column here". All identifier lookups are now **positional**.
2. **Intermediate rules materialise.** `TableElement → ColumnDef → ColId` where 0.6.0 inlined them, so direct-child tests looked one level too high.
3. **A synthetic `_ROOT`** wraps the parse above the start rule.

## Three latent grammar bugs, exposed by the stricter toolchain

None was caused by peglib; all were real defects 0.6.0 concealed.

- **`!ReservedKeyword` never fired once at 0.6.0** — `ReservedKeyword` appears 0 times in 10,725 lines of baseline CST, and `CURRENT_TIMESTAMP` parsed as a plain identifier. No production accepted SQL's reserved nullary functions; added `SpecialFuncExpr`.
- **`IsClause` had a bare `NotKW NullKW` alternative**, so `DEFAULT true NOT NULL` parsed as `DEFAULT (true NOT NULL)` and the column **silently stayed nullable** — while `NOT NULL DEFAULT true` was fine. PostgreSQL has no postfix `expr NOT NULL`.
- **Statement counting was wrong in both directions** — 0.6.0 reported a constant 2 per file (68 total); the truth is 148 against 150 semicolons.

## Coverage

The old corpus was 34 DDL files. It contained 6 SELECTs and still had **zero** `TargetElem`, `ColLabel`, `WindowSpec`, `JoinClause`, `CaseExpr` — structurally blind to every SELECT-side change made here. Presence of a statement type is not coverage of it.

This adds **191 statements harvested from real SQL string literals** in `aether/` and `examples/` — queries the codebase actually issues, not hand-written fixtures, so the corpus stays independently motivated and exercises shapes nobody thought to assert.

    TargetElem  0 -> 47      SelectStmt 6 -> 74      UpdateStmt 0 -> 10
    ColLabel    0 ->  2      InsertStmt 6 -> 53      DeleteStmt 0 ->  9

`CorpusParseTest` adds two sensors, each corresponding to a defect that shipped past a fully green suite, and **both mutation-checked** (removing the `_ROOT` unwrap → red; making comments non-trivia → red):

- every `.sql` in the repo parses — the `--`-not-trivia bug made 18 of 34 files unparseable while every unit test stayed green
- one statement per corpus line — catches the `_ROOT` collapse

A byte-exact CST dump is deliberately **not** checked in: at 2 MB it regenerates wholesale on any CST change and would be re-baselined mechanically the first time it went red. The differential remains a manual procedure, documented in the test.

## Known gap

Nested block comments (`/* outer /* inner */ still */`) are unsupported. PostgreSQL nests them per the SQL standard, but nested comments are not a regular language, so a rule compiled into peglib's DFA lexer cannot express it — same class as the `DollarString` back-reference. Documented at the rule; needs upstream support. **Found by harvesting real SQL** — no hand-written fixture had exercised it.

## Trade-off accepted

`$` removed from the identifier continuation class. PostgreSQL permits it after the first character, so this narrows the language — but with `$` allowed, `world$$` in `$$hello world$$` lexes as one 7-char identifier by maximal munch and dollar quoting cannot work at all. Measured: **0 occurrences** of `$`-containing identifiers across every `.sql` and every SQL string literal in the repo.


siy (commented):
Reviewed from the aether-main stream. **The migration looks right, and the red CI is the fixture being wrong rather than the code.** Details below, plus one correction to the PR body.

## The `forge-tests` failure is a fourth latent bug, not a regression

CI fails on:

```
[ERROR] jOOQ XML schema differs from tracked file:
  Line 596:  -<delete_rule>NO ACTION</delete_rule>   +<delete_rule>CASCADE</delete_rule>
  Line 670:  -<is_unique>YES</is_unique>             +<is_unique>NO</is_unique>
  Line 679:  -<is_unique>YES</is_unique>             +<is_unique>NO</is_unique>
```

I checked all three against the source DDL. **The new output is correct in every case; the tracked baseline encodes old-parser bugs.**

| `expected-jooq-schema.xml` says | source DDL | correct |
|---|---|---|
| `fk_order_lines_order` → `NO ACTION` | `V002__orders.sql:19` — `... REFERENCES orders(id) ON DELETE CASCADE` | **CASCADE** |
| `idx_products_category` → `is_unique YES` | `V001__catalog.sql:35` — `CREATE INDEX` | **NO** |
| `idx_products_metadata` → `is_unique YES` | `V001__catalog.sql:36` — `CREATE INDEX ... USING gin` | **NO** |

The showcase schema contains **three `CREATE INDEX` and zero `CREATE UNIQUE INDEX`**, yet the baseline marks two of three unique. The old extractor was not near-missing, it was inventing uniqueness.

Both defects point the same way and are user-visible in a worse way than the three grammar bugs already documented: they corrupt EXTRACTED SCHEMA rather than parse structure. Anything reading that XML — generated code, a human checking constraints — would believe deletes do not cascade when they do, and that two ordinary indexes enforce uniqueness when they do not.

**Action: regenerate `examples/jooq-xml-showcase/expected-jooq-schema.xml`.** That should turn CI green on its own.

## Two notes on the PR body

1. **The result table currently reads as green-modulo-Docker, and it is not.** It reports 813/0 and attributes the only red to two `StatementSplitter*DiffTest` needing Docker. Neither `forge-tests` nor `build-and-test` is passing, and this jOOQ failure is a different thing entirely. Worth correcting, since the table is what a reader will trust.

2. **These three belong in "latent grammar bugs, exposed by the stricter toolchain"** as a fourth entry. Framed as a CI failure they look like fallout; they are a headline result of the migration — the same "0.6.0 concealed it" story, at the schema-extraction layer.

## The formatter fix

`FlowPrinter` + `AnnotationCommentPreservationTest`: I hit this bug independently a few hours ago in the aether-main stream and had a report drafted; this fix is more complete than what I would have filed, so I am dropping mine.

For the record, my reproducer agrees with yours exactly — **all four comment styles are dropped**, not only `///`:

```
MARK-TRIPLE-SLASH  before=1 after=0
MARK-DOUBLE-SLASH  before=1 after=0
MARK-BLOCK         before=1 after=0
MARK-JAVADOC       before=1 after=0
```

Your test covers all four plus methods, stacked annotations, idempotence, and the above-annotation position as a regression guard, which is the right set. One observation worth keeping with it: this was invisible to review because a formatting diff is normally read with whitespace ignored, and **`git diff -w` does not distinguish a deleted comment from a re-wrapped one**. I only caught it by comparing non-whitespace content hashes of every touched file against `HEAD`. That check is cheap and might be worth adding wherever the formatter runs in bulk.

The comment that exposed it on my side (`ClusterBootstrapCommand.clusterNameOverride` — why that field is deliberately NOT a value object) is restored above the annotation, so it survives regardless of merge order.

## Merge-order note

No conflict expected with the aether-main stream: this PR touches `pg-tools`, `jbct-format` and docs, while the value-object work in flight touches `environment-integration`, `aether-config`, `aether-deployment`, `cli`, `node` and the providers. The only shared file is `aether/docs/reference/feature-catalog.md` (1 line here).

Issue #619, opened 2026-08-19
pg-parser: nested block comments are unsupported (needs a counting scanner upstream)


siy:
## Summary

`pg-parser` does not support **nested block comments**. PostgreSQL nests `/* ... */` per the SQL standard; our grammar stops at the first `*/` and mis-parses the remainder.

```sql
-- PostgreSQL: the whole comment closes at the final */, statement is `SELECT 1 AS c`
SELECT 1 /* outer /* inner */ still a comment */ AS c;
```

Observed: `trailing input not consumed` — the parser closes the comment at the inner `*/` and then chokes on `still a comment */ AS c`.

## Why it cannot be fixed in the grammar

```peg
BlockComment <- '/*' (!'*/' .)* '*/'
```

The obvious repair is recursion — `BlockComment <- '/*' (BlockComment / !'*/' .)* '*/'` — but `BlockComment` is named from `%whitespace`, and under peglib 0.7.x every `%whitespace` alternative is compiled into the **DFA lexer**. Nested comments are not a regular language, so no DFA can match them. Pinning the rule to PARSER doesn't help either: the parser runs over tokens, and by then the lexer has already mis-split the comment.

This is the same class of limitation as the `DollarString` back-reference — it needs a **counting scanner** emitted for that token kind in peglib, not a grammar change on our side.

## Scope of impact

- A migration or `@PgSql` query containing nested block comments fails to parse.
- Single-level `/* ... */` and `--` line comments are unaffected.
- Rare in practice — it went unnoticed until the parser corpus was extended with real harvested SQL (#618). No hand-written fixture had ever exercised it.

## Resolution options

1. **Upstream** — ask peglib for a counting scanner for delimited-with-nesting token kinds. Generalises: nested comments exist in Rust, Swift, Haskell, D, SQL. Structurally similar to the tag-delimited-block idea that came up during the 0.7.2 migration.
2. **Pre-pass** — strip nested comments before parsing. Cheap, but breaks span fidelity, which `pg-schema` and `pg-codegen` rely on for error positions.
3. **Accept and document** — current state, documented in [`pg-persistence-spec.md`](../blob/release-1.0.0-rc3/aether/docs/specs/pg-persistence-spec.md#sql-grammar-limitations), at the rule in `postgres.peg`, and in the feature catalog.

Recommend (1); (3) holds until then.

## Reproduce

[...]

PR #626, opened 2026-08-21
docs: nested block comments can silently parse as a different statement, not just fail (#619)

siy:
Re-verified #619 against peglib 0.7.2 and found the documented severity is wrong. Upstream request now filed as **siy/java-peglib#45**.

## The correction

Three places describe the nested-block-comment gap as "fails to parse". It does not reliably fail — **it can silently accept a different statement.**

The lexer closes at the inner `*/` and the remainder leaks into the statement as live SQL. When that leaked text composes into something valid, there is no diagnostic at all. Both of these parse cleanly with **2** `TargetElem` items where correct nesting demands **1**:

```sql
SELECT 1 /* /* */ , 999 -- */
 FROM t;                        -- intended: SELECT 1 FROM t   | actual: SELECT 1, 999 FROM t

SELECT 1 /* a /* b */ -- */
 , 2 FROM t;                    -- intended: SELECT 1 FROM t   | actual: SELECT 1, 2 FROM t
```

The trailing `-- */` reads as a line comment and swallows the orphaned outer `*/`, leaving nothing to trip the parser.

Even the loud cases mislead. `SELECT 1 /* outer /* inner */ still a comment */ AS c;` reports `expected end of input at 1:37` — because the parser first *accepts* the truncated `SELECT 1 still` (reading `still` as an implicit column alias) and only then chokes. The error points several columns past the real cause and never mentions comments.

The ticket's "fails to parse, rare in practice" framing held only for the inputs it happened to try. Rarity is the only thing bounding the blast radius.

## Changes

- `pg-persistence-spec.md`, `feature-catalog.md`, and the note at the rule in `postgres.peg` — severity corrected, with the reproducing SQL
- `NestedBlockCommentGapTest` — a **tripwire**, plus a single-level control so the assertions are about nesting and not block comments in general
- Stale `-Pgenerate-parser` instruction removed from `CorpusParseTest`'s doc comment (that profile was retired in `03d547e26`); same correction posted to #619, whose repro block had it too

## Why the test pins the wrong behaviour on purpose

`NestedBlockCommentGapTest` asserts what ships today, including the silent divergence. A `@Disabled` test with the correct expectation would document intent but never fire — it cannot tell you the day the gap closes. An assertion on current behaviour goes **red** when peglib grows a counting scanner, which is exactly the signal to delete the file, re-add the two excluded statements to `corpus/dml-select.sql`, and revert these doc edits. The test says so in its own doc comment.

## Why there is no grammar-side fix

Confirmed against peglib's own docs rather than inferred: `%whitespace` alternatives are absorbed into the DFA at a trivia kind keyed off the leading literal, and the recursive repair `BlockComment <- '/*' (BlockComment / !'*/' .)* '*/'` is **refused outright by peglib's analyzer** as `grammar.whitespace-cycle` — `%whitespace` may not transitively reference itself. `%parser` does not help either, since trivia is consumed during lexing. Hence the upstream ticket.

## Verification

pg-parser **314 tests, 0 failures** (including the 3 new); pg-schema **112, 0 failures**. Both built in one reactor — never pg-schema alone, which would resolve a stale pg-parser from `~/.m2`.


siy:
**Correction — one of the two examples above is wrong.**

`SELECT 1 /* a /* b */ -- */` + newline + ` , 2 FROM t;` is **not** a divergence. That comment is *balanced* — two opens (`/*`, `/*`), two closes (`*/`, `*/`) — so the span ends at the final `*/` and `, 2` sits legitimately **outside** it. Two select-list items is the correct reading, and the pre-`%nest` lexer produced two as well: its early close left `-- */` as a line comment which swallowed the orphaned `*/`, landing on the same answer by a different route. Same output, so nothing diverged.

Only the other example is a genuine silent divergence:

```sql
SELECT 1 /* /* */ , 999 -- */
 FROM t;
```

Here `, 999` sits **inside** the balanced span. Correct reading is `SELECT 1 FROM t` — one item. The old lexer read `SELECT 1, 999 FROM t` — two items, no diagnostic. Verified both before and after: 2 items on peglib 0.7.2, 1 item on 0.7.3 with `%nest`.

The finding stands on that one case; I over-generalised from it to a second example without checking the delimiter counts, and a verification agent accepted my framing instead of testing it. The rule worth carrying: **count the delimiters before deciding what the right answer is** — a balanced nested comment is exactly the case where the buggy and correct lexers agree.

PR #641, opened 2026-08-26
fix: slice-processor errors on non-Promise dependency methods instead of silently dropping or misreading them

### LINKED ISSUE #612: Generated dependency proxies silently drop non-Promise methods: extractPromiseTypeArg yields empty and onPresent skips, with no diagnostic
Source: design review of Aether at 1.0.0-rc3 (2026-08-17). The review flagged this one as agent-sourced and asked that it be confirmed before acting — **confirmed 2026-08-18**, mechanism below.

## What

`jbct/slice-processor/.../generator/FactoryClassGenerator.java` (~:995-1010) builds the method list for a generated dependency proxy:

```java
if (!method.getModifiers().contains(Modifier.STATIC) && !method.getModifiers().contains(Modifier.DEFAULT)) {
    extractPromiseTypeArg(method.getReturnType()).map(responseType -> toProxyMethodInfo(method, responseType))
                                                 .onPresent(methods::add);
}
```

If the return type is not a `Promise`, `extractPromiseTypeArg` yields empty, `onPresent` does nothing, and **the method is silently omitted from the generated proxy**. No error, no warning.

## Why it matters

A slice interface method that returns `Result<T>`, `Option<T>` or a bare `T` — all legal JBCT return shapes — disappears from the proxy. The caller then fails at runtime, or binds to something unexpected, with nothing at compile time pointing at the cause.

This is the same silent-wrong-behaviour class the runtime refuses elsewhere: the slice-processor already errors on codec tag collisions and on unregistered resources rather than degrading.

## What to do

Make it a compile error naming the offending method and its return type, with the rule stated ("a slice dependency method must return `Promise<T>`; local shapes belong on a step or leaf").

If non-`Promise` methods on a slice interface are legitimate in some case, that case should be named and the diagnostic scoped to the rest — but silence is not the right answer either way.

## Acceptance

- A slice interface with a non-`Promise` method fails the build with a diagnostic naming the method.
- Mutation check: remove the diagnostic and confirm the new test goes red.

siy:
Fixes #612 — and the defect turned out to be double, not single.

## What the issue saw, and what was actually there

The issue reported non-`Promise` dependency methods being **silently dropped** from generated proxies. True for non-generic returns — but `extractPromiseTypeArg` never checked the type *is* `Promise` at all: it took the first type argument of **any** generic declared type. So the other half was worse than a drop: a dependency method returning `Result<T>` or `Option<T>` was silently **misread as `Promise<T>`** and generated a wrong-shaped proxy method.

## The fix

- `extractPromiseTypeArg` checks the erasure against `org.pragmatica.lang.Promise` — only a genuine Promise extracts.
- Every abstract instance method of a dependency interface that fails the check reports a **compile ERROR anchored on the method element**, stating the rule: *a slice dependency method must return `Promise<T>` — slice-to-slice calls are remote-capable, and local shapes (`Result`, `Option`, bare values) belong on a step or leaf.* Same refuse-loudly policy as codec tag collisions and unregistered resources — the precedent the issue itself cites.
- `default` and `static` members stay exempt (not proxy candidates) — pinned in-test by placing both on the erroring interface and asserting the diagnostic names only the abstract offender.

## Acceptance

- `should_fail_on_non_promise_dependency_method` — bare `String` return: build fails, diagnostic names `external.InventoryService.checkStockSync returns java.lang.String`. Removing the diagnostic makes `hadErrorContaining` go red (the mutation check).
- `should_fail_on_result_returning_dependency_method` — the misclassification half: `Result<Integer> quote(...)` fails with the same rule text instead of generating a wrong proxy.

## No `ENVELOPE_FORMAT_VERSION` bump — reasoning on the record

Invariant 3 keys on the *emitted* factory/manifest structure. This change adds an error path and tightens which methods generate; every slice that compiles produces byte-identical envelope structure to before. Verified by the full suites: **slice-processor 303 + slice-processor-tests 46 green, zero existing slices refused** — Promise-only was already the de-facto contract; now it is the enforced one.


siy:
Review findings, none blocking merge:

1. The tests asserting the new diagnostic use `hadErrorContaining` without pinning the total error count. A spurious extra diagnostic emitted on a default or static interface method alongside the expected one would still pass. Worth adding an exact-count assertion (e.g. `hadErrorCount(1)`) so an unrelated regression on those member kinds doesn't slip through unnoticed.

2. Dependency methods inherited from a super-interface of the dependency interface remain invisible to the processor — silently dropped, no diagnostic. This is pre-existing behavior (predates this PR) and a sibling of the #612 silence family. Filed as a follow-up: #663.

3. The error wording for a raw (non-parameterized) `Promise` return type is slightly confusing — worth a pass for clarity, non-blocking.

Issue #606, opened 2026-08-18
Three examples teach patterns that do not run or are wrong: banking never uses its injected DB, transfer compensation drops its Promise


siy:
Source: design review of Aether at 1.0.0-rc3 (2026-08-17). Every claim below re-verified against `release-1.0.0-rc3` on 2026-08-18; two are worse than the review recorded.

Examples are the primary evaluation surface and the thing adopters copy. Three of them teach patterns that do not run, or are wrong.

## (a) Banking does not persist — and the injected database is never used at all

`examples/banking/account/.../AccountService.java:82-86` injects `@Sql SqlConnector db`, then stores state in two `ConcurrentHashMap`s.

**Stronger than reported:** `grep -n "db\."` over that file returns **only the two import lines**. The record component `db` is threaded through the factory and never touched. The example demonstrates injecting a database it does not use.

A reviewer asking "does Aether do real persistence?" concludes wrongly in either direction.

## (b) Transfer compensation discards its own result

`examples/banking/transfer/.../TransferService.java:167`:

```java
private void compensateDebit(TransferContext ctx) {
    accounts.credit(ctx.accounts().source().id(), ctx.sourceAmount());   // Promise -> dropped
    recordTransfer(ctx, TransferStatus.COMPENSATED);
}
```

`credit(...)` returns a `Promise` that is never awaited, composed, or inspected. Three things follow:

1. the compensation is not durable (node dies between debit and compensation → money is lost);
2. the compensation's **own failure is silently swallowed**;
3. the transfer is recorded `COMPENSATED` **whether or not the compensating credit succeeded**.

This ships as the reference pattern for moving money.

## (c) The linter cannot catch (b) — worth knowing before relying on it

`JBCT-RET-07` (`CstDiscardedResultRule`) exists for exactly this; its own doc says *"Discarding their result is always a bug on Result / Promise because errors get swallowed."*

It does not fire here. Detection is syntactic — expression statements ending in a *known monadic terminal method*, plus static factory calls on `Result`/`Promise`/`Option`. `accounts.credit(id, amount);` is a plain instance call whose **return type** is `Promise`, and a syntax-only linter has no type resolution to know that. Linting the file reports one error (`JBCT-RET-04`, Void vs Unit) and misses the money.

Not necessarily a rule bug — closing it needs type information the linter deliberately does not have — but it means "the linter would have caught it" is false for this whole class.

## (d) url-shortener ships a validating factory nothing calls

[...]

PR #643, opened 2026-08-26
fix: generated routes construct request records through their validating factory — failure is a typed 400

### LINKED ISSUE #605: HTTP request bodies bypass domain validation: the generator validates path/query value objects and constructs body records with the canonical constructor
Source: design review of Aether at 1.0.0-rc3 (2026-08-17), mechanism re-verified 2026-08-18.

One of three findings that share a root cause — see the "developer-facing path is unverified" note at the bottom.

## What

`RouteSourceGenerator` binds parameters two different ways depending on where they come from.

**Path/query** (`:1312-1318`) composes the value object's own validation into the parser:

```java
"PathParameter." + parser + "().mapped(" + binding.voQualifiedName() + ".valueMapping().lift())"
```

A bad path segment fails before the handler runs and becomes a typed error.

**Body** (`:1168`, `:982`) calls the canonical constructor directly:

```java
".to((" + handlerParams + ") -> delegate." + method.name() + "(new " + parameterType + "(" + constructorArgs + ")))"
```

A record's canonical constructor cannot return `Result`, so **validation expressed as a static factory is structurally unreachable on this path**.

The generator states this deliberately (`:704-707`): *"body-bound value-object fields are handled by the JSON layer and are never validated here."* That is a coherent boundary — but nothing downstream re-establishes it.

## Consequence, in the flagship example

`examples/url-shortener-v2/.../UrlShortener.java:29` — `ShortenRequest.shortenRequest()` performs null / blank / URL-pattern checks returning `Result<ShortenRequest>`.

**Verified: its only callers are in `UrlShortenerTest`.** `POST /` reaches `shorten(ShortenRequest)` with whatever JSON deserialization produced. The example ships a validating factory that production traffic never touches, and its unit tests pass because they call the factory the production path skips.

## Why this is the sharpest of the three

It is a correctness footgun for every adopter who copies the example: the code looks right, the tests are green, and the validation silently never runs.

## Decide, then make it consistent

- **(a) Generator learns the factory.** When a body record declares a static factory returning `Result<Self>`, call it and map failure to a typed 400. Closes the gap by construction; costs a generator change and a stated convention.
- **(b) Keep the JSON-layer boundary and teach validate-in-method-body.** Then the examples and the book must show it, and `url-shortener` must stop shipping a factory nothing calls.

Do not leave it as-is.

## Acceptance

- A body record with a failing factory yields a typed 4xx, not a constructed invalid record (if (a)); or the examples and docs demonstrate the in-method pattern and no example ships an uncalled factory (if (b)).
- Mutation check: make the factory always fail and confirm a request is rejected.

Related: #407 (ValueMapping composite form). Siblings in this cluster: examples-teach-wrong-patterns, testkit-cannot-test-slice-to-slice.

siy:
Fixes #605 — owner decision **(a): the generator learns the factory**.

## The rule (stated once, on the detection helper, applied everywhere)

Wherever a generated route constructs the request record, it constructs **through the record's declared validating factory** when one exists — mapping failure to the same typed 400 that path/query parameter validation already produces:

```java
Type.factory(args).mapError(cause -> HttpStatus.BAD_REQUEST.with(cause)).async().flatMap(__validated -> delegate.method(__validated))
```

— and keeps the canonical-constructor path **byte-identical** when none does. Pure-body routes (Jackson already built the record) re-validate by decomposing through the record accessors into the factory. Applied at every construction site: all eight route shapes, security and plain branches — including path-only/query-only constructions, where record-level cross-field validation was equally unreachable (consistency extension of the issue's scope, stated in the doc comment).

**Detection convention:** a static method returning `Result<Self>` whose parameter types match the canonical components in order. Duplicates: first declared wins, WARNING names both.

## Evidence

- **No fallout, proven not assumed:** all five pre-existing generated Routes files (testslice, voslice, hostslice, collisionslice, versionedslice) are `diff`-identical to a pre-change baseline.
- **Mutation pin:** forcing detection to `none` turns 6 of 11 new assertions red; the 4 no-factory controls stay green (they must — that's the byte-identity half).
- **Duplicate-factory path exercised:** the WARNING fires naming both candidates; the first wins.
- New `factoryslice` fixture (slice-processor-tests) covers: body-with-factory, body-without (decoy non-matching factory — canonical path untouched), merged path+body, path-only.
- slice-processor **301** + slice-processor-tests **57** green — verified twice (implementing agent + independent re-run).

## Deviations from the working plan, on the record

- The conditional `HttpStatus` import was unnecessary: `errorMapper()` already imports it unconditionally in every generated Routes file.
- Tests live in slice-processor-tests, not compile-testing: `loadRouteConfig` resolves `routes.toml` via `Filer.getResource → Path.of(uri)`, which cannot work against compile-testing's in-memory file manager — the existing route tests live there for the same reason, and the pattern is stronger (asserts the text *and* compiles it).

## No `ENVELOPE_FORMAT_VERSION` bump

Routes-source change only; manifest/factory structure untouched (invariant 3 keys on those generators' output).

## Downstream

On merge, `url-shortener-v2`'s `ShortenRequest.shortenRequest()` validation goes **live on `POST /`** — the flagship example stops shipping dead validation. That example-level story (and its siblings) is #606, next in this batch.


siy:
Review findings:

Factory detection compares parameter and record-component types by comparing `TypeMirror.toString()` textually, and silently falls back to the canonical constructor on any mismatch. A near-miss factory — for example one whose parameter carries an annotation, or a generic record component spelled slightly differently — loses validation again with no signal that anything went wrong. That's the exact failure mode this PR fixes, reintroduced through the fallback path. A "factory found but unmatched" WARNING at the call site would close this gap. Filed as follow-up: #662.

Minor, non-blocking: "first declared wins" when resolving among multiple candidate factories relies on `getEnclosedElements()` preserving declaration order, which the `javax.lang.model` API doesn't guarantee.

Issue #648, opened 2026-08-26
pg-codegen: parameter-order/placeholder-order mismatch surfaces as a 'does not override' error in generated code instead of a diagnostic at the method


siy:
Found while rewriting `examples/banking/account` onto `@PgSql` (#606 implementation): the generator binds method parameters to SQL placeholders in **SQL positional order, not method declaration order**, and when the two disagree the failure surfaces as a **"does not override"* compile error inside generated code** — an error about a symptom three steps removed from the cause.

## What happened

A method whose declaration order was `(accountId, amount)` against SQL whose placeholders appeared amount-first produced generated code whose method signature no longer matched the interface — javac reports the generated class "does not override" the interface method. Nothing points at the actual mismatch (parameter order vs placeholder order), and the generated file is the last place a slice author looks.

## Why it matters

Same diagnosability class as #612 (silent proxy drops) and #646 (silent validation skips): the contract exists but its violation is reported far from its cause, or not at all. The processor already knows both orders at generation time — the check is cheap and local.

## What to do (either closes it)

- **Diagnose at the source:** when declaration order and placeholder order disagree, emit a compile ERROR on the @Query method element naming both orders — "parameter `amount` is declared second but bound first; @PgSql binds in SQL placeholder order".
- **Or bind by name** where placeholder syntax allows it, making declaration order irrelevant — bigger change, only if the placeholder syntax already carries names.

## Acceptance

- A deliberately reordered pair produces a diagnostic naming the two orders, anchored on the method, with no "does not override" noise.
- The banking example's AccountPersistence (post-#606) compiles unchanged.

siy:
[...]

PR #650, opened 2026-08-26
fix: examples teach what runs — banking persists through PgSql, compensation failure is loud, every example states its scope

### LINKED ISSUE #606: Three examples teach patterns that do not run or are wrong: banking never uses its injected DB, transfer compensation drops its Promise

siy:
Fixes #606 — owner decision: **banking persists for real**, not a toy disclaimer.

## (a) Banking → `@PgSql`

`AccountService` injected a `SqlConnector` it never touched (`grep "db\."` matched only the imports) while keeping state in ConcurrentHashMaps. Now: a new `@PgSql AccountPersistence` adapter, every statement compile-time-validated against `schema/V001__create_tables.sql` (the processor rejected the first draft — the validation is real). Credit/debit are single conditional `UPDATE … WHERE … RETURNING` statements — insufficient-funds is decided by the row match, atomically; rejected updates map to the existing typed errors. Tests follow the house pattern (in-memory persistence double), 14 green.

**Honest boundary, recorded at the declaration instead of hidden:** the processor rejects data-modifying CTEs and `@PgSql` exposes no transaction surface, so `openAccount` is two statements — a crash between them orphans an account row. That limitation is a real DX finding, not this example's to solve.

## (b) Compensation composes; its failure is loud

The reference pattern for moving money dropped its own compensation Promise. Combinator choice, reasoned not defaulted: **`fold`** — Promise has no error-path `flatMap`; `onFailure` is an unawaited side effect that cannot gate a chain, and `orElse` fails twice (drops the original cause, and would make the *transfer* succeed when the *compensation* succeeded). Semantics now: original failure propagates only after the compensating credit completes; `COMPENSATED` only on success; new `TransferStatus.COMPENSATION_FAILED` carries **both causes** in `TransferSummary.failureDetail` (a recorded field is assertable; the examples have no logging idiom to borrow).

**Mutation-proven both ways:** record-COMPENSATED-unconditionally → 2 tests red; restore the original fire-and-forget → 3 tests red. That satisfies the issue's mutation-check acceptance literally.

## (b′) The same bug, found and fixed in ecommerce

The issue's acceptance says "no example discards a Promise/Result" — and the implementation sweep found `PlaceOrder.releaseStockOnFailure` dropping its release Promise identically. Fixed to the same standard: `fold`, `OrderError.StockReleaseFailed(paymentFailure, releaseFailure)` naming both causes, stub-based test pair, mutation-checked (restoring the drop → 2 red).

## (c) Teach-headers

Every slice example now opens with what it demonstrates and what it deliberately does not (catalog already had one). Banking's states: saga/compensation across slices AND PgSql persistence.

## Verification

- `mvn -f examples/pom.xml -o clean test`: **60 tests green** (was 25) — 14 account, 15 transfer, 6 place-order among the new.
- `mvn -f examples/pom.xml -o jbct:check`: **clean**.
- Banking-only lint delta vs HEAD: 41 → 38; the single added finding is CAUSE-02 on the new error record, matching its three pre-existing siblings (typed-error migration territory, not this PR's).

## Filed en route

- #648 — pg-codegen binds parameters in SQL positional order; a mismatch surfaces as a "does not override" error in generated code, three steps from the cause.
- `examples/ecommerce/README.md` is stale (describes a module that isn't there) — header went into `PlaceOrder.java` rather than adding correct text next to wrong text.


siy:
Review findings:

1. `classifyRejection` can label a rejected credit as `InsufficientFunds`. This is reachable via a race through the shared `confirmChange` path, not just a mislabeling on the happy path — worth tightening the classification so a credit rejection can't surface under a debit-only reason.

2. The teach-headers cover entry slices only, while the CHANGELOG states "every slice example" is covered. Either extend the headers to the remaining slices or narrow the CHANGELOG wording to match what's actually covered.

3. `TransferStatus.FAILED` is unused and undocumented — either wire it up where a transfer can actually land in that state, or document why it currently can't and remains reserved.

Issue #651, opened 2026-08-26
pg-schema: correlated subqueries cannot see the outer statement's scope — legal EXISTS guards report 'Table or alias not found'


siy:
Found during the #649/#646 fix (pre-existing — verified byte-identical behavior before and after that change, so it is NOT part of that arc): **correlated subqueries never see the outer statement's scope.**

## Repro

```sql
UPDATE reservations SET state = 'released'
WHERE EXISTS (SELECT 1 FROM bookings b WHERE b.reservation_claim_id = reservations.claim_id)
```

reports `Table or alias not found: reservations` — the outer target is not visible inside the subquery's scope, so every correlated reference is a false error. The same applies to correlated subqueries under SELECT, not just DML.

## Why it matters

Correlated `EXISTS`/`NOT IN` predicates are the standard shape for existence guards; the ticketing corpus's reap/guard statements use them (its `expireOrphanedConfirmations` carries a `NOT IN (SELECT …)` today — uncorrelated, but one refactor away). Any adopter writing a correlated guard gets a hard `[PG-VALIDATE]` error on legal SQL.

## Fix direction

Subquery scope construction should CHAIN to the enclosing scope (inner resolution first, outer as fallback) rather than starting empty — the standard lexical-scope model. The #649 fix already made statement-level scopes explicit objects, which is the natural place to hang the parent link.

## Acceptance

- The repro validates clean; `b.nonexistent` inside the subquery still errors; an unqualified ambiguous name (present in both scopes) resolves inner-first.
- SELECT-side correlated subqueries validate clean too (the defect is not DML-specific).

siy:
[...]

Issue #646, opened 2026-08-26
pg-codegen: RETURNING validated against a WHERE-subquery's SELECT scope — and never validated at all without one


siy:
Reported by the ticketing session with an A/B minimal repro (`ticketing/.../booking/BookingStore.java:185` vs `:203`); mechanism confirmed by code read in `pg-schema`/`pg-codegen` — this is written from the source, not the symptom.

## Symptom

`UPDATE reservations … WHERE … AND claim_id NOT IN (SELECT reservation_claim_id FROM bookings …) RETURNING seat_id, event_id, version` → three spurious `[PG-VALIDATE] no matching SELECT column` warnings, one per row-record component. The byte-identical statement **without** the subquery is silent. Build green, generated mapper correct — pure warning noise, but it trains people to ignore the one channel that catches real schema drift.

## Mechanism (confirmed)

`QueryValidator.selectOutputColumnNames` (pg-schema `validator/QueryValidator.java:81-87`):

```java
var cores = queryRoot.findAll("SelectCore");
if (cores.size() != 1) return Option.empty();
return outputColumnNames(cores.getFirst());
```

`findAll` walks the **whole statement tree**. Consequences, each following directly:

1. **UPDATE…RETURNING, no subquery** → zero SelectCores → `empty` → caller (`QueryAnnotationProcessor.java:913-914`) skips the check. The reporter's A-case is silent **because RETURNING mapping is never validated there at all** — not because it validates. The "working" case is a false negative, worth knowing before trusting the sweep's silence.
2. **One subquery anywhere in the statement** → exactly one SelectCore → the subquery's target list becomes "the query's output columns", and RETURNING is validated against it. `bookings.reservation_claim_id` matches none of `seat_id, event_id, version` — exactly three warnings.
3. **Two subqueries** → size 2 → skipped again (silent).
4. **INSERT … SELECT … RETURNING** → the source SELECT's core is "the" core; RETURNING misvalidates against the source projection.

This is the same select-by-position-not-by-tree-walk lesson the 0.7.2 migration recorded for the parser ("never identify by rule name under fallback; select by POSITION"), and the same subtree-attribution family the CAUSE-pack arc fixed in five other places: a nested scope's nodes attributed to the enclosing statement.

## Fix direction

[...]

PR #652, opened 2026-08-26
fix: pg validation scopes names by statement structure — EXCLUDED and self-reference in DO UPDATE, RETURNING/WHERE against the target

### LINKED ISSUE #649: pg-codegen hard-fails ON CONFLICT DO UPDATE: EXCLUDED read as a column, target-table qualifier consumed as a column — blocks the ticketing build
**Build-blocker on the ticketing corpus** (reported by that session; verified pre-existing against bare HEAD `810d2d9` with stash/restore — toolchain, not their diff). Four compile-time `[PG-VALIDATE]` ERRORS, zero workaround that preserves semantics.

## Failure

`INSERT … ON CONFLICT (key) DO UPDATE SET col = EXCLUDED.col … WHERE target.col < EXCLUDED.col` fails validation two ways:

- `Column 'excluded' not found` — **`EXCLUDED.x` read as a column reference, not a relation qualifier.** The pseudo-table is not registered in the DO UPDATE scope.
- `Column 'reservations' not found in table 'reservations'` — **the qualified self-reference's qualifier consumed as a column name** against its own table.

Representative statement (PricingStore.upsertCurrent):

```sql
INSERT INTO current_price (scope_key, event_id, tier, amount_minor, currency, version, updated_at)
VALUES (:scopeKey, :eventId, :tier, :amountMinor, :currency, :version, now())
ON CONFLICT (scope_key) DO UPDATE SET amount_minor = EXCLUDED.amount_minor,
currency = EXCLUDED.currency, version = EXCLUDED.version, updated_at = now()
WHERE current_price.version < EXCLUDED.version
```

Four identical-shape sites (BookingStore.claimSeat:104, SeatProjectionStore.upsertStatus:44, PricingStore.upsertCurrent:39, PriceProjectionStore.upsertPrice:24) — and they are the system's **monotonic version guards**: every `WHERE target.version < EXCLUDED.version` ordering guarantee (seat convergence, both price projections, the contended-seat claim). Dropping qualification changes which row the predicate reads; splitting the upsert reintroduces the race. **No clean local escape — this blocks the corpus build outright.**

## Regression window and prime suspect

Installed pg-codegen: rc3 built **2026-08-24 22:55**; the corpus was green on the Aug 22-23 toolchain. What changed in pg-tools in that window is the **peglib 0.7.3 migration + `postgres.peg` `%nest`**. The 0.7.3 migration's own recorded hazard is the exact shape of this failure: *literals that were anonymous inline tokens resolve to named keyword kinds after the bump, and "do not dispatch on keyword kinds" was the 0.7.2 lesson* — if upsert-scope resolution identifies `EXCLUDED` or qualifier structure through token kinds or identifier-fallback kinds, the kind shift breaks it silently. The Java-side 0.7.3 bump was gated by a 2161-file differential corpus run; **whether the pg grammar's bump had an equivalent differential over upsert shapes is exactly the question** — presence of INSERT in a corpus is not coverage of ON CONFLICT DO UPDATE (the #618 lesson).

First diagnostic step: parse the representative statement with the current pg-parser and dump the CST around `OnConflictClause` — if `EXCLUDED` arrives under a different rule/kind than the validator expects, it's the 0.7.3 kind shift; if the CST is fine, the defect is in the validator's scope construction and likely predates the window (latent, newly reached).

## Relation to #646

Same family — names resolved against a scope chosen by position rather than statement structure — and the #646 fix direction (resolve against the statement's target relation, positional core discovery) is necessary but NOT sufficient here: the DO UPDATE clause additionally needs `EXCLUDED` registered as an in-scope pseudo-relation, and the target relation must be self-referencable by name. Fixing #646 without this leaves the four hard errors standing.

## Acceptance

- The representative statement validates clean; all four reporter sites compile (reporter will confirm in one run — corpus offer stands).
- `EXCLUDED.nonexistent_col` DOES error (the pseudo-relation is scoped, not whitelisted).
- A mis-qualified reference (`wrong_table.version` in the DO UPDATE WHERE) still errors.
- Differential pg corpus gains the four upsert shapes so the next parser bump cannot silently regress them.

siy:
Fixes #649 (build-blocker) and #646 — one arc, because they share a root: names resolved against scopes chosen by lexical discovery instead of statement structure.

## The #649 mechanism, traced not guessed

The reporter's corpus was green until the Aug-24 toolchain, yet the validator never had `EXCLUDED` support. Both true — the trace found **double mis-routing**:

1. `validateRoot` had a keyword-presence fallback: `UpdateKW`+`SetKW` anywhere with no `UpdateStmt` → run `validateUpdate` over the whole statement. Every `ON CONFLICT DO UPDATE` INSERT matches. So INSERTs were being validated as UPDATEs.
2. In `validateUpdate`, `item.findAll("ColId").getFirst()` picked the SET target. **peglib 0.7.3 lexes `version` as `Token VersionKW`** — the recorded "never dispatch on keyword kinds" hazard, sitting in the validator's extractor — so post-bump the keyword-spelled SET target was skipped and the RHS's first identifier (`EXCLUDED`, or the self-reference qualifier) was reported as a missing column. Pre-0.7.3, the same wrong path passed by accident.

That also explains the reporter's one-error-per-statement observation: each statement has exactly one keyword-spelled SET target — not first-error-stop, not contextual resolution. The peer session's "newly-reached, never-supported" hypothesis was confirmed in refined form: same code, wrong node.

## The fix

- Structural statement dispatch — the keyword fallback is gone.
- `ON CONFLICT DO UPDATE` gets a real scope: target relation **self-referencable by name** (what `WHERE current_price.version < EXCLUDED.version` needs), alias honored, `EXCLUDED` as a pseudo-relation carrying the target's columns — **scoped, not whitelisted**: `EXCLUDED.nonexistent_col` errors.
- #646: `RETURNING` + UPDATE/DELETE `WHERE`/`USING`/`FROM` validate against the statement's target scope; `selectOutputColumnNames` no longer discovers cores tree-wide (subquery-free statements were silently skipped; one-subquery statements validated RETURNING against the subquery's projection). Subqueries keep their own scopes.
- `CstExtractor.extractColumnList` selects by position, not the `ColId` rule name.

## Latent defects the newly-real validation surfaced — fixed in the same pass

- `WITH … AS (…) UPDATE` → "Table not found: id"
- `public.reservations` → "Table or alias not found: public"
- The test-persistence blueprint's `INSERT INTO kv_store (key, value)` was validated on **neither** column, and its upsert reproduced the #649 error verbatim.

One pre-existing defect found and deliberately not bundled: correlated subqueries cannot see the outer scope → **#651**, byte-identical before/after this change.

## Evidence

- pg-tools **832 → 858 green** (+26; QueryValidatorTest 50 tests / 8 nested classes) — implementer's run and an independent re-run.
- Mutation checks, run and restored: drop `EXCLUDED` registration → 4 red; stub `validateReturning` → 2 red (the reporter's canary); revert to whole-tree SelectCore discovery → 4 red.
- The four #649 upsert shapes + the #646 A/B pair are now parser-corpus statements — the next parser bump cannot silently regress them (the corpus contained `ON CONFLICT DO UPDATE` but not one `EXCLUDED.` reference; presence is not coverage).

## Post-merge

The reporting session (ticketing) will run its one-pass confirmation on the next toolchain install: all four stores compile, claimSeat clean, plus its SEQ-01 and CAUSE-pack censuses.


siy:
Review findings, neither blocking:

1. `BareRefPolicy.CHECK` on `UPDATE`/`DELETE ... WHERE` is fresh false-positive surface on unmodeled SQL constructs. This is mitigated by the decision to decline permissive CTEs, but the real sensor for whether this holds up is a downstream corpus run — worth watching that signal after this lands rather than assuming it's fully covered by test cases alone.

2. `resolveTableRef` / `extractAlias` still use `findAll` within a table-ref subtree. Pre-existing, narrow in practice, but worth a cleanup sweep alongside the next parser bump rather than carrying it forward indefinitely.

Issue #608, opened 2026-08-18
Inner development loop: every slice change costs a full mvn install, and the debug path is undocumented though it already works


siy:
Source: design review of Aether at 1.0.0-rc3 (2026-08-17).

## What

Every slice change is `mvn clean install` (two annotation processors + jbct plugin + blueprint packaging), after which Forge picks the artifact out of `~/.m2` by GAV coordinate (`examples/url-shortener/deploy-forge.sh`, `run-forge.sh`). No hot reload, no file watching, and no documented IDE debug workflow — `docs/slice-developers/forge-guide.md` contains no mention of debugging or remote attach.

## Why it matters

Measured against Spring Boot devtools' sub-second restart, this is the friction a developer feels fifty times a day, and developer experience is what makes defaults.

Forge itself is a genuine asset with no Spring equivalent — 5-node cluster, chaos injection, load generator, dashboard, one JVM. The artifact-repository indirection between edit and run is what blunts it.

## What to do

- **Watch mode** that rebuilds only the changed slice and redeploys it into a *running* Forge without the Maven round-trip. Even scoped to "one slice, no dependency changes, no schema changes" it removes most of the pain.
- **Document the debug workflow.** Forge is one plain JVM, so remote-attach works today and nobody is told. This part is nearly free.

## Acceptance

- Edit a slice method body, save, and observe the change served by a running Forge without `mvn install`.
- `forge-guide.md` documents attaching a debugger, with the exact JVM flags.

Cheapest half first: the debug documentation is a doc change and should not wait on watch mode.

siy:
First half (debug workflow) shipped as a PR: forge-guide.md Debugging section + FORGE_JVM_OPTS pass-through in all six run-forge.sh scripts. Watch mode remains open here — it needs a design pass (staleness guard vs envelope, classloader swap, scope limits); will bring a proposal rather than an implementation first.

siy:
[...]

PR #653, opened 2026-08-26
docs: Forge debug workflow — JDWP attach documented, run-forge.sh honors FORGE_JVM_OPTS

siy:
First half of #608 — the issue's own "cheapest half first": the debug workflow is a doc change and should not wait on watch mode.

## What lands

- **`forge-guide.md` gains a Debugging section.** Forge is one plain JVM — the whole 5-node cluster and every deployed slice — so standard JDWP remote attach has always worked; nobody was told. The section carries the exact flags (attach-when-ready `suspend=n` and debug-startup `suspend=y` variants), the IDE attach recipe, and the two field facts: breakpoints in slice code bind normally (JDWP is classloader-agnostic; keep the slice source project open for frame mapping), and a hit breakpoint freezes all five simulated nodes — expect SWIM suspicion noise after long pauses, harmless in Forge.
- **All six example `run-forge.sh` scripts honor `FORGE_JVM_OPTS`** — previously bare `exec java -jar` with no way to pass JVM flags. Deliberately unquoted (several flags must word-split); commented at the site. `bash -n` clean on all six.

## What deliberately does not land

Watch mode — rebuild-and-redeploy into a running Forge without the `mvn install` round trip — is the issue's real feature half and needs a design pass (staleness guard vs the envelope, classloader swap semantics, "one slice, no dependency changes" scoping). #608 stays open for it; acceptance for this PR is the issue's second bullet only ("forge-guide.md documents attaching a debugger, with the exact JVM flags").


siy:
Review findings:

The unquoted `${FORGE_JVM_OPTS:-}` expansion in `run-forge.sh` undergoes pathname expansion, and the documented recommended value contains a bare `*` (`address=*:5005`). It currently works only because the glob fails to match anything in the working directory and passes through unchanged — that's incidental, not a guarantee, and breaks under `failglob` or in a directory that happens to contain matching filenames. Quoting the expansion fixes it. Filed as follow-up: #664.

Also: the claim that the SWIM noise after long debugger pauses is "harmless in Forge" reads as an unverified failure-mode claim rather than something demonstrated — worth either verifying it or softening the wording to reflect that it hasn't been confirmed under an actual paused-then-resumed debug session.

Issue #655, opened 2026-08-26
Member-scan audit residue: ORD-01 ranks anonymous-class methods as outer members; isStepInterface suppressed by local records


siy:
The #645 fix shipped with an audit of every member-scan caller (`directlyEncloses` / `directMethods` / `directNestedTypes`) for the method-local-declaration hole. Two residual findings survived the audit and are NOT fixed in that PR — recorded here so they don't evaporate:

## 1. ORD-01 misattributes anonymous-class methods (reproduced)

An anonymous class body (`new Runnable() { public void run() {...} }` inside a member) has **no `TypeKind` node at all**, so its methods' nearest enclosing TYPE_KIND is the outer type — they rank as the OUTER type's members in `CstMemberOrderingRule` and can manufacture ordering inversions. Same family as #645's local-record bug (fifth-and-a-half sighting), different node shape: local types HAVE a TypeKind (guardable), anonymous classes have none (needs an enclosing-member walk instead). Reproduced during the #645 work.

## 2. `FileTypeClassifier.isStepInterface` suppressed by a local record (latent)

A local record declared inside a default method makes `isStepInterface` classify the file away from STEP_INTERFACE — a local declaration leaking into a type-level classification decision. Latent (no corpus hit yet), found by audit, worth a guard when touched.

## Also noted from the same audit, no action needed

UC-02's nested-Request check is LENIENT to local records (a local `record Request` satisfies it) — an FN, acceptable for a WARNING rule, recorded here for completeness.

## Acceptance

- An anonymous-class method inside a well-ordered type produces no ORD-01 finding; a genuinely misordered outer member still fires.
- A step interface with a local record in a default method still classifies STEP_INTERFACE.

Issue #645, opened 2026-08-26
JBCT-ORD-01 is unsatisfiable for the Aether slice implementation record — local records inside static factories rank as type members


siy:
Reported by the ticketing session (PFD/JBCT/Aether posterchild repo: 24 single-use-case slices, 214 tests, pinned 1.0.0-rc3), where this rule alone accounts for **24 of 53 suppression tokens** — every slice suppresses it.

## What

The Aether slice contract requires the implementation record to be declared **inside its own static factory method**:

```java
public interface AccountService {
    static AccountService accountService(...) {
        record accountService(...) implements AccountService { ... }   // the mandated shape
        return new accountService(...);
    }
}
```

`JBCT-ORD-01` ranks record declarations (rank 0) before factory methods (rank 3), so the required ordering **can never hold** — the record is lexically inside the factory, which ORD-01 reads as "record after method". A rule that is unsatisfiable on the framework's own mandated shape is a 100%-FP rule for every Aether slice.

## Suspected mechanism (verify, don't assume)

A **local record inside a method body** should arguably not be a member-ordering participant at all — if ORD-01's member scan picks it up, that is the subtree-attribution bug class again (a nested declaration attributed to the enclosing type's member list): the same defect family fixed four times in the CAUSE-pack arc (REC-01, obligations census, CAUSE-02 double emission, STY-04 sealed-utility). The fix would then be scoping the member walk to direct members, not a slice-idiom special case.

If ORD-01 instead sees the record because of how the CST nests slice-factory records, the narrower fix is the reporter's: a record nested inside a static factory method is not a type member and never ranks.

## Evidence available

The ticketing session offers a per-site classification of all 53 suppression tokens on request (24 ORD-01, 20 SEQ-01, 6 UC-02, 3 VO-01) — a measured suppression census on a conforming corpus, useful beyond this one rule.

## Acceptance

- The slice idiom above lints clean with no suppression.
- A genuinely misordered top-level member record (declared after methods in the type body) still fires — the rule's real target survives.
- Ticketing drops its 24 ORD-01 tokens (reporter will verify on next toolchain refresh).

siy:
Suppression census from the ticketing corpus is in, and it **confirms the mechanism hypothesis down to the call site** (verified against the rule source before posting):

[...]

PR #656, opened 2026-08-26
fix: framework-shape lint FPs — local records exit ORD-01 and SEQ-01, Topic<Self> marks facts, UC-02 learns fact consumers and scheduled hooks

### LINKED ISSUE #647: Slice-framework shape exemptions: VO-01 misses Topic<Self> fact records, UC-02 flags fact consumers and the scheduled dual-entry shape
Companion to #645 — the remaining 9 framework-shape FP tokens from the ticketing suppression census (54 tokens audited, zero hiding real defects; 24 are #645, 20 were repo-side dead weight, 1 unrelated). Both fixes are annotation- or structure-gated, deliberately narrow.

## JBCT-VO-01 — pub-sub fact records (3 tokens)

The rule's own docstring (:31-35) concedes cross-file `@ResourceQualifier` facts "carry no per-file marker… not detectable here". The census found the marker that IS in-file: every fact record declares

```java
public static final Topic<OwnType> TOPIC = Topic.of(...);
```

a `Topic<Self>`-typed static constant. Extend the existing @Slice/@PgSql-style exemption with that structural check — a `Topic` constant parameterized by the declaring type. Specific enough not to catch an ordinary DTO.

## JBCT-UC-02 — two distinct causes (6 tokens)

**(a) Fact consumers (5):** the slice's sole entry method is `execute(FactType event)` — the subscription contract IS the request type, and a synthetic Request wrapper would be pure indirection. Exemption: in `declaresRequestResponse`, an interface whose sole entry method's parameter carries a subscription qualifier (meta-annotated `@ResourceQualifier`) is conforming.

**(b) Dual-entry scheduled slice (1):** `execute(Request)` for the operator route plus a zero-arg `Promise<Unit>` hook required by the Scheduled contract's shape. Splitting the hook into a second @Slice would fragment one use case into two deployables to satisfy the linter. Exemption: a zero-param `Promise<Unit>` method whose annotation is meta-annotated `@ResourceQualifier` does not count toward `entryMethodCount`.

## Acceptance

- The five fact consumers, the scheduled dual-entry slice, and the three fact records lint clean with no suppression.
- A record with a `Topic<OtherType>` constant (not self-parameterized) still fires VO-01; an interface with two genuine business entry methods still fires UC-02 — the rules' real targets survive, pinned by near-miss tests.
- Ticketing's census re-run (attached to #645) shows these 9 tokens gone.

siy:
Fixes #647; advances #645 (the ORD-01 fix and SEQ-01's class-1 — the census re-run and the class-2/3 measurement change keep #645 open). Driven end-to-end by the ticketing suppression census: 54 tokens on a conforming corpus, zero hiding real defects, 33 of them these four rules.

## The four fixes

| Rule | Was | Now |
|---|---|---|
| **ORD-01** | The slice-impl record (declared *inside* its static factory) ranked as a direct type member at rank 0 — lexically doomed behind the rank-3 factory containing it; unsatisfiable in user code, 24 suppressions | A method-local type is `IGNORED` before ranking — via `isLocalDeclaration`, extracted to `FileTypeClassifier` from the primitive `CstValueObjectFactoryRule` already had |
| **SEQ-01** | A local record declaration statement counted its whole body's depth-0 dots as one chain (80 "steps" on a real slice); the blanket suppressions this forced were hiding **genuine** long chains | A statement carrying a local type declaration is never a chain; nested statements were already independently visited. The masked true positives now fire |
| **VO-01** | Pub-sub fact records undetectable ("no per-file marker", per its own docstring) | The marker existed: a `Topic<Self>` static constant. `Topic<OtherType>` still fires |
| **UC-02** | Fact consumers (`execute(FactType)`, the subscription contract IS the request) and the Scheduled contract's zero-param `Promise<Unit>` hook both flagged | Both exempt when the parameter/method carries a user-defined qualifier annotation |

**Marked call:** UC-02's qualifier recognition is an *exclusion* list (`@Override`, `@Nullable`, house annotations, …) because qualifier annotations are user-defined (`@SeatEvents`) and a single-file linter cannot read meta-annotations. Residual failure mode is an FN on a WARNING rule; the ticketing census re-run is the sensor for the list's adequacy.

## The probe as fixture

The reporter's three-class probe is in the suite: the control chain **fires at 7** (deliberately above threshold so an off-by-one can't fake the fix), the record-declaration line is **silent**, and the switch/ternary lines are pinned **at today's wrong counts** with the class-2/3 follow-up named — the split cannot be forgotten, and a fix for it must touch those assertions.

## Audit deliverable (#645's scope-widening)

Every member-scan caller (`directlyEncloses`/`directMethods`/`directNestedTypes`) classified for the method-local hole: ORD-01 was the bug; CAUSE-pack and classifier callers are safe (a local type's members re-scope to it); UC-02's nested-Request check is lenient (FN, acceptable); **two residual findings filed as #655** — ORD-01 also ranks anonymous-class methods (no TYPE_KIND at all — a different guard shape), and a latent `isStepInterface` suppression.

## Evidence

- jbct-lint **756 → 775 green** (+19), implementer run and independent re-run.
- Five mutation checks, each guard removed individually → its tests red (ORD-01 1, SEQ-01 3 incl. the phantom line-8 finding, VO-01 3, UC-02 2), restored.
- RuleFixtures untouched — no positive fixture was accidentally a framework shape.

## Post-merge

Ticketing re-runs its census on the next toolchain install (attached to #645): expected result is 24+20+6+3 = 53 tokens droppable, with the 2 genuine SEQ-01 chains already fixed at their saga seams on their side.


siy:
Pre-merge tightening from field review (ticketing): the fact-consumer exemption now carries a **parameter-type gate** — an annotated parameter whose type is Request/Response-NAMED does not exempt the missing nested pair, so an incidental annotation cannot exempt an ordinary use case (the exemption fails closed). Their four invented qualifier names (@SeatSoldSubscription etc.) were already recognized — the list EXCLUDES known non-qualifiers rather than enumerating qualifiers, so it is not calibrated-to-corpus. Also confirmed: qualifier detection is scoped to the entry method/parameter, not file-wide, so a co-present publisher annotation cannot mask a dual-entry case. 776/776 green.

siy:
Review findings:

The fail-closed parameter-type gate fails open on two constructible spellings, both of which grant the UC-02 exemption on genuinely Request-typed parameters:

- An annotation argument containing `)`, e.g. `@Pattern(regexp = "(ab)+")` — the stripping pattern `[^)]*` stops at the first `)`, which lands inside the argument, and what's left parses as the type `")"`.
- A fully-qualified annotation, e.g. `@com.acme.Traced` — the annotation-matching pattern `@\w+` only consumes `@com`, and what's left parses as the type `.acme.Traced`.

Both are narrow false-negatives on a WARNING-severity rule, not a correctness gate, so non-blocking. Hardening filed as follow-up: #661.

Nit: the appended fail-closed test sits outside the existing `@Nested` groups and pulls in FQN imports rather than following the surrounding test structure. Also, VO-01 misses `Topic<pkg.Self>` when the type argument is FQN-spelled rather than simple-named.

PR #706, opened 2026-08-28
fix: glob-proof FORGE_JVM_OPTS expansion in the six run-forge.sh (#664)

### LINKED ISSUE #664: run-forge.sh: quote FORGE_JVM_OPTS expansion (documented value contains *)
Review findings: `run-forge.sh` expands `${FORGE_JVM_OPTS:-}` unquoted. The documented recommended value for enabling JDWP debug attach is `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005`, which contains a bare `*`.

Under default shell globbing, that `*` happens to pass through unchanged only because it fails to match any filename in the working directory (failed-glob passthrough) — it is not actually protected from expansion. Under `set -f` disabled and a working directory that does contain matching files, or under `shopt -s failglob` / `set -o failglob`, this expands into a filename list or aborts the script outright, silently corrupting the JVM options passed to the process.

**Fix:** quote the expansion, e.g. `"${FORGE_JVM_OPTS:-}"` (or build the options into an array and expand with `"${opts[@]}"` if multiple tokens are needed).

Cross-refs: #653, #608

siy:
Closes #664.

The six example `run-forge.sh` scripts expanded `${FORGE_JVM_OPTS:-}` unquoted, so the documented JDWP value's bare `*` (`address=*:5005`) was only surviving by failed-glob passthrough — a matching file in the working directory, or `failglob`/`nullglob`, corrupts the JVM options silently.

Fix, applied identically in all six: split the value with `read -ra FORGE_JVM_OPTS_ARR <<< "${FORGE_JVM_OPTS:-}"` and expand `"${FORGE_JVM_OPTS_ARR[@]}"`. Word-splitting for multiple flags is preserved, glob expansion is not performed on the tokens, and an empty/unset value expands to zero words (plain quoting would have passed a single empty-string argument to `java`, and would have collapsed multiple flags into one argument).

Verified mechanism empirically (bash, matching file present): `-Da=*` survives literally; empty value yields `java -jar ...` with no extra argv; `read -ra <<< ""` returns 0 so `set -e` is unaffected. `bash -n` clean on all six. The stale "deliberately unquoted" comment above the exec line is rewritten to describe the new mechanism, per the stale-surfaces rule.

### DIFF
--- a/examples/catalog/run-forge.sh
+++ b/examples/catalog/run-forge.sh
@@ -62,5 +62,6 @@ FORGE_ARGS=(
 )
 
 # FORGE_JVM_OPTS: extra JVM flags (e.g. JDWP remote debug -- see docs/slice-developers/forge-guide.md#debugging).
-# Deliberately unquoted: several space-separated flags must word-split.
-exec java ${FORGE_JVM_OPTS:-} -jar "$FORGE_JAR" "${FORGE_ARGS[@]}"
+# Split with read -ra: flags word-split without glob expansion (a literal * in JDWP address= survives, #664).
+read -ra FORGE_JVM_OPTS_ARR <<< "${FORGE_JVM_OPTS:-}"
+exec java "${FORGE_JVM_OPTS_ARR[@]}" -jar "$FORGE_JAR" "${FORGE_ARGS[@]}"
--- a/examples/catalog/run-forge.sh
+++ b/examples/catalog/run-forge.sh
@@ -62,6 +62,7 @@ FORGE_ARGS=(
 )
 
 # FORGE_JVM_OPTS: extra JVM flags (e.g. JDWP remote debug -- see docs/slice-developers/forge-guide.md#debugging).
-# Split with read -ra: flags word-split without glob expansion (a literal * in JDWP address= survives, #664).
-read -ra FORGE_JVM_OPTS_ARR <<< "${FORGE_JVM_OPTS:-}"
+# Split with read -rd into an array: flags word-split on whitespace INCLUDING newlines, no glob expansion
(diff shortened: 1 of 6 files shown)


siy:
**CTO review round — verdict: MERGE-WITH-NITS.** Empirically verified on bash 3.2: pre-fix glob hazard reproduced, post-fix literal pass; empty/unset → zero argv words under set -e; all six scripts covered, no seventh consumer. NITS: (1) newline-embedded values now truncate after line 1 (`read` consumes one line) — genuine but rare regression; optional fix `IFS=$' \t\n' read -rd '' -a ARR <<< "${FORGE_JVM_OPTS:-}" || true`; (2) quoted-substring naive split — pre-existing, no claim made, fine; (3) a stray ide-plugins-plan.md commit rides this branch — rebase on current release HEAD before merge.

siy:
Review nit (1) applied: the split is now `IFS=$' \t\n' read -rd '' -a FORGE_JVM_OPTS_ARR <<< "${FORGE_JVM_OPTS:-}" || true` in all six scripts — newline-embedded values no longer truncate after the first line. Empirically verified before committing: a newline-separated two-flag value splits into exactly two argv words with the glob literal preserved; empty/unset still yields zero argv words under `set -e`; `bash -n` clean on all six. Nit (2) (quoted-substring splitting) left as-is per the verdict — pre-existing, no claim made. Nit (3): branch rebased onto current release HEAD, stray docs commit gone (`origin/release-1.0.0-rc3..HEAD` = this PR's two commits). Ready for re-review — not self-merging.

PR #707, opened 2026-08-28
fix: Jackson Option/Result element contextualization resolves the generic binding (#696)

### LINKED ISSUE #696: Jackson OptionDeserializer skips contextualization for Option<T>, so values arrive as raw Object — RefundRequest.partialAmount() throws ClassCastException at first use
## What

`OptionDeserializer.createContextual` filters on `prop.getType().hasContentType()`, which is **false for `Option<T>`**. Contextualization is therefore skipped and the element type is never applied, so values deserialize as raw `Object`:

- `Option<Long>` yields an `Integer`
- `Option<Money>` yields a `LinkedHashMap`

The field's declared type is a lie at runtime, and nothing fails until something reads it.

## Live impact, not theoretical

`examples/ecommerce/payment` — `RefundRequest.partialAmount()` throws `ClassCastException` **at first use**. Any endpoint whose request record carries an `Option<T>` of a non-String type has the same defect latent in it.

## How it surfaced

Found while building the I4 durable-timer Stage C gates (#351). The natural seam for a caller-controlled timer delay was `Option<Long> delayMillis` on the request record; it failed in a live forge cluster. The working seam is a **boxed `Long`** plus `Option.option(...)` at the use site — note that a primitive `long` does NOT work either, because Jackson 3 rejects an *absent* primitive creator property (`Type mismatch: expected long, got unknown`, HTTP 500), which would have made the field mandatory-in-fact rather than optional.

That combination is worth stating plainly for anyone hitting this: **on a request record, an optional non-String field must be a boxed type today.** `Option<T>` looks like the idiomatic choice and is the one that breaks.

## Why it stays invisible

The failure is at the READ, arbitrarily far from the deserialization that caused it, and it presents as a `ClassCastException` in application code rather than as a deserialization error. Nothing in the request-handling path reports that a declared `Option<Long>` produced an `Integer`.

## Fix direction

`createContextual` needs to resolve `Option`'s element type through the generic binding rather than through `hasContentType()`. [design intent — unverified] — I did not attempt the fix; the diagnosis is from reading the deserializer against the observed runtime types.

## Acceptance

- `Option<Long>`, `Option<Money>` and `Option<SomeRecord>` on a request record each deserialize to the declared element type.
- A regression test covering a non-String `Option` element, since `Option<String>` happens to work by accident and would pass a naive test.
- `RefundRequest.partialAmount()` exercised end-to-end.

siy:
Closes #696.

**Mechanism.** `OptionDeserializer.createContextual` filtered on `prop.getType().hasContentType()`, which is false for `Option<T>`: Option is registered on the plain class, so a declared `Option<T>` arrives as a `SimpleType` whose element type lives in the generic binding, not in a Jackson content type. Contextualization was silently skipped and elements deserialized as raw `Object` — `Integer` for `Option<Long>`, `LinkedHashMap` for record elements — blowing up only at first use. `ResultDeserializer` carried the identical defect (`.filter(JavaType::hasContentType)`), so this PR fixes both.

**Fix.** Element type resolution now goes `hasContentType() ? getContentType() : containedTypeCount() == 1 ? containedType(0) : none` — the content-type path is kept first (no behavior change anywhere it worked), the generic binding covers the SimpleType reality, and a raw `Option`/`Result` (no binding) keeps today's fallback. Root-level reads (no `BeanProperty`) are untouched.

**Proof (red/green).** With the new regression tests and the fix stashed, exactly the three typed-element tests fail with the ticket's own signatures: `ClassCastException: Integer cannot be cast to Long` (`Option<Long>`), `LinkedHashMap cannot be cast to Amount` (`Option<record>` and `Result<record>`); nothing else in the module moved. With the fix: `Tests run: 28, Failures: 0, Errors: 0`, BUILD SUCCESS (`mvn -pl integrations/json/jackson -am test`).

Tests deliberately pin non-String elements only — `Option<String>` passes by accident pre-fix and would have made a naive regression test worthless (per the ticket's acceptance note). The ticket's third acceptance item (`RefundRequest.partialAmount()` end-to-end in `examples/ecommerce/payment`) needs a forge cluster run and is left for the merge-round example verification; the unit layer now pins the exact defect that broke it.

### DIFF
--- a/integrations/json/jackson/src/main/java/org/pragmatica/json/OptionDeserializer.java
+++ b/integrations/json/jackson/src/main/java/org/pragmatica/json/OptionDeserializer.java
@@ -58,17 +58,30 @@ public class OptionDeserializer extends ValueDeserializer<Option<?>> {
 
     @Override
     public ValueDeserializer<?> createContextual(DeserializationContext ctxt, BeanProperty property) {
-        return option(property).filter(prop -> prop.getType()
-                                                   .hasContentType())
-                     .<ValueDeserializer<?>> map(prop -> createContextualDeserializer(ctxt, prop))
+        return option(property).map(BeanProperty::getType)
+                     .flatMap(OptionDeserializer::elementType)
+                     .<ValueDeserializer<?>> map(type -> createContextualDeserializer(ctxt, property, type))
                      .or(this);
     }
 
-    private OptionDeserializer createContextualDeserializer(DeserializationContext ctxt, BeanProperty prop) {
-        var contentType = prop.getType().getContentType();
-        var deser = ctxt.findContextualValueDeserializer(contentType, prop);
+    /// Option is registered on the plain class, so a declared `Option<T>` arrives as a SimpleType:
+    /// `hasContentType()` is false and `T` lives in the generic binding instead (#696). Without this
+    /// resolution the element deserializes as raw Object (Integer for Option<Long>, LinkedHashMap
+    /// for record elements) and the declared type is violated at first use.
+    private static Option<JavaType> elementType(JavaType type) {
+        return type.hasContentType()
+               ? option(type.getContentType())
+               : type.containedTypeCount() == 1
+                 ? option(type.containedType(0))
+                 : none();
+    }
 
-        return new OptionDeserializer(contentType, deser);
+    private OptionDeserializer createContextualDeserializer(DeserializationContext ctxt,
+                                                            BeanProperty property,
+                                                            JavaType elementType) {
+        var deser = ctxt.findContextualValueDeserializer(elementType, property);
+
+        return new OptionDeserializer(elementType, deser);
     }
 
     @Override
--- a/integrations/json/jackson/src/main/java/org/pragmatica/json/OptionDeserializer.java
+++ b/integrations/json/jackson/src/main/java/org/pragmatica/json/OptionDeserializer.java
@@ -84,11 +84,14 @@ public class OptionDeserializer extends ValueDeserializer<Option<?>> {
     private OptionDeserializer createContextualDeserializer(DeserializationContext ctxt,
                                                             BeanProperty property,
                                                             JavaType elementType) {
-        // A same-wrapper element (Option<Option<X>> — forbidden in JBCT, but a crash is never the
-        // contract) passes the raw-class guard at every level and would recurse through
-        // findContextualValueDeserializer with the same property: read it through the type-directed
-        // path instead of a contextualized delegate.
-        var deser = elementType.getRawClass() == Option.class
+        // Two element shapes must NOT get a property-contextualized delegate, or contextualization
+        // re-enters with the same property and recurses: a same-wrapper element (Option<Option<X>> —
+        // forbidden in JBCT, but a crash is never the contract), and a container element
+        // (Option<List<Option<X>>> — Collection/Map deserializers propagate the property into
+        // content contextualization, so a same-wrapper element beyond the container re-derives it).
+        // The type-directed path (valueType only) stays fully typed for containers: readValue
+        // contextualizes against the element's own JavaType with no property.
+        var deser = elementType.getRawClass() == Option.class || elementType.isContainerType()
                     ? null
                     : ctxt.findContextualValueDeserializer(elementType, property);
 
--- a/integrations/json/jackson/src/main/java/org/pragmatica/json/OptionDeserializer.java
+++ b/integrations/json/jackson/src/main/java/org/pragmatica/json/OptionDeserializer.java
@@ -59,6 +59,7 @@ public class OptionDeserializer extends ValueDeserializer<Option<?>> {
     @Override
     public ValueDeserializer<?> createContextual(DeserializationContext ctxt, BeanProperty property) {
         return option(property).map(BeanProperty::getType)
+                     .filter(type -> type.getRawClass() == Option.class)
                      .flatMap(OptionDeserializer::elementType)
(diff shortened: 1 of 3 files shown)


siy:
**CTO review round — verdict: REQUEST-CHANGES.** Core fix and tests sound; one new crash mode needs a guard before merge. **MAJOR:** the new `elementType` derivation + `findContextualValueDeserializer(elementType, property)` re-invokes `createContextual` with the SAME outer property, so nested wrapper fields — `Result<Option<X>>`, `Option<Result<X>>`, `Option<Option<X>>` — re-derive the wrapper forever: unbounded recursion → StackOverflowError (`Result.lift` masks it as failure; raw-mapper users crash). `Result<Option<T>>` appears 20+ times as a signature; DTO use is a matter of time. FIX: guard with `type.getRawClass() == Option.class` (resp. `Result.class`) before `containedType(0)`, plus a pinning test. Also: add an `Option<List<X>>` test (content-type branch — the same guard fixes its pre-existing recursion too). Verified: the three CCE pins are genuinely red pre-fix; exactly two deserializers in the module, mechanisms symmetric. Stray ide-plugins-plan.md commit rides the branch — rebase before merge.

siy:
Review fixes applied and proven.

**The recursion guard** (both deserializers): `createContextual` now filters on `property.getType().getRawClass() == Option.class` (resp. `Result.class`) before deriving the element — since `createContextual` sees only the property's type, an inner wrapper resolved for another wrapper's element would otherwise re-derive the outer forever. One extension beyond the prescribed guard, found by the pinning tests: **same-wrapper nesting (`Option<Option<X>>`) passes the raw-class check at every level and still recursed** — the delegate for a same-wrapper element is therefore built on the type-directed read path (`valueType` set, no contextualized delegate) instead of `findContextualValueDeserializer` with the same property.

**Red (pre-guard, verbatim):** `Tests run: 8, Failures: 3` in `OptionElementContextualization` — `readString_doesNotRecurse_forResultOfOptionField`, `forOptionOfResultField`, `forOptionOfOptionField` all failed with `java.lang.StackOverflowError` (6 occurrences in the log). The nested pins deliberately assert no-crash + outer wrapper structure only — inner elements remain untyped through nested wrappers, same as before this PR.

**Also added per the verdict:** `readString_deserializesTypedListElements_forOptionListField` (`Option<List<Amount>>` → typed `Amount` elements). It passes even pre-guard — collections contextualize against their own type — and stays as a green pin.

**Green:** JsonMapperTest 32/32; full module suite `Tests run: 32, Failures: 0, Errors: 0`, BUILD SUCCESS, zero StackOverflow occurrences. Branch rebased onto current release HEAD — the stray ide-plugins-plan.md commit is gone (`git log origin/release-1.0.0-rc3..HEAD` shows only this PR's two commits).

siy:
**Re-review: one recursion family SURVIVES — REQUEST-CHANGES continues.** Confirmed fixed: all direct nestings terminate (Result<Option<X>>, Option<Result<X>>, Option<Option<X>>, Result<Result<X>>; cross-wrapper via the raw-class filter, same-wrapper via the skipped delegate), the guard also fixes the pre-existing List<Option<X>> container recursion, no regressions, pins genuinely red-provable. **Remaining shape (derived from Jackson's contract): same wrapper separated by a container** — `Option<List<Option<X>>>`, `Result<List<Result<X>>>` (a plausible bulk-result shape), `Option<Map<K,Option<V>>>`. Mechanism: createContextualDeserializer resolves the container elementType via findContextualValueDeserializer(elementType, property), and Jackson's Collection/Map deserializers propagate that SAME property into content contextualization — the inner same-wrapper filter passes (property raw == wrapper) and re-derives the container: cycle. **Fix: skip the delegate when `elementType.isContainerType()`** — valueType-only readValue contextualizes with null property, stays typed, breaks the chain. Add one pin for the container-separated shape.

siy:
Container-separated recursion fixed per the re-review mechanism.

**Fix:** the delegate skip now covers BOTH shapes that re-enter contextualization with the same property — same-wrapper elements (previous commit) and **container elements** (`elementType.isContainerType()`): Collection/Map deserializers propagate the property into content contextualization, so a same-wrapper element beyond the container re-derived it forever. Container elements now go through the type-directed path (`valueType` only) — `readValue` contextualizes against the element's own `JavaType` with no property, which both breaks the chain and stays fully typed.

**Red (verbatim):** new pin `readString_doesNotRecurse_forContainerSeparatedSameWrapper` (`Option<List<Option<Long>>>`) against the previous commit: `Tests run: 9, Failures: 1` — `AssertionFailedError: Should not fail: Cause: java.lang.StackOverflowError` (at `DeserializationConfig.hasAbstractTypeResolvers`); every other test unchanged.

**Green:** full module `Tests run: 33, Failures: 0, Errors: 0` (root aggregate 856/0/0), BUILD SUCCESS, zero StackOverflow occurrences; critically, `readString_deserializesTypedListElements_forOptionListField` (`Option<List<Amount>>` → typed `Amount` elements) still passes THROUGH THE NEW PATH — the container skip preserves element typing, per the reviewer's prediction.

siy:
**Final re-review: CONFIRMED-FIXED — verdict upgraded to MERGE.** isContainerType() delegate-skip present in both deserializers, correctly OR-ed with the same-wrapper check; all container shapes terminate (Collection/Map/array/Collection-like); no untyping — containers under wrappers were never typed pre-PR, and the type-directed readValue path resolves content from the JavaType itself (Option<List<Amount>> pin green through the new path). One theoretical, non-blocking nit for the follow-up batch: ReferenceType elements (e.g. Option<AtomicReference<Option<X>>>) would still propagate the property — no such wire shape exists in-repo; add `|| elementType.isReferenceType()` when convenient.

Issue #386, opened 2026-06-30
Durable pub-sub: at-least-once + DLQ + idempotency + projection (at-most-once today)


siy:
## Summary
Pub-sub delivery is **at-most-once** today — no redelivery, replay, DLQ, idempotency, or projection support. Make the delivery contract honest and durable, since the framework leans on projections/convergence as a first-class pattern that assumes durability it currently can't deliver.

## Current behaviour (HEAD)
`TopicPublisher.publish` collects subscriber promises via `Promise.allOf` and propagates failure (`aether/aether-invoke/.../TopicPublisher.java:35-45`) — so the publisher *sees* a failure, but there is **no retry / redelivery / replay / DLQ**. A subscriber that is down at publish time, or a transient store error during convergence, **permanently desyncs the projection**. Recorded in `aether/docs/reference/guarantees.md` §5 (pub-sub = at-most-once).

The subscriber signature `Promise<Unit> execute(Fact)` invites developers to write `.recover(...)` and reason about failure, but with no redelivery the recover-vs-not distinction is moot — the API shape implies a guarantee the runtime doesn't keep.

## Proposal
Pick one and make the contract honest:
- **(a)** honor at-least-once — retry/backoff + DLQ + an idempotency-key resource; or
- **(b)** change the subscriber signature so it no longer implies a delivery guarantee.

Then ship a **durable `PubSub` impl**, and ideally a built-in **projection** abstraction (checkpoint / replay / idempotent upsert), since that's the encouraged style. Publish a doc table stating the delivery guarantee per impl.

## Acceptance
Either a subscriber failure visibly triggers redelivery/DLQ, or the signature no longer implies one; a doc table states the delivery guarantee per `PubSub` impl.

## Related / notes
- `guarantees.md` §5 — pub-sub at-most-once (authoritative).
- #274 (CLOSED) was *adjacent* — it fixed `TopicSubscriptionKey` missing nodeId + cosmetic namespaces; durable delivery as a feature was never tracked.
- ⚠ The wishlist's specific "`InMemoryPubSub.deliverToSubscribers` discards the returned Promise / always returns success" is the **ticketing posterchild's own** in-memory impl, **not** pragmatica — that discard should be fixed in the ticketing repo, separately from this feature.

---
Source: ticketing posterchild `AETHER-WISHLIST.md` item #2 (P0).

siy:
[...]

PR #708, opened 2026-08-28
fix: slice-processor dependency-method scan walks super-interfaces (#663)

### LINKED ISSUE #663: slice-processor: dependency methods inherited from super-interfaces are silently invisible
Review findings: dependency methods declared on a super-interface of a slice's dependency interface are not visible to the slice-processor's dependency-method scan. They are silently dropped — no generated proxy method, no diagnostic — rather than either being supported or rejected with an error.

This is a sibling of the silence family tracked in #612 (non-Promise dependency methods silently dropped instead of erroring): both are cases where the processor's dependency-collection pass fails to surface a method it should either handle or reject, and the developer gets no signal that anything was skipped.

**Recommendation:** either walk super-interfaces when collecting dependency methods (matching normal Java interface-inheritance semantics), or, if that's out of scope, emit an explicit diagnostic when a dependency interface extends another interface, so the gap is visible rather than silent.

Cross-refs: #641, #612

siy:
Fixes #663 (does not auto-close on a release branch — close manually on merge).

## Mechanism (what the code actually did — it differs from the issue's "silently dropped")

`FactoryClassGenerator.collectProxyMethods` walked only `getEnclosedElements()` of the dependency interface. A method declared on a super-interface never reached the dependency-method scan, but the outcome was not a silently *succeeding* build: the truncated proxy record still `implements`-ed the interface, so javac failed on the **generated** factory. The processor itself emitted no diagnostic — the "silence" is that nothing pointed at the actual gap, and the error was anchored in code the user never wrote:

```
/SOURCE_OUTPUT/test/OrderServiceFactory.java:25: error: inventoryService is not abstract and does not override abstract method stockOf(java.lang.String) in external.StockQuery
        record inventoryService(MethodHandle<Integer, String> checkStockHandle) implements InventoryService {
```

An inherited **non-Promise** method bypassed the #612 gate the same way and surfaced as the same missing-override noise instead of the #612 error. (An inherited `default` method would compile silently and run its default body locally — same intentional exemption as declared defaults.)

## Chosen resolution: (a) walk super-interfaces

No blocker found in the proxy machinery: every consumer (proxy record, method-handle wiring, request records, codec entries) reads from the one `proxyMethodsCache` that `collectProxyMethods` fills, so extending the scan extends all of them coherently. Details:

- **BFS over the super-interface closure** — declared methods first, then supers level by level in declaration order; generation stays deterministic.
- **Override de-duplication** — one proxy entry per signature; the nearest declaration wins. Signatures use erased parameter types **after** `Types.asMemberOf` substitution, so a generic super's `lookup(K)` seen through `extends KeyedQuery<String>` is proxied as `lookup(String)` and its override collapses onto it instead of duplicating.
- **#612 gate intact for inherited methods** — an inherited non-Promise method hits the same compile error; the message names the declaring interface plus an `(inherited by dependency ...)` clause. Declared-method message text is unchanged.
- **`java.lang.Object`-level members excluded** — explicit re-declarations of `toString()`/`hashCode()`/`equals(Object)` (JLS 9.2) are never proxy candidates, keyed on type identity via `Types.isSameType` over the parameter's erasure rather than string spelling (review fix), so an annotated `equals(@Nullable Object)` re-declaration is excluded too; the generated record satisfies them itself. Applied uniformly (declared and inherited) — judgment call: before this PR a *declared* `String toString();` drew a #612 error on a legal, always-satisfied member.
- Only `abstract` methods are proxy candidates (review fix) — javax.lang.model reports implicit modifiers, so this one check exempts `static`, `default`, AND Java 9+ `private` interface methods at every level. The earlier modifier-skip list (static/default only) let a private super-interface instance method into the scan, changing generated output for a previously-compiling slice: a private non-Promise helper drew a spurious #612 error, and a private Promise-returning method wrongly gained a record component + wiring + codec entry.

## Red/green evidence

New tests (in `jbct/slice-processor/src/test/java/org/pragmatica/jbct/slice/SliceProcessorTest.java`), run first against the **unfixed** generator — `Tests run: 5, Failures: 3`:

- `should_generate_proxy_for_dependency_method_inherited_from_super_interface` (two-level chain, pins transitivity) — RED: `error: inventoryService is not abstract and does not override abstract method stockOf(java.lang.String) in external.StockQuery`
- `should_substitute_type_arguments_for_generic_super_interface_dependency_method` — RED: `... does not override abstract method lookup(java.lang.String) in external.KeyedQuery`
- `should_fail_on_non_promise_dependency_method_inherited_from_super_interface` — RED: `Expected an error containing "external.LegacyQuery.syncLookup returns java.lang.String", but only found: inventoryService is not abstract and does not override abstract method syncLookup(java.lang.String) in external.LegacyQuery`
- `should_count_overridden_dependency_method_once` and `should_keep_flat_dependency_interface_proxy_unchanged` (control: exact proxy-record header pinned) — green before and after, as expected: they see only declared methods today.

With the fix — `Tests run: 5, Failures: 0`, and full suites:

```
Tests run: 313, Failures: 0, Errors: 0, Skipped: 0   (slice-processor, incl. 5 review-fix tests/pins)
Tests run: 57,  Failures: 0, Errors: 0, Skipped: 0   (slice-processor-tests)
JBCT Slice Processor ............................... SUCCESS
Slice Processor Integration Tests .................. SUCCESS
```

`mvn -f jbct/pom.xml test -pl slice-processor,slice-processor-tests -am` — test-only, nothing installed.

## Envelope invariant

This IS a generated-output structural change for the affected inputs (the proxy record gains `MethodHandle` components, proxy methods, wiring and codec entries for inherited methods), so per `aether/docs/contributors/envelope-versioning.md` it is recorded in the version-history table — but **no `ENVELOPE_FORMAT_VERSION` bump**: the doc's GA freeze (owner ruling 2026-07-18, #386) directs structural changes to ride without bumps, marked "(no bump — GA freeze)". Additionally, no previously-loadable artifact changes shape: dependency interfaces without super-interfaces are byte-for-byte identical (pinned by the control test), and inputs with inherited methods previously produced non-compiling generated code, so no jar built from them exists. The manifest itself lists dependency interfaces, not their methods — manifest structure is unchanged.


siy:
**CTO review round — verdict: REQUEST-CHANGES.** One real compat hole; the probed areas otherwise hold. **BLOCKER:** the scan filter skips only STATIC/DEFAULT — a Java 9+ *private* super-interface instance method passes it, `asMemberOf` resolves it with no visibility check, and it gains a record component + wiring + codec entry. Such inputs COMPILED before (old scan never saw supers), so generated output changes for a previously-compiling slice — undermining this PR's own support for the no-envelope-bump claim — and a private non-Promise helper now draws a spurious #612 error. FIX: require `Modifier.ABSTRACT` (implicit modifiers are reported; covers static/default/private at once). **ALSO FIX:** `isObjectLevelMethod` string-compares "java.lang.Object" — a super re-declaring `equals(@Nullable Object)` escapes exclusion → spurious #612; compare erasure via `types.isSameType`. **ADD PINS:** diamond tie-break determinism (extends-clause order via `getInterfaces()`), and multi-level generic indirection (`A<T> extends B<List<T>>` — asMemberOf composes correctly but only single-level is pinned). Verified: #612 gate parity for inherited vs declared is identical.

siy:
## CTO review fixes applied (commit 7411f4cf0)

Rebased onto `origin/release-1.0.0-rc3` first — the stray docs commit no longer rides the branch; `git log origin/release-1.0.0-rc3..HEAD --oneline` shows only this PR's two commits (`594164d4b` original fix, `7411f4cf0` review fixes).

### 1. BLOCKER — private super-interface instance methods entered the scan
The filter skipped only `STATIC`/`DEFAULT`; a Java 9+ *private* super-interface instance method passed it, `asMemberOf` resolved it, and generated output changed for a slice that compiled before this PR.

**Fix** (`FactoryClassGenerator.collectProxyMethods`): require `Modifier.ABSTRACT` on collected methods — javax.lang.model reports implicit modifiers, so this single check exempts static, default, AND private at once.

**Red (pre-fix branch state, tests written first):**
- `should_ignore_private_super_interface_helper_method` — spurious #612 on the private non-Promise helper:
  ```
  external/CachedQuery.java:8: error: Slice dependency method external.CachedQuery.cacheKey returns java.lang.String;
  a slice dependency method must return Promise<T> — ... (inherited by dependency external.InventoryService)
  ```
- `should_not_proxy_private_promise_returning_super_interface_method` — the private `prefetch` wrongly gained a proxy entry (record component + wiring), and the generated factory then failed javac:
  ```
  /SOURCE_OUTPUT/test/OrderServiceFactory.java:37: error: method does not override or implement a method from a supertype
  ```

### 2. `equals(@Nullable Object)` escaped the Object-method exclusion
`isObjectLevelMethod` string-compared `"java.lang.Object"`; a TYPE_USE annotation rides along in `TypeMirror#toString`, so the annotated re-declaration escaped and drew a spurious #612.

**Fix**: compare the parameter's **erasure** via `types.isSameType` against `java.lang.Object`'s type (new `isJavaLangObject` helper) — type identity, not string spelling.

**Red (pre-fix):** `should_exclude_annotated_object_equals_redeclaration_from_dependency_scan`:
```
external/AuditedQuery.java:8: error: Slice dependency method external.AuditedQuery.equals returns boolean; ...
(inherited by dependency external.InventoryService)
```

### 3. Pins added (both run against the pre-fix state: green-on-arrival, as suspected)
- `should_pick_first_extends_clause_declaration_for_diamond_dependency_method` — two supers declare the same signature; the winner follows extends-clause order (`getInterfaces()`): first-listed super's declaration supplies the proxy method (pinned via its parameter name `alphaKey`, `betaKey` absent), exactly one `lookupHandle` component.
- `should_compose_substitution_across_generic_super_interface_levels` — `KeyedQuery<K> extends BatchQuery<List<K>>` bound at `K=String`: `asMemberOf` composes `U -> List<K> -> List<String>`; generated `putAll(List<java.lang.String> items)` pinned.

### Green
Pre-fix: `Tests run: 5, Failures: 3` (the three regressions above red, both pins green). Post-fix: `Tests run: 5, Failures: 0`. Full suites (`mvn -f jbct/pom.xml test -pl slice-processor,slice-processor-tests -am`, test-only, nothing installed):
```
Tests run: 313, Failures: 0, Errors: 0, Skipped: 0   (slice-processor)
Tests run: 57,  Failures: 0, Errors: 0, Skipped: 0   (slice-processor-tests)
JBCT Slice Processor ............................... SUCCESS
Slice Processor Integration Tests .................. SUCCESS
```

CHANGELOG entry and PR body updated where they claimed the old exemption set / suite count.

siy:
**Re-review: CONFIRMED-FIXED — verdict upgraded to MERGE.** The ABSTRACT filter closes the private-method hole with no leak (implicit modifiers are reported, so bodyless interface methods carry ABSTRACT at any nesting; static/default/private all lack it). Over-exclusion is impossible by JLS 9.2 (an interface cannot declare a non-boolean equals(Object) or Object-erasure equivalent, so no legitimate Promise-returning dependency method can collide with the Object exclusion — documented at the exclusion site; equals(MyType) still proxies). Both demanded pins verified real and mutation-sensitive (diamond tie-break: alphaKey wins, betaKey absent, handle count 2; multi-level substitution: putAll(List<String>)). Queued for the serial merge round.

PR #709, opened 2026-08-28
fix: #661 UC-02 exemption type gate reads the parameter TYPE node — regex stripping failed open on paren annotation args and FQN annotations

### LINKED ISSUE #661: UC-02 exemption type gate: annotation stripping fails open on nested-paren annotation arguments and FQN annotations
Review findings: the fail-closed parameter-type gate introduced for UC-02 exemptions fails open on two constructible spellings, both of which grant the exemption on genuinely Request-typed parameters.

**Repro 1 — annotation argument containing `)`:**

```java
@Pattern(regexp = "(ab)+")
CreateOrderRequest body
```

The stripping regex (`[^)]*`) stops at the first `)`, which lands inside the annotation argument rather than at the end of the annotation. What remains after stripping parses as the bare string `")"` instead of `CreateOrderRequest`, so the gate does not recognize the parameter as Request-typed and the exemption is granted incorrectly.

**Repro 2 — fully-qualified annotation:**

```java
@com.acme.Traced CreateOrderRequest body
```

The annotation-matching regex (`@\w+`) only consumes `@com`, leaving `.acme.Traced CreateOrderRequest body` behind. The remaining text parses as type `.acme.Traced`, again masking the true parameter type and granting the exemption.

Both are narrow false-negatives on a WARNING-severity rule (not a correctness gate), but they defeat the fail-closed intent the gate was built for.

**Recommendation:** stop doing text surgery on the parameter source. Either strip annotations with balanced-paren tracking, or — preferably — read the parameter's TYPE node directly from the parsed AST/parameter model instead of reconstructing it from source text after regex stripping.

Cross-refs: #656, #647

siy:
## Mechanism (confirmed — matches #661's diagnosis exactly)

`CstUseCaseStructureRule.hasQualifiedFactParameter` gated the UC-02 fact-consumer exemption on `parameterTypeName(param)`: the raw parameter source text with annotations regex-stripped (`@\w+(\([^)]*\))?\s*`), first whitespace-delimited token taken as the type. Both #661 spellings defeat it, granting the exemption on genuinely Request-typed parameters:

1. **Annotation argument containing `)`** — `@Pattern(regexp = "(ab)+") CreateOrderRequest body`: `[^)]*` stops at the `)` inside `"(ab)"`, the regex consumes `@Pattern(regexp = "(ab)`, and the residue `+") CreateOrderRequest body` parses as type `+")` (the issue says `)`; same failure, off by the exact residue). Not Request-suffixed → exemption wrongly granted.
2. **Fully-qualified annotation** — `@com.acme.Traced CreateOrderRequest body`: `@\w+` consumes only `@com`, the residue `.acme.Traced CreateOrderRequest body` parses as type `.acme.Traced`. Not Request-suffixed → exemption wrongly granted.

## Fix

Per the issue's recommendation, no text surgery: the type is read from the parameter's `Type` CST node (`PlainParam <- Annotation* Modifier* Type Identifier`), which the parser has already separated from the annotations.

- New `CstNodes.typeSimpleName(Cursor type)` (jbct-parser): the identifier of the type's last name segment, read at **token** level — `RefType <- AnnotatedTypeName ('.' AnnotatedTypeName)*`, identifier = last non-trivia token of the final segment before its `TypeArgs` subtree. Handles simple names, FQN types, type annotations, generics, single-token primitive leaves; returns `none` when no name segment exists.
- The rule's gate (`carriesFactType`) grants the exemption only when the type name is **positively** read and non-`Request`/`Response`; an unreadable type now DENIES the exemption instead of granting it — fail-closed in both directions.
- CST shapes were verified empirically with a throwaway probe (parameter subtree dumps for both spellings plus FQN types, generics, primitives, arrays) before writing the extraction.

## Red (new tests against the unfixed rule)

```
[ERROR]   CstUseCaseStructureRuleTest$Violations.flags_request_typed_parameter_when_the_annotation_argument_contains_a_paren:117 expected: <true> but was: <false>
[ERROR]   CstUseCaseStructureRuleTest$Violations.flags_request_typed_parameter_when_the_qualifier_annotation_is_fully_qualified:129 expected: <true> but was: <false>
[ERROR] Tests run: 18, Failures: 2, Errors: 0, Skipped: 0
```

Both red tests pin the specific `no nested Request/Response` diagnostic, so they cannot pass through an unrelated UC-02 finding.

## Green (with the fix)

```
Tests run: 18, Failures: 0, Errors: 0, Skipped: 0   (CstUseCaseStructureRuleTest)
```

Controls: `clean_on_fact_consumer_with_a_qualified_parameter` (pre-existing), plus new `clean_on_fact_consumer_with_paren_annotation_argument_and_fully_qualified_fact_type` and `clean_on_fact_consumer_with_a_generic_fact_type` — legitimately exempt parameters keep their exemption under the same annotation spellings.

Full in-reactor suite (`mvn -f jbct/pom.xml test -pl jbct-lint -am`): jbct-core 32, jbct-parser 135, jbct-lint 783 — all green, zero regressions.

Closes #661 (manual close on merge — release-branch workflow).


siy:
**CTO review round — verdict: MERGE-WITH-NITS.** Fail-closed verified by grammar analysis against java25.peg + CstNodes: every shape (simple, FQN, generics, type-use/segment annotations, ref-arrays with sibling DIMS, varargs via LAST_PARAM, C-style dims) resolves through kind-preserving nodes, and every absence path yields or(false) — no permissive fallthrough found. The exemption flip (`@Q byte[]` exempt→flagged) is bounded, documented, and lands at WARNING severity. Both red pins verified genuinely red pre-fix. NITS (non-blocking): add green pins for varargs, ref-array, and FQN-qualifier fact consumers — correct today by grammar analysis, unprotected against a future typeSimpleName refactor.

siy:
Review nits applied: the three green pins are in — varargs, reference-array, and FQN-qualifier fact consumers (`CstUseCaseStructureRuleTest$CleanCases`), each verified passing via surefire XML (no failure/error children) and framed in-code as refactor protection for `typeSimpleName`. Full jbct-lint module: 786/786, BUILD SUCCESS. Branch rebased onto current release HEAD; stray docs commit gone. Ready for re-review — not self-merging.

PR #710, opened 2026-08-28
fix: route factory near-miss warns instead of silent validation skip; match via Types.isSameType (#662)

### LINKED ISSUE #662: Route request factory detection: silent fallback to canonical constructor when factory signature text-mismatches
Review findings: factory detection for generated route request construction compares parameter and component types by comparing `TypeMirror.toString()` textually. On any mismatch — including a near-miss caused only by cosmetic differences in how a type is spelled, such as an annotated type or a generic record component — the detector silently falls back to the canonical constructor.

That fallback is exactly the failure mode this PR set out to fix: routes silently constructing request records through the canonical constructor instead of the validating factory, so validation is silently skipped. A factory that exists but doesn't structurally match what the detector expects reproduces the original bug with no diagnostic pointing at why.

**Recommendation:** when a candidate factory method is found (by name/arity convention) but its parameter types fail the textual match against the record's components, emit a "factory found but unmatched" WARNING at the call site instead of silently falling back. This turns a silent validation gap back into a visible, actionable signal.

Minor, non-blocking: the "first declared wins" resolution when multiple candidate factories exist relies on `Element.getEnclosedElements()` returning elements in declaration order, which the `javax.lang.model` API does not guarantee. Worth revisiting alongside the fix above.

Cross-refs: #643, #605

siy:
Fixes #662.

## Confirmed mechanism

The code matches the issue's description. `RouteSourceGenerator.declaredFactory` compared candidate factory parameter types against record component types by `TypeMirror.toString()` textual equality (`parameterTypeNames(method).equals(componentTypes)`), and `returnsResultOf` compared the carried type argument textually as well. Any mismatch — including a purely cosmetic one such as a type-use annotation on a factory parameter — returned `Option.none()`, and every construction site (`constructAndDelegate`, `bodyHandlerExpr`) silently fell back to the canonical constructor / raw deserialized value: routes silently skipping the record's validation, the exact failure mode #643 (#605) was built to fix, with no diagnostic.

## Fix

1. **"Factory found but unmatched" WARNING** — a *factory-shaped* method (static, component-count arity, returning `Result` of this record in some spelling) that fails the component-type match now draws a WARNING anchored at that method's declaration, naming the expected component types and the found parameter/return types, and stating that routes fall back to the canonical constructor with validation skipped. The fallback behavior itself is unchanged. Helpers with a different arity or carrying a different type are not shaped and stay silent — no new noise on ordinary records.
2. **Semantic match via `Types.isSameType`** — the processing environment is available at the construction site (`SliceProcessor.init` already holds `processingEnv.getTypeUtils()`), so `Types` is now passed into `RouteSourceGenerator` and the per-position comparison is `isSameType` on the mirrors, then `isSameType` after boxing when exactly one side is primitive, with the old textual comparison retained as a strict-superset compatibility fallback. A type-use-annotated factory parameter is the same type and now matches (validation runs) instead of silently disabling the factory. Full erasure is deliberately **not** applied: it would equate `List<String>` with `List<Integer>` and emit a factory call that does not compile.

**Untouched, per the issue:** "first declared wins" among multiple matching candidates still relies on `getEnclosedElements()` declaration order, which `javax.lang.model` does not guarantee — left as the non-blocking residual noted in #662.

## Red/green evidence

New harness `FactoryMismatchWarningTest` (slice-processor-tests) drives the real `SliceProcessor` through the platform javac with the module's framework classpath and a `routes.toml` in the class output, so diagnostics and compilability of the generated routes are asserted from one compilation. Fixtures: a near-miss factory (`long` parameter vs `int` component), an exactly-matching control, and a type-use-annotated factory parameter.

RED — tests added, processor unfixed (`mvn -f jbct/pom.xml test -pl slice-processor,slice-processor-tests -am`):

```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0 ... FactoryMismatchWarningTest$NearMissFactory
[ERROR]   FactoryMismatchWarningTest$NearMissFactory.mismatchedParameterType_emitsFactoryFoundButUnmatchedWarning:121 [a factory-shaped method whose parameter types fail the component match must produce a visible diagnostic, not a silent validation skip]
[ERROR]   FactoryMismatchWarningTest$AnnotatedFactoryParameter.typeUseAnnotationOnFactoryParameter_isCosmeticAndStillMatches:150 [a type-use annotation changes the spelling, not the type; the factory must still be detected and used]
[ERROR] Tests run: 61, Failures: 2, Errors: 0, Skipped: 0
```

The two failures are exactly the warning assertion and the annotated-parameter match; the canonical-fallback-compiles and exact-match control tests were green, confirming the harness. The annotated failure also empirically confirms the issue's cosmetic-spelling near-miss on javac 25.

GREEN — with the fix:

```
[INFO] Tests run: 61, Failures: 0, Errors: 0, Skipped: 0
[INFO] JBCT Slice Processor ............................... SUCCESS   (303 tests, 0 failures)
[INFO] Slice Processor Integration Tests .................. SUCCESS
[INFO] BUILD SUCCESS
```

Zero regressions across both modules; no new warnings on any existing fixture (`grep -c "looks like a validating factory"` on the build log: 0).

## Envelope invariant

No `ENVELOPE_FORMAT_VERSION` bump: this change adds a compile-time diagnostic in `RouteSourceGenerator` and does not touch `FactoryClassGenerator`/`ManifestGenerator` output structure. Verified against unfixed→fixed builds of slice-processor-tests: all generated sources under `target/generated-sources/annotations` are **byte-identical** (`diff -r` clean), and all `META-INF/slice/*.manifest` files are structurally identical — the only differing lines are the per-build `#<date>` properties header and `generated.timestamp=`, which change between any two builds. The `isSameType` widening can change which construction path is emitted only for records whose factory previously text-mismatched — precisely the cases the issue defines as wrongly falling back.


siy:
**CTO review round — verdict: MERGE-WITH-NITS.** (1) Minor defect + coverage gap: the new boxing clause is BIDIRECTIONAL — the reverse direction (int param, Integer component) now matches and the generated call auto-unboxes the accessor: absent JSON field → NPE → 500 instead of typed 400. And no fixture exercises boxing at all — deleting the clause stays green, against the repo's mutation-check norm. Either restrict to the advertised direction or add fixtures for both. (2) Comment fix: the "identically-named type variables" fallback example is unreachable; the genuinely load-bearing fallback case is wildcard-containing spellings (isSameType refuses wildcards). (3) Verified clean: no false-positive completion path in the textual fallback; ERROR mirrors push toward match, not spurious warns; byte-identical proof valid for the fixture corpus; enclosed-elements ordering preserved.

siy:
CTO review fixes applied (commit e4a16752d, branch rebased onto `release-1.0.0-rc3` — `origin/release-1.0.0-rc3..HEAD` now carries only this PR's two commits; the stray docs commit is gone).

**1. Boxing clause restricted to the safe direction** (`RouteSourceGenerator.sameComponentType`). The clause was bidirectional; the reverse direction — primitive factory parameter (`int`) over boxed record component (`Integer`) — was a live defect path: the generated call auto-unboxes the accessor, so an absent JSON field became null → NPE → 500 instead of the typed 400 the factory path exists to produce. Now only a **boxed parameter over a primitive component** matches (`boxedParameterOverPrimitiveComponent`: accessor returns the primitive, the call site boxes — a total conversion). The refused reverse is factory-shaped, so it surfaces as the existing "looks like a validating factory but does not match" WARNING, not a silent skip.

**2. Comment accuracy.** The textual-fallback justification now names the real driver — wildcard-containing spellings, which `Types.isSameType` refuses — instead of the unreachable "identically-named type variables" example. The detection-rule doc and the CHANGELOG entry were updated to the direction-restricted wording.

**Fixtures + red/green evidence** (`FactoryMismatchWarningTest`):
- (a) `BoxedParamRequest` — boxed param over primitive component: matches, no warning, factory call asserted in generated routes. **Mutation-pinned empirically**: with the boxing clause deleted, `boxedParameterOverPrimitiveComponent_matchesWithNoWarning` fails (run shown red, clause restored).
- (b) `PrimitiveParamRequest` — primitive param over boxed component: must not match, must warn. **Red against the original bidirectional clause** (the `doesNotContain("PrimitiveParamRequest.primitiveParamRequest(")` assertion fired — the silent wrong match was real), green after the restriction.

**Verification:**
- Full suites `mvn -f jbct/pom.xml clean test -pl slice-processor,slice-processor-tests -am`: slice-processor **303** + slice-processor-tests **63** green, 0 failures, 0 skipped.
- Byte-identical re-check: clean-built the fixture corpus at `origin/release-1.0.0-rc3` (pre-#662 baseline) and at this branch's HEAD; `diff -r` over `target/generated-sources/annotations` — all 15 generated source files **byte-identical**. The restriction does not touch the existing corpus.

Issue #717, opened 2026-08-28
Slice dependency edges dropped at runtime: manifest pseudo-coordinates (version=UNRESOLVED) resolve to zero dependencies, then a wrong 'compiled against an older runtime' diagnostic


siy:
Split out of #704 per CTO instruction (defect 3 of 3, including the diagnostic-honesty half — #703 family); full experiment evidence: https://github.com/pragmaticalabs/pragmatica/issues/704#issuecomment-5454290447. Previously folded into #712 as a sibling; that ticket now covers only the packaging half.

**Defect, two coupled halves.**

*Half 1 — the dropped edge.* The slice manifest declares its slice dependency, but as a package-derived pseudo-coordinate the runtime cannot map to a deployed artifact:

```
dependencies.count=1
dependency.0.artifact=org.pragmatica.example.ticketing.eventmanagement.capacity:seatsellability   # real artifact: org.pragmatica.example:ticketing-seat-sellability
dependency.0.interface=org.pragmatica.example.ticketing.eventmanagement.capacity.seatsellability.SeatSellability
dependency.0.version=UNRESOLVED
```

The runtime resolver then logs, back to back, for the same artifact whose manifest says count=1:

```
Resolving 0 slice dependencies for org.pragmatica.example:ticketing-acquire-hold:1.0.0-SNAPSHOT with context
No slice dependencies for org.pragmatica.example:ticketing-acquire-hold:1.0.0-SNAPSHOT, creating slice directly with context
```

…and instantiates the factory without the dependency's classes in the loader context. Affected in ticketing: exactly the two SeatSellability injectors (acquire-hold, buy-ticket).

*Half 2 — the affirmatively wrong diagnostic.* The resulting NoClassDefFound is reported as:

```
Parameter mismatch in ...AcquireHoldFactory: slice was compiled against an older runtime
(references removed class org/pragmatica/example/ticketing/.../SeatSellability); rebuild against this runtime version — will NOT retry
```

Forge IS rc3 (verified via --version) and the class exists in the deployed sibling jar — the message sends a developer to rebuild against a runtime that is not stale, while the actual failure is the resolver never wiring a declared edge. The fix must correct the diagnostic as well as the resolution (or at minimum stop claiming a cause it has not established).

[...]

Issue #712, opened 2026-08-28
slice-processor packaging omits shared classes outside the slice package — slices fail at load with NoClassDefFoundError


siy:
Two independent reproductions today (2026-08-28), same mechanism, different repos.

**Mechanism.** The slice jar packaging includes the slice's own package plus the per-subsystem store classes, but NOT shared classes from outside those packages that the slice's generated factory references at construction — even when the slice's own manifest declares them. At load the factory hits `NoClassDefFoundError`, which the deployment FSM classifies as a deterministic failure (will NOT retry), and under default ALL_OR_NOTHING atomicity that rolls back the whole blueprint (`ClusterDeploymentState.java:1313-1334`) — one badly-packaged slice takes down the deployment for every clean slice too (the #704 churn).

**Reproduction 1 — ticketing (pragmaticalabs/ticketing @ 6a28282), rf=1 forge run** (full evidence: https://github.com/pragmaticalabs/pragmatica/issues/704#issuecomment-5454290447):
- `ticketing-sweep-holds`'s manifest declares `publish.message.classes=org.pragmatica.example.ticketing.shared.event.SeatReleased` and `publish.topic.0.messageType=...SeatReleased` — but none of the jar's 23 classes is that event record (verified by `unzip -l`).
- At load: `NoClassDefFoundError: org/pragmatica/example/ticketing/shared/event/SeatReleased` at `SweepHoldsFactory.lambda$sweepHolds$0(SweepHoldsFactory.java:36)`.
- All four publisher slices fail identically: sweep-holds, cancel-ticket (SeatReleased), set-price, adjust-price (PriceChanged).

**Reproduction 2 — notification-hub example (this repo), #672 e2e attempt** (details in PR #711's body):
- `ClassNotFoundException: org.pragmatica.aether.example.notification.NotificationEvent` → ALL_OR_NOTHING rollback.
- Verified: the class is present in the module jar but absent from both per-slice jars.

**Where to look.** The slice-jar class-selection logic in the packaging path (`PackageSlicesMojo` / the manifest's `impl.classes` collection in the slice-processor). The manifest's `publish.message.classes` entry proves the processor KNOWS the class is needed — the packaging step just doesn't copy it.

[...]

Issue #704, opened 2026-08-27
Capacity: does concurrent deployment of 24 slices at rf=3 on a single host stall activation under consensus backpressure? (activation itself verified working at 3 nodes)


siy:
Found while writing verified from-zero setup instructions for the ticketing reference application (pragmaticalabs/ticketing). Reproduced **three times independently** on a clean local Forge run.

**Verified symptom.** The cluster forms correctly — `CTM: Activated, desired=5, active=5, ready=5`, dashboard answers HTTP 200 — but **every slice instance stays `LOADED` and never reaches `ACTIVE`**: all 24 slices across all 5 nodes, `health=UNHEALTHY` 100%, observed via `GET :8888/api/slices/status`. Because no slice activates, no HTTP route is ever wired, so every documented API call returns 404. This includes slices with **zero errors in their own logs** (e.g. create-event), which is what makes it look like a platform-level stall rather than per-slice failures.

**Unconfirmed cause, stated as unconfirmed.** Rabia consensus backpressure was logged during the same window, including one `Consensus apply timed out after 30000ms`. That is a plausible contributing factor and nothing more — the run deployed 24 slices at 3 replicas each concurrently on a single machine, so resource exhaustion on one host is equally consistent with the evidence. No isolating experiment was run. `[design intent — unverified]` on any causal claim here.

**Why this matters beyond one laptop.** Ticketing is the reference application the book's Part III drafts against, and it is the first thing a new user runs. A from-zero developer who follows correct instructions today gets a cluster that reports itself healthy and an API that 404s — the worst diagnostic shape available, because every layer claims success.

**Separately fixed en route (in the ticketing repo, not here):** ticketing's `aether.toml` shipped with its entire `[database]` block commented out, which silently prevents every `@PgSql`-backed slice from activating at all (`PgSqlConnectorFactory.supports()` requires `async_url` to be present). That was a real defect and is fixed. The stall above persists **after** that fix, so the two are independent.

**What would settle the cause, roughly in cost order:**
1. Re-run with a much smaller slice count (or replica factor 1) on the same host — separates load-induced stall from a functional defect.
2. Instrument or inspect the `LOADED → ACTIVE` transition for the gate that never opens: is activation waiting on a consensus commit, a readiness probe, a resource provision, or a per-partition owner that never arrives?
3. If it reproduces at small scale, it is a functional defect in activation and outranks its milestone.

[...]

PR #719, opened 2026-08-28
fix: package manifest-declared message/event classes and full shared tree into slice jars (#712)

siy:
## Mechanism — what the closure missed and why

Slice jars are packaged by `PackageSlicesMojo.createImplJar` from exactly two sources: `SliceManifest.allImplClasses()` (name-based, with nested-class expansion) and `addSharedCode` (convention-based directory copy). Both had a gap, and topic message types fell through both:

1. **`SliceManifest.allImplClasses()`** (`jbct-core`) merged only `implClasses + requestClasses + responseClasses`. The manifest's own `publish.message.classes` and `stream.event.classes` keys — written by `ManifestGenerator` precisely so the runtime can register serializers for these types — were never parsed into the packaged set. Publisher message types are constructor-injected type arguments (`MessagePublisher<T>`, `StreamPublisher<T>`), so they appear in no method-derived list; nothing else could catch them.
2. **`PackageSlicesMojo.addSharedCode`** walked the sibling `shared` package with non-recursive `Files.list`, while the slice-subpackage walk right below it uses recursive `Files.walk`. A record in `shared.event` (ticketing's `SeatReleased`) or a component type in `shared.model` never reached the jar.

Reproductions explained: `ticketing-sweep-holds` declares `publish.message.classes=...shared.event.SeatReleased` but the jar's 23 classes omit it (gap 1 + gap 2) → `NoClassDefFoundError` in `SweepHoldsFactory.lambda$sweepHolds$0`. notification-hub's `NotificationEvent` (declared in `stream.event.classes`, living in the slice's own package top level, not a method param type) hit gap 1 alone.

**Both directions checked**: same-module topic *subscribers* were already covered — reactive methods are slice interface methods, so their single message param lands in `request.classes`; stream *consumers* (incl. batch) are folded into `stream.event.classes` by `collectStreamEventTypes`. Cross-module message-type delivery is the dependency-edge sibling #717, not this PR.

## Fix

- `SliceManifest` gains a `messageClasses` component: the union of `publish.message.classes` and `stream.event.classes`, folded into `allImplClasses()` (now deduplicated). Consumption-side only — `ManifestGenerator` is untouched.
- `addSharedCode` walks the sibling `shared` package tree recursively (`addPackageTreeClasses`), mirroring the slice-subpackage walk.
- Stale surfaces swept: `PackageSlicesMojo` class doc, `allImplClasses` doc, RFC-0004 inclusion rules, CHANGELOG.

## Closure depth

Declared message/event classes **by name + their nested/member classes** (`classToPathsWithInner`, same expansion every other named class gets) **+ the entire sibling `shared` package tree, recursively**. No bytecode reference-graph walk — the existing closure never walks references for request/response types either; transitive components are covered by the `shared`-tree convention, which the recursive walk now actually honors. A referenced type living outside `shared`, the slice package, and the declared lists remains out of closure — same rule as before this PR, now uniformly applied.

## Red/green evidence

RED (fix stashed, tests present) — jbct-core:

```
[ERROR]   SliceManifestTest.allImplClasses_publishMessageClassOutsideSlicePackage_includedInPackagedSet:38 publish.message.classes entry must be part of the packaged class set, got: [org.pragmatica.example.ticketing.sweepholds.SweepHolds, org.pragmatica.example.ticketing.sweepholds.SweepHoldsFactory, org.pragmatica.example.ticketing.sweepholds.SweepRequest, org.pragmatica.example.ticketing.sweepholds.SweepResponse] ==> expected: <true> but was: <false>
[ERROR]   SliceManifestTest.allImplClasses_streamEventClass_includedInPackagedSet:58 stream.event.classes entry must be part of the packaged class set, got: [org.pragmatica.aether.example.notification.NotificationService, org.pragmatica.aether.example.notification.NotificationServiceFactory] ==> expected: <true> but was: <false>
```

RED — jbct-maven-plugin (jar built from a ticketing-shaped fixture: message record + member class in `shared/event`, component in `shared/model`):

```
org.opentest4j.AssertionFailedError: manifest-declared publish message class must land in the slice jar, got: [org/pragmatica/example/ticketing/sweepholds/SweepHolds.class, ... org/pragmatica/example/ticketing/sweepholds/SweepHoldsFactory.class, org/pragmatica/example/ticketing/sweepholds/SweepRequest.class, META-INF/, META-INF/MANIFEST.MF, org/] ==> expected: <true> but was: <false>
```

GREEN (fix restored):

```
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0 -- in org.pragmatica.jbct.slice.SliceManifestTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- in org.pragmatica.jbct.maven.PackageSlicesMessageClassesTest
```

FULL suites (`mvn -f jbct/pom.xml test -pl jbct-core,jbct-maven-plugin,slice-processor,slice-processor-tests -am`):

```
[INFO] JBCT Parser ........................................ SUCCESS
[INFO] JBCT Core .......................................... SUCCESS
[INFO] JBCT Formatter ..................................... SUCCESS
[INFO] JBCT Linter ........................................ SUCCESS
[INFO] JBCT Maven Plugin .................................. SUCCESS
[INFO] JBCT Slice Processor ............................... SUCCESS
[INFO] Slice Processor Integration Tests .................. SUCCESS
[INFO] BUILD SUCCESS
```

Zero regressions. The jar-level GREEN also empirically confirms the archiver tolerates the now-systematic duplicate add (message class reachable via both `allImplClasses` and the shared tree) — jar builds, single logical entry per path asserted.

## Envelope statement

No manifest keys added, removed, or changed — `ManifestGenerator` output is byte-identical; the fix reads keys the manifest already carries. Adding classes to the JAR is not a manifest structure change, so no `ENVELOPE_FORMAT_VERSION` question arises; the GA freeze at 1000 (owner ruling #386, PR #708 pattern) is untouched.

Refs #712. Final e2e verification against the rf=1 reproduction (#704 setup) happens in the coordinating session before the ticket closes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [1.0.0-rc3] - Unreleased
 
+### Fixed (2026-08-28 — #712: slice jars ship the message classes their manifests declare)
+- **A slice jar now contains the topic message and stream event classes its own manifest
+  declares, plus the full sibling `shared` package tree.** Two packaging gaps compounded:
+  `SliceManifest.allImplClasses()` — the single source `PackageSlicesMojo` packages from — merged
+  only impl/request/response classes, never the manifest's own `publish.message.classes` and
+  `stream.event.classes` (publisher message types are constructor-injected `MessagePublisher<T>` /
+  `StreamPublisher<T>` type arguments, so they appear in no method-derived list); and the sibling
+  `shared` package walk was non-recursive (`Files.list`) while the slice-subpackage walk was
+  recursive, so records in `shared.event` and their component types in other `shared.*`
+  subpackages never reached the jar. Evidence: ticketing-sweep-holds' jar omitted the declared
+  `shared.event.SeatReleased` and slice activation died with `NoClassDefFoundError`;
+  notification-hub's `NotificationEvent` was present in the module jar but absent from the slice
+  jars. Closure depth: declared classes by name plus their nested/member classes plus the entire
+  sibling `shared` tree — no bytecode reference walk, matching the existing convention-based
+  closure. Manifest keys are unchanged (consumption-side fix only), so the envelope format is
+  untouched. Same-module subscribers were already covered via `request.classes` (reactive methods
+  are interface methods); cross-module message-type delivery is the dependency-edge sibling #717.
+  [verified: `jbct/jbct-core/src/test/java/org/pragmatica/jbct/slice/SliceManifestTest.java` and
+  `jbct/jbct-maven-plugin/src/test/java/org/pragmatica/jbct/maven/PackageSlicesMessageClassesTest.java`
+  (red against unfixed code, green after; jar-level assertion on a built archive); final e2e
+  against the #704 rf=1 reproduction runs in the coordinating session before #712 closes]
+
 ### Fixed (2026-08-28 — #644: periodic tasks arm in start(), not at assembly)
 - **A created-but-never-started node now performs no periodic work and holds no timers.**
   `AetherNode.assembleNode` used to hand all fourteen of the node's recurring tasks to
--- a/docs/rfc/RFC-0004-slice-packaging.md
+++ b/docs/rfc/RFC-0004-slice-packaging.md
@@ -53,8 +53,10 @@ commerce-order-service-1.0.0.jar
 │   ├── OrderResult.class               # Response types
 │   └── internal/                       # Subpackage classes
 │       └── OrderValidator.class
-├── org/example/shared/                 # Sibling shared package
-│   └── CommonUtils.class
+├── org/example/shared/                 # Sibling shared package (full tree, recursive)
+│   ├── CommonUtils.class
+│   └── event/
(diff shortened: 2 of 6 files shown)


siy:
**CTO review round — verdict: MERGE-WITH-NITS.** All four adversarial checks passed BY MECHANISM, not just corpus: shared parseList handles both key families with dedup and absent-key safety (backward-compat pinned exact set AND order); the recursive walk is structurally bounded under {parent}/shared with no symlink follow (cannot leak); envelope claim structurally true (no generator in diff, manifest copied verbatim); duplicate entries safe via plexus-archiver's hard-coded DUPLICATES_SKIP with byte-identical double-adds (transformation only touches *Factory at slice-package root, which no walk reaches). NITS for the next batch: (a) PR-body's "single logical entry per path asserted" is false — jarEntries uses HashSet; switch to a List-based count assertion; (b) addClassFiles silently skips missing declared classes (accepted #717 scope) — add a log line; (c) the jar test replicates createImplJar via reflection rather than invoking it — divergence risk worth closing. Queued for the serial merge round.

PR #730, opened 2026-08-29
feat: #386 D5 slice-processor half — context-carrying topic subscribers

siy:
Implements the D5 slice-processor half of #386 durable pub-sub, per the "Generated-shape spec for the D5 slice-processor half" comment on that issue. Stream B's companion records (`MessageContext`, `ContextualEvent`) landed in 07e7bab57; this branch is rebased onto that and compile-references them.

## The four spec items

**1. Detection.** A subscription method may declare the existing 1-arg shape (default, unchanged) or `Promise<Unit> handler(T event, MessageContext context)`, the second parameter matched by exact FQN `org.pragmatica.aether.slice.topic.MessageContext`. A same-named type from another package stays an ordinary arity error.

**2. Validation.** The 2-arg shape is a compile error unless the topic's `resources.toml` section declares `durability = "durable"`. `Topic<T>` carries no durability and `TopicConfig`/`TopicDurability` are bound at runtime, so there was no existing compile-time signal; `TopicDurabilityLoader` reads `resources.toml` from the class output — where `process-resources` has already placed it by the time annotation processing runs — mirroring how `RouteConfigLoader` already reaches `routes.toml`. No new dependency.

**3. Generated adapter.** For the 2-arg shape the `SliceMethod` takes a `ContextualEvent`, casts its erased payload and invokes the two-argument method; the adapter record's delegate override mirrors the declared two-argument signature. The 1-arg path is byte-for-byte unchanged, pinned explicitly.

**4. Manifest.** Additive `reactive.N.context = "message"`, emitted only for the 2-arg shape; absent means legacy. No structural change to existing keys. `ENVELOPE_FORMAT_VERSION` stays frozen at 1000, recorded in `envelope-versioning.md` with the "(no bump — GA freeze)" marker.

## Fail-closed, and it says which failure it hit

Durability that cannot be positively established is a refusal (ruled on #386; same shape as #661 / PR #709). The two causes are reported **distinctly**: a topic read and found ephemeral gets the "ephemeral dispatch carries no envelope" explanation, while an unreadable `resources.toml` says the durability could not be *determined* and was "refused rather than assumed". An author whose build lost the file must not be sent to fix a declaration that may already say `durable`.

That distinction is why `MessageContextRule` exists as a pure decision over an already-loaded durability view. The annotation-processor harness compiles in memory and can never present a readable `resources.toml`, so the declared-ephemeral branch — the one real authors actually hit — is unreachable from there and would otherwise have shipped with no pin at all. It also means the rule has no seam and no overload through which a permissive default could ever be supplied.

## MessageContext is not a payload parameter

Stated once, in `MethodModel.payloadParameters()`. Request-record synthesis, codec registration and `Topic<T>` payload matching therefore all keep seeing the event type alone — no phantom `<Name>Request` record is synthesized and none is registered for codec.

The one site that must NOT use that view is the adapter record's delegate override, which mirrors the declared interface signature. Getting that wrong was a real bug, caught by the real-compile fixture.

## Two judgment calls, both ruled

**Fail-closed on an unreadable `resources.toml`** — confirmed, with the distinguishable-cause condition above.

**A context-carrying handler may not also carry method interceptors** — a compile error the spec does not name. The chain is typed `Fn1<Promise<Unit>, T>` on the payload alone, so there is nowhere to put the context; the alternatives were dropping it silently (the lie D5 exists to prevent) or a cryptic javac error inside generated code the user never wrote (the #663 failure mode). The deeper design question is filed as #732 against the dispatch contract, post-GA. The guard is load-bearing rather than cosmetic: it is what makes `effectiveParamTypeString` and the interceptor-wrapper generation unreachable for these handlers, since all five of that method's callers are interceptor-gated. `generateWrapperRecord` carries a comment warning against "fixing" its branching, because the only way to make it compile with a 2-arg handler is to drop the context.

## Testing, and a note on method

`jbct/slice-processor` 340 tests, `jbct/slice-processor-tests` 75 tests, all green, no lint warnings.

The acceptance half of D5 is structurally unreachable from `SliceProcessorTest`, so it is proven by real-compile fixtures in `slice-processor-tests` — `com.example.durabletopic` (direct) and `com.example.durabletopicstep` (transitive through a plain-interface step) — where the module compiles its own generated sources and a wrong adapter is a build failure rather than a silent gap.

On how that was arrived at, because the method mattered more than the result here: **I had 331/331 green and could have shipped. Instead I asked whether the tests could REACH the code, deleted the entire adapter branch as an adversarial probe, and the suite stayed green — proving dead coverage. Then the real fixture found a genuine two-generation-site defect that would have made every real slice using this feature fail to compile.** The adapter record's delegate override was still being emitted with the 1-arg signature, so the generated record did not implement its own slice interface.

Coverage was verified adversarially rather than assumed: twelve mutations applied one at a time with full reverts, all now caught. The three that initially survived drove the strengthened loader assertions and both fixtures. Making the two failure causes distinct then turned two of my own tests red and revealed they had been named and commented as proving the ephemeral branch while actually exercising the undetermined one; both were retargeted.

**Reviewer note — a coverage weakness worth checking rather than taking on trust:** three D5 behaviours rest entirely on the `slice-processor-tests` fixtures, because `SliceProcessorTest` structurally cannot reach them under fail-closed. If CI does not run that module, those behaviours ship unguarded. I have not verified that CI does.

## Review round one — all four findings fixed (7d94cd597)

The reviewer generalized the two-generation-site defect and found three more instances. That was the right correction: I had fixed the site that failed to compile instead of enumerating every site keyed off parameter shape. I have now done that enumeration and classified each site as needing the DECLARED signature (adapter-record override, proxy method impls, wrapper-record impls) or the PAYLOAD view (request-record synthesis, `effectiveParamTypeString`, codec collection, `Topic<T>` matching, manifest metadata and `request.classes`).

**1. Interceptor guard scope — fixed by widening, not by re-branching.** The suggested alternative, branching the wrapper on declared parameters, produces `onOrderPlaced(OrderEvent e, MessageContext c) { return onOrderPlacedFn.apply(e); }` — it compiles and silently drops the context, the exact lie D5 exists to prevent; the interceptor bodies have no context in scope at all. So the guard moved to model scope, precisely: a subscription site now records whether it is direct, because the wrapper only walks the slice's own methods and a transitive handler must not be refused for an unrelated direct interceptor.

**2. `MessageContext` no longer leaks into `request.classes`** — `collectRequestTypes` uses the payload view, so the manifest stops advertising a type the generated `codec()` deliberately omits.

**3. Slice-dependency proxy — refused rather than quietly corrected.** A caller invoking a context-taking method through the slice-to-slice proxy has no envelope and could only fabricate a context. Refusing removes the phantom `dep_<Method>Request(T, MessageContext)` by construction instead of filtering it out, and follows the existing #612 non-Promise dependency-method rejection, naming the inheriting dependency when inherited.

**4. The byte-for-byte claim now matches what it proves** — the pin compares the complete emitted `SliceMethod` entry and the complete delegate override contiguously, indentation-normalized. Equality-grade for those blocks, which is what "unchanged" should mean; the earlier four substring assertions did not support the stronger wording.

All four were mutation-verified one at a time with full reverts: all four killed, no survivors.

### ⚠ Gate hazard found while fixing these

Maven skips annotation processing in `slice-processor-tests` when only the **processor** changed, since that module's own sources are untouched — so its fixtures re-validate artifacts a previous processor generated. I hit it live: the first gate after fixing the manifest leak still showed the leak, with the manifest on disk timestamped an hour before the edit.

Delete `jbct/slice-processor-tests/target` before gating any slice-processor change; it is documented in the fixture test class. CI is unaffected (fresh checkout). This sharpens the coverage note below rather than resolving it: finding 2 is caught *only* by that downstream module, so it is precisely the regression a stale target would pass.

## Not in this branch

The runtime half is stream B's: manifest-flag threading through registration to dispatch, the manager's context-aware branch, and the idempotency guard keyed `(projectionName, generation, messageId)`.


siy:
**CTO review round — verdict: REQUEST-CHANGES.** The author found and fixed a two-generation-site defect; adversarial review finds the same class at THREE more sites, one of which fails silently.

**1. CRITICAL — third generation site, unfixed.** `FactoryClassGenerator.generateWrapperRecord` (:268, emitted whenever `model.hasMethodInterceptors()` at :252) walks ALL methods and branches on `hasSingleParam()`, which this PR redefined to the PAYLOAD view (`MethodModel:258`). A context-carrying subscriber therefore gets a 1-arg override inside a record declared `implements <Slice>` — the record fails to implement its own 2-arg method. Same defect in the interceptor bodies at :637, :694, :827, which emit `Fn1<Promise<Unit>, OrderEvent> x = impl::onOrderPlaced;` — a 2-arg method reference assigned to a 1-arg `Fn1`. **Scope mismatch is the root cause: the new guard (`SliceProcessor:452`) is PER-METHOD (`site.method().hasInterceptors()`) while wrapper generation is PER-MODEL (`SliceModel:393`, `anyMatch`). So a slice with a contextual subscriber plus an interceptor on ANY OTHER METHOD generates non-compiling code** — and the sole interceptor test places the interceptor on the handler itself, so it cannot see this. Fix: branch on DECLARED params (as the adapter override now does), or widen the guard to model scope.

**2. MessageContext leaks into the manifest.** `ManifestGenerator.collectRequestTypes` (:448) flatMaps raw `parameters()`; `isStandardType` (:480) filters only `java.lang`/`java.util`. The manifest advertises a request type that no codec registers — directly contradicting this PR's own 'codec sees the event alone' pin.

**3. Slice-dependency proxy unaware of the shape — the silent one.** `ProxyMethodInfo` (:972) uses raw params, so a slice depending on another slice's contextual subscriber synthesizes a phantom `dep_...Request(OrderEvent, MessageContext)` (:1128) — exactly what the required no-phantom-request-record pin forbids. **It COMPILES, so it fails silently**; the fixture uses a plain-interface step and misses this path. The pin must be extended to cover the slice-dep path.

**4. Nit:** the 'byte-for-byte unchanged' claim rests on four substring assertions, not an output comparison. Either compare generated output or restate the claim to what the assertions actually prove.

**Verified sound, no action:** fail-closed is complete (`:496-507` → `.or(false)` at :464, runs before `generateFactory`; the unreadable case stays distinct from ephemeral, as ruled). And CI DOES build and test `slice-processor-tests` (`ci.yml:6,37`; reactor module via root pom:51 / jbct pom:54; no skip, no path filter) — so the three real-compile-only behaviours ship guarded. The author's honest coverage-location flag checks out and is NOT a blocker.

siy:
**CTO re-review — REMAINING-FINDINGS. The critical defect is DOCUMENTED, not fixed, and the unreachability claim it rests on is false.**

**Finding 1 — NOT FIXED.** The change is a 7-line comment (`FactoryClassGenerator:283-289`) asserting unreachability, and it reproduces the exact error originally flagged: refusal is per-METHOD (`SliceProcessor:269`, `site.method().hasInterceptors()`) while wrapper generation is per-MODEL (`:253`, `SliceModel:393` = `anyMatch`). **Counterexample: a slice with an interceptor on method A plus a contextual subscriber B carrying no interceptor still emits a 1-arg override of B's 2-arg signature — javac failure.** Same at :637/:694/:826.

**The five-caller claim is wrong in two distinct ways.** Caller 5 (`generateProvideWithContext:894`) is resource provisioning and is not interceptor-gated at all (harmless there, since the payload view is correct — but it falsifies the enumeration). The other four gate at MODEL scope, which never implies the contextual METHOD was refused. Model-scope gating cannot establish method-scope unreachability; that is the same scope confusion as the defect itself, one level up.

**Finding 2 — NOT FIXED.** `ManifestGenerator.collectRequestTypes:448` still flatMaps raw `parameters()` and `isStandardType:480` does not filter it, so `request.classes` still lists MessageContext. The new pin asserts only `doesNotContain("OnOrderPlacedRequest")` — it cannot catch this, so the pin is passing without covering the claim it was added for.

**Genuinely good and retained:** `MessageContextRule` preserves fail-closed (`none()` → violation), adds no permissive overload, and genuinely pins declared-ephemeral, undetermined, their distinctness, and interceptor. Nit 4 (the byte-for-byte claim resting on substring assertions) remains unaddressed.

**REQUIRED before the next review pass:** make it unreachable rather than argue it is — branch on DECLARED parameters at every wrapper/interceptor site, or widen the refusal to model scope — and add the falsifying pin: a fixture slice with an interceptor on one method and a contextual subscriber on another, which must COMPILE. An unreachability claim needs the same evidence standard as any other claim; a comment is not a mechanism.

siy:
**CTO confirmation pass — CONFIRMED-FIXED, 4/4.** Two cosmetic defects introduced by the fix commit must clear before merge.

**The divergence is verified as built and holds.** All four interceptor sites (`generateWrapperRecord:268`, `:637`, `:694`, `generateInterceptorWrapping:795`) walk `model.methods()` only, so the direct/transitive split tracks reachability exactly; the direct refusal condition (`site.direct() && sliceModel.hasMethodInterceptors()`) is the wrapper's own gate verbatim, so no slice shape can place a contextual handler in a wrapper record, and transitive handlers are correctly not refused for an unrelated direct interceptor. **Recording why the reviewer's original suggestion was declined, since it is the more instructive half:** branching on declared params would have emitted a wrapper that COMPILES and SILENTLY DROPS the delivery context — the precise lie this feature exists to prevent — and the interceptor bodies have no context in scope at all, so making them compile would require fabricating one. A fix that satisfies the compiler by discarding the payload of the feature is worse than the defect it closes.

**(2)** `collectRequestTypes:451` uses the payload view; fixture `request.classes` went "OrderPlaced,MessageContext" → "OrderPlaced". One raw `parameters()` remains at `:325` (stream branch) and is benign — it takes `getFirst()`, which is the payload under both shapes. **(3)** The slice-dep refusal cannot hit a legitimate shape (`collectProxyMethods` runs only for non-resource, non-plain-interface deps, so plain-interface steps never reach it), and the phantom record is gone BY CONSTRUCTION — `continue` fires before `ProxyMethodInfo` is built, matching the #612 precedent. **(4)** The restated claim matches what the assertion proves.

**REQUIRED BEFORE MERGE — two hygiene defects in `FactoryClassGenerator`:** at `:1660` the signature of `reportNonPromiseDependencyMethod` and its first statement are merged onto one line; and the pre-existing #612/#663 javadoc block at `:1632-1638` now sits above `takesMessageContext`, so it documents a boolean predicate while the method it actually describes is left undocumented. Cosmetic, but a doc block that describes a different method than the one beneath it is the same defect class this PR spent four rounds eliminating — clear both, then this merges.

Issue #610, opened 2026-08-18
Routing state has no repair path: endpoint ReplicatedMaps run at DHTConfig.FULL with anti-entropy disabled, so a missed write persists until the key is next written


siy:
Source: design review of Aether at 1.0.0-rc3 (2026-08-17); owner flagged this one as under review. Wiring re-verified 2026-08-18.

## What

Endpoints, per-node slice state and HTTP routes are DHT `ReplicatedMap`s (`aether/aether-dht/.../AetherMaps.java:22-24`) instantiated at `DHTConfig.FULL` (`aether/node/.../AetherNode.java:1213` — verified).

`FULL` is W=1/R=1 and, per its own docstring (`integrations/dht/.../DHTConfig.java:113-120`), *"eventually consistent, NOT linearizable"* with **anti-entropy and rebalancing disabled**.

## The design choice is right; the gap is repair

Eventual consistency is correct for routing — a stale route just fails over, and paying consensus latency would be wrong.

But with anti-entropy off, a node that misses a write holds a stale endpoint set **until the next write to that key**, with nothing to reconcile it. Ephemeral keys and normal churn probably mask this. "Probably" is doing real work in that sentence, and that is the whole issue.

## Do this first, regardless of which option wins

**Write the failing test**: drop a write on one node, assert convergence *without* a subsequent write to that key. Today that test would hang.

It is cheap, it is falsifiable, and it converts "probably masked by churn" into a measurement. Choosing the repair mechanism is a design decision that should not block establishing whether the bug is reachable at all.

## Options, cheapest first

- **(a) Ride SWIM.** Membership gossip already reaches every node and carries incarnation numbers — the same propagation and repair properties endpoint state needs. Removes the DHT from the routing path entirely. Most elegant, largest change, couples routing to membership.
- **(b) Enable anti-entropy for these maps only.** Cheap, bounded, reversible, keeps the current structure.
- **(c) Move endpoints into consensus.** Correct but pays write throughput; almost certainly wrong for routing.

(b) is the reversible one and (a) is structural — worth deciding explicitly rather than sliding into.

## Acceptance

- The convergence test above exists and passes.
- The chosen mechanism is recorded with its reasoning, including why the rejected options were rejected.

siy:
Closing on evidence, not on a fix; draft PR #842 carries the evidence and does not merge.

[...]

PR #842, opened 2026-09-04
evidence: #610 routing maps are never written in production; DHT FULL has no repair

siy:
## What this is

A draft, non-mergeable evidence PR for #610. It contains one failing test and no production code. Opened to make the investigation record public and reviewable; per team decision this closes as a dead-surface finding rather than a repair.

## #610's premise, as filed

`endpoints`, `slice-nodes`, and `http-routes` are DHT `ReplicatedMap`s at `DHTConfig.FULL` (W=1/R=1, eventually consistent, anti-entropy explicitly disabled for that config). The ticket's claim: a node that misses a write holds a stale copy until the key is next written, with nothing to reconcile it.

## What the test proves

`integrations/dht/src/test/java/org/pragmatica/dht/DHTFullReplicationConvergenceTest.java` drives a 3-node in-process `DHTClient` cluster directly at `DHTConfig.FULL`, drops one node's copy of a put, and asserts convergence without a further write. It fails on its 3-second bound, every time — burning the full timeout, not erroring out. **This confirms the DHT-level mechanism has no repair, exactly as the ticket describes.** `[mechanism: DHTAntiEntropy.runAntiEntropy() returns immediately when config.isFullReplication() — hard-guarded off by design]`

## What further investigation found: the mechanism has no production trigger

None of the three named maps has a single `.put()` call site in production code anywhere in the repo.

- `endpoints()` / `sliceNodes()` / `httpRoutes()` (`aether/aether-dht/.../AetherMaps.java:22-24`) have zero non-test callers outside their own declarations.
- `EndpointRegistry.registerEndpoint` (`aether/aether-invoke/.../EndpointRegistry.java:129`) writes a **local in-memory** `Map<EndpointKey,Endpoint>` — never reaches the DHT map.
- `ClusterDeploymentState.trackSliceState` (`aether/aether-deployment/.../ClusterDeploymentState.java:1391-1393`) writes a **plain in-process** `Map<SliceNodeKey,SliceState>` field — never reaches the DHT map.
- HTTP route publication (`aether/aether-invoke/.../HttpRoutePublisher.java:305-315`) goes through **Rabia consensus** (`KVCommand.Put` against `NodeRoutesKey`, a different key type) — never reaches the DHT map.
- The only two classes that reference these DHT maps directly for read/write — `DhtNodeCleanup` and `CachedReplicatedMap` (both `aether/aether-dht`) — are themselves never instantiated in production.
- The only live wiring for all three maps is the **receive side**: `AetherNode.java:2637-2651` (message-router entries for `PutRequest`/`RemoveRequest`) → `AetherNode.java:5363-5380` (`handleRemotePutResponse`/`handleRemoteRemoveResponse`) → `aetherMaps.dispatchRemotePut`/`dispatchRemoteRemove` → `NamespacedReplicatedMap.onRemotePut`/`onRemoteRemove`. Every node can accept a replicated write for these namespaces; no node ever originates one.
- `DHTConfig.FULL` has exactly one other non-test reference, `aether/ember/.../EmberCluster.java:1027` — it passes the same config into the same `AetherNodeConfig`/`AetherNode` construction path, not a separate consumer.

The authoritative, linearizable state for this data lives entirely in `KVStore` (`integrations/cluster/.../kvstore/KVStore.java`, implements `StateMachine<KVCommand<K>>`, Rabia-backed) and in-process structures — a different subsystem, with no bridge class connecting it to `aether-dht`'s eventually-consistent maps.

## Conclusion

The three DHT maps are permanently empty in production, so the ticket's specific failure mode ("stale until next write") cannot occur via them today — there is no first write either. The underlying DHT-level defect is real (see the committed test) but has no live surface. Recommend closing #610 on this evidence and tracking removal of the three dead maps plus their receive-side wiring as a separate dead-surface cleanup.

## Scope

- 1 commit, 1 file: a failing test only.
- No production code.
- Not for merge.

### DIFF
--- a/integrations/dht/src/test/java/org/pragmatica/dht/DHTFullReplicationConvergenceTest.java
+++ b/integrations/dht/src/test/java/org/pragmatica/dht/DHTFullReplicationConvergenceTest.java
@@ -1,0 +1,190 @@
+/*
+ *  Copyright (c) 2020-2025 Sergiy Yevtushenko.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.pragmatica.dht;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import org.pragmatica.consensus.NodeId;
+import org.pragmatica.consensus.ProtocolMessage;
+import org.pragmatica.lang.Option;
+import org.pragmatica.lang.io.TimeSpan;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.pragmatica.dht.DHTAntiEntropy.dhtAntiEntropy;
+import static org.pragmatica.dht.DHTNode.dhtNode;
+import static org.pragmatica.dht.DistributedDHTClient.distributedDHTClient;
+import static org.pragmatica.dht.storage.MemoryStorageEngine.memoryStorageEngine;
+
+/// #610: endpoints / per-node slice state / HTTP routes are DHT `ReplicatedMap`s run at
+/// `DHTConfig.FULL` (W=1/R=1, eventually consistent) with anti-entropy and rebalancing disabled.
+/// [DHTAntiEntropy#runAntiEntropy] bails immediately on `config.isFullReplication()`, so a node
+/// that misses a write holds a stale value with nothing to reconcile it until the key is next
+/// written. This is the ticket's required first step: drop ONE node's copy of a SINGLE write at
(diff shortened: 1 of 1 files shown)


siy:
Closing per the team decision recorded in this PR's own body: a dead-surface finding, not a repair. #610 is already closed. This PR is non-mergeable by construction (one deliberately failing test, no production code), so leaving it open keeps a permanent red in the PR list and misreads as CI breakage.

Re-verified the load-bearing absence claim at the rc4 tip before closing, because 'zero non-test callers' is exactly the claim shape that gets believed without checking.

**Instrument validated first.** An initial run returned 0 for a `.get(` / `.put(` control — impossible — because zsh glob-expanded an unquoted `--include=*.java`. Every number in that run was an artifact. Re-run with quoting: control `.get(` = 2220, `.put(` = 1160 in src/main.

**Result — claim holds, with one correction.**

| accessor | src/main callers | src/test callers |
|---|---|---|
| `endpoints()` | 0 (see below) | 6 |
| `sliceNodes()` | 0 | 5 |
| `httpRoutes()` | 0 | 3 |

The one apparent `.endpoints(` production hit is `aether-config/.../ConfigLoader.java:524` — `builder.endpoints(Map.copyOf(endpoints))`, an unrelated config-builder method, not `AetherMaps.endpoints()`. Counting it would have understated the finding.

**The finding is sharper than the PR states.** `AetherMaps` is not unwired dead code — it is *constructed in production* at `aether/node/.../AetherNode.java:1441` (`AetherMaps.aetherMaps(dhtClient.scoped(DHTConfig.FULL))`), with all twelve serializer/deserializer references bound, and it is threaded as a parameter at `AetherNode.java:5505` and `:5516`. The wiring is live and terminates in nothing: the object is built and passed around, and its accessors are never called.

That distinction decides what happens next, and it is not settled here: dead weight to remove, or routing that silently never had a backing store. A follow-up ticket carries that question; removal must not proceed until it is answered, since removal breakage travels toward dependents.

Evidence preserved in this closed PR and in the follow-up. Branch retained for now.

Issue #654, opened 2026-08-26
Consumer cursor commit at detach/close is a discarded Promise — consensus write failure is silent, redelivery contract undocumented


siy:
Found 2026-08-26 while documenting `StreamConsumerRuntime.close()` for the #642 stop-ordering fix (review item M2).

`ConsumerRuntimeState.flushCursorForKey` calls `ConsumerCursorStore.commit(...)`, which returns `Promise<Unit>` — a consensus write — and the call site **discards it unobserved**: no await, no onFailure, no log, no retry. A failed cursor commit at consumer detach (including the node-stop path, where it races the tail of shutdown) is silent; the observable consequence is redelivery from the last successfully committed offset on the next attach. Consumers must therefore be idempotent for this window [design intent — unverified].

Options (not ruled): observe the promise and log failure loudly at detach; bound-await it inside the #488 ordering window (cursor store + partition manager are still alive there post-#642-fix); or document redelivery-on-detach-failure as the contract and say so in the consumer docs. Whichever lands, the current silent discard should become deliberate rather than accidental.

Context: the #642 batch moved `streamConsumerRuntime.close()` inside the #488 ordering constraint (before `streamPartitionManager.close()`), so the commit now runs while its dependencies are alive — this ticket is about the discarded outcome, not the ordering.

siy:
Fixed by PR #845, merged into release-1.0.0-rc4 on 2026-09-05. Merged-base push-CI concluded SUCCESS on b16ba5688 across build-and-test, forge-tests and fragment, so the eight-PR combination is verified, not just each PR individually. Closing manually because GitHub does not auto-close linked issues when a PR merges into a non-default branch.

Issue #488, opened 2026-07-18
Streaming: declarative [streams.X] consumer registration is a dangling KV write — no delivery loop consumes it; ConsumerRuntimeState (auto-resume push loop) is test-only


siy:
Surfaced by the #478 review's production-path trace (2026-07-18). The declarative stream-consumer surface is half-built:

- slice-processor recognizes `[streams.X]` consumer methods (MethodModel.java:117,539) but FactoryClassGenerator uses them ONLY for codec collection (:2196) — no generated subscribe/fetch/delivery call.
- `NodeDeploymentState.buildStreamSubscriptionPutCommand` (:1231-1244) publishes `StreamRegistrationKey → StreamRegistrationValue(node, consumerGroup, batchMode, eventType)` into consensus KV on every deployment — and NOTHING reads it back (only writer + key/value/codec definitions reference the type). A dead consensus write per deployment.
- `ConsumerRuntimeState` — the real push loop with attach-time cursor auto-resume + ~1000-event/30 s checkpointing (subscribe :77 → loadCursorAndStart :88 → deliverEvents :253) — has no production constructor: `streamConsumerRuntime(...)` is called only by its own interface factories and tests.
- Net: live app consumption is stateless caller-supplied-offset pull; `StreamAccess.fetchFromCommitted` (#478) is the resume contract but currently has zero production callers — apps must opt in.

Work: decide the target model (wire registration → delivery loop → generated consumer dispatch with auto-resume, presumably via ConsumerRuntimeState or its successor), or excise the dangling registration write if declarative delivery is not the plan. Either way reconcile the TWO coexisting resume mechanisms (one-shot `fetchFromCommitted` vs unwired `ConsumerRuntimeState`) — reviewer flagged the coexistence for eventual reconciliation.

The arch-examples CQRS projection (arch-examples-spec-draft) consumes via explicit poll + fetchFromCommitted today and becomes the natural acceptance test when declarative delivery lands. Relates: #478, #474 arc, streaming plan A4.

siy:
[...]

PR #845, opened 2026-09-04
fix: observe cursor commit outcome on consumer detach and checkpoint (#654)

siy:
## Summary

`ConsumerRuntimeState.flushCursorForKey` called `ConsumerCursorStore.commit(...)` — a consensus write — and discarded the returned `Promise<Unit>` unobserved: no await, no onFailure, no log, no retry. This ran on every consumer detach, including the node-stop path inside the #488 ordering window. `checkpointIfNeeded`'s periodic commit had the same defect.

## Changes

- `close()` bound-awaits every commit that is in flight at the moment `awaitFinalCursorCommits()` takes its snapshot: `Promise.allOf(commits).await(CURSOR_COMMIT_SHUTDOWN_BOUND)`, a 5-second bound. Node stop always continues past the bound; it never blocks on a wedged consensus write `[mechanism: Promise.allOf derives its own aggregate promise; .await(TimeSpan) on timeout returns a manufactured Timeout result without cancelling or resolving the underlying commits]`.
- Both `flushCursorForKey` and `checkpointIfNeeded` route through `observedCommit` → `issueTrackedCommit`, which mints a `TrackedCommit` (key, state, the commit's own `Promise`, and a fresh per-commit `AtomicBoolean reported`) and registers it in a new `inFlightCommits` registry for the life of that one commit.
- **The dedup token is scoped per COMMIT, not per consumer.** `reportCommitOutcome` is the single funnel all three report paths — `reportIfUnsettled` (bound expiry), `onCursorCommitFailure` (genuine failure), `recordIfRecovered` (a recovered checkpoint-publish failure) — route through: `tracked.reported().compareAndSet(false, true)` decides ownership. The CAS winner increments `cursorCommitFailureCount` once and logs once at ERROR; every later report for the SAME commit loses the CAS, does not increment again, and logs at WARNING ("late resolution of an already-reported cursor commit") instead. This closes the bug an earlier version of this fix had: the token lived on `ConsumerState` and was shared by every commit that consumer ever issued, so a periodic checkpoint commit still in flight when `close()` issued that same consumer's final commit shared one flag — the periodic commit's later genuine failure could be silently uncounted (0 increments where 1 was owed), or a check-then-act race between the bound expiry and the commit's own resolution could double-count (2 increments where 1 was owed).
- A commit that starts (or restarts, via a periodic checkpoint tripped by a delivery already in flight when `close()` begins) *after* the bound's snapshot is taken is not bound-reported at all: its own token still lets a genuine later failure or recovered checkpoint publish count it exactly once, but if it never settles it is simply never counted and stays in `inFlightCommits` for the life of the runtime.
- Surfaced through the existing read surface: `StreamConsumerRuntime.cursorCommitFailureCount()` (node-wide, survives detach) and `SubscriptionSnapshot.lastCursorCommitFailure` (per-consumer, cleared on detach). On the Management API, `DeclarativeConsumerPartition` gained `lastCursorCommitFailure` rather than overloading `DeclarativeConsumerDetail.diagnostic`.
- `ConsumerCursorStore` gained a default `lastRecoveredFailure(...)` method so a composed store (e.g. `ClusterCursorStore`, which chains the consensus checkpoint publish onto the local commit's own `Promise`) can report a recovered sub-stage failure without a push-callback. The two failure origins are distinguishable on the same surface: `lastCursorCommitFailure` is prefixed `local commit:` or `checkpoint publish:` depending on which write actually failed.
- Redelivery contract documented in `management-api.md`: a commit that misses the shutdown bound is treated as failed for that shutdown even if it later succeeds; on failover the consumer resumes from the last consensus-published checkpoint, at least once.

## Verification

- **Guard-revert probe**: reverting the per-commit CAS guard in `reportCommitOutcome` causes `StreamConsumerRuntimeTest.CursorCommitObservability#close_countsUnsettledCommit_whenFinalCommitNeverSettlesWithinBound` (aether-stream) and `StreamConsumerRuntimeClusterCursorTest#close_countsUnsettledCommit_whenConsensusPublishNeverSettlesWithinBound` (aether/node) — both of which assert the count stays at 1 after a late resolution — to instead report 2, proving the guard is what prevents the late resolution from double-counting.
- **Per-commit-vs-per-consumer probe**: reintroducing the earlier bug class — a single `AtomicBoolean` dedup token shared per consumer key (simulated via `ConcurrentHashMap.computeIfAbsent`) instead of a fresh token minted per commit in `issueTrackedCommit` — causes `close_countsBothUnsettledCommits_whenPeriodicAndFinalCommitShareOneConsumer` to fail: expected 2L, got 1L. This proves the test discriminates the per-commit-vs-per-consumer scoping bug rather than passing vacuously.
- Both probes were single-hunk, isolated production-code edits, run against the already-committed correct code, and reverted via `git checkout --` before any official verification build; neither probe's edit was ever committed.
- Fresh `build-runner` verification (`HCLOUD_TOKEN` unset throughout, no manual `mvn verify`): reactor `install` BUILD SUCCESS; `jbct:check` clean on both `aether-stream` (0 format issues, 0 lint errors) and `aether/node` (0 format issues, 0 lint errors); `StreamConsumerRuntimeTest` 31/31 green, `StreamConsumerRuntimeClusterCursorTest` 5/5 green (counts from surefire XML `<testcase>` elements written by this run, not the `.txt` summary header, which undercounts `@Nested` classes).
- [ ] End-to-end redelivery-after-timeout on a live multi-node cluster with failure injection — not run this PR `[design intent — unverified]`, flagged as such in the changelog fragment.

Closes nothing automatically (release branch); tracks #654.


siy:
Note for the merger: commit `da64de35c` implements `ManageableNode.schemaOrchestrator()` in three test doubles — this is the #855 base fix, duplicated here so the branch test-compiles; it is byte-identical to commit 1 of the #854/#855 PR, so merging that PR first makes this commit a no-op merge. #855 is attributed to that PR, not this one.

Issue #893, opened 2026-09-06
ContentStoreFactory.close is a second silent no-op override: reports success while releasing nothing, and no existing ticket covers it


siy:
Found during the full audit of all 23 non-test `ResourceFactory` implementations, 2026-09-06 (#268/#891 stream).

`ContentStoreFactory:44-46` overrides `close` with a no-op returning `Promise.unitPromise()` — **the identical shape to #271**, on a different resource, and tracked nowhere until now.

## Why it is separate from the neighbouring tickets

The audit found **4** overrides of `close`, not 3: two real (`NotificationListenerFactory` — unlisten plus `connection.close()`; `DurableEntityFactory` — `fenced.unload()`, logged on throw) and **two no-ops** (`NotificationSenderFactory`, which is #271, and this one).

- **#891** concerns the *default* dispatch missing `AsyncCloseable` and the `DatabaseConnector.stop()` convention. This factory overrides the default, so #891's fix never reaches it.
- **#271** is the same defect on `NotificationSenderFactory` — a different file and a different resource.
- **#268** concerns the cache and consumer machinery, not what a factory's own close does.

So none of the three covers it, and each of their fixes leaves this one intact.

## Note on provenance

An earlier design spec classified `ContentStoreFactory` as *closing* its resource, from a crude two-line grep — and flagged its own classification as a heuristic rather than a result. **The heuristic was wrong**: it does not close, it no-ops. Worth recording as a small instance of the pattern this codebase keeps producing, where a plausible-looking check reports the opposite of the truth.

## Verified vs inferred

- **Verified by reading:** the override at `:44-46` and its `unitPromise()` body; the audit's classification of all 23 implementations, done by reading rather than grep alone, with positive controls stated.
- **Inferred, NOT executed:** that a `ContentStore`'s underlying resources are observably retained after release. Nobody has measured it. Note also that a retention measurement will show retention regardless until **#892** lands, since `releaseAll`'s identity can never match a registered consumer — so a null result there proves nothing about this ticket.

## Acceptance

1. `ContentStoreFactory.close` actually releases whatever the content store holds, or — if there is genuinely nothing to release — the override is **removed** so the resource falls through to the default dispatch, and the reason is recorded. A no-op override that silently reports success is worse than no override, because it is indistinguishable from a working one.
2. A test asserting the underlying resource's release **actually ran**, not that the returned promise succeeded — it already does.
3. Audit the other three overrides against the same standard while here.

[...]

Issue #892, opened 2026-09-06
Two slices from one artifact share one registered identity (Artifact.asString()), so the first unload closes for both


siy:
Found by the #268/#891 stream, 2026-09-06; **mechanism verified independently on disk by the CTO from generated code, not from reading the generator.**

This is the **third** independent reason no provisioned resource is ever closed, and it sits downstream of the other two. Fixing #268 and #891 does not produce a close in production without this.

## The mismatch

**Consumer registration** — `SliceLoadingContext.resources()` wraps the provider in `SliceAwareResourceProvider` using `delegate.sliceId()`, which `DependencyResolver:92-95` sets to `artifact.asString()`. From `Artifact.java:48-50`:

```java
public String asString() {
    return groupId + ":" + artifactId + ":" + version.withQualifier();
}
```

Three colon-separated segments, including a version.

**`releaseAll` argument** — the generated slice's `stop()` passes a compile-time literal built by `FactoryClassGenerator.computeSliceArtifactCoordinate` (`:1764-1770`):

```java
return groupId + ":" + artifactId + "-" + toKebabCase(sliceName);
```

Two segments, no version, and the slice name kebab-cased onto the artifactId.

## Evidence from generated code on disk

40 generated factories in the tree; 34 emit `stop()`/`releaseAll`, and every one carries the two-segment shape:

```
org.pragmatica-lite.aether.example:inventory-inventory-service
org.pragmatica-lite.aether.example:payment-payment-service
org.pragmatica.aether.example:catalog-catalog
org.pragmatica.aether.example:pg-showcase-order-processing
```

A registered consumer for the first would be `org.pragmatica-lite.aether.example:inventory:<version>`. The two strings differ in **segment count** and in the **artifactId segment itself**, so this is not a near-miss a version-strip would rescue. **0 of 34 can ever match.**

## Consequence

`consumerSet.remove(sliceId)` removes nothing. The consumer set never empties, the cache entry is never dropped, and `close` is never called — for any resource, on any slice, ever.

**Three independent defects on one release path, each individually sufficient, each hiding the next:**

1. **#268** — production `provide()` bypassed `promiseCache`, so `releaseAll` drained an empty map. Close never attempted.
2. **#891** — even when attempted, the default dispatch missed the project's own `AsyncCloseable` convention (and the `DatabaseConnector.stop()` third convention). Close no-ops.
3. **This** — the release call's identity can never match a registered consumer, so the release path never reaches the close at all.

[...]

Issue #271, opened 2026-06-11
@Notify: factory.close() noop leaks SMTP event loop; retries permanent failures; config-key/example mismatches


siy:
## Findings R8+R9+R10 (resource-surface assessment 2026-06-11)

1. **R8 (High):** `NotificationSenderFactory.close()` returns `unitPromise()` (`:35-37`), overriding even the AutoCloseable default → `SmtpClient.close()` (owns a Netty `EventLoopGroup`, `integrations/net/smtp/SmtpClient.java:84-89`) never called → event-loop leak per unload (compounds the SPI leak #268).
2. **R9 (High):** retry is indiscriminate — permanent failures (HTTP 401/403 `AuthError` `HttpEmailSenderCore.java:70-72`; SMTP 5xx) get the full exponential schedule. No transient/permanent split.
3. **R10 (Medium):** the sole consumer (`examples/notification-hub/notification-emailer/resources.toml`) sets `username`/`password`/`pool_size`/`helo_hostname` under `[notification.smtp]`, but `SmtpConfig` expects nested `auth Option<SmtpAuth>` and has no poolSize/heloHostname (`SmtpConfig.java:27-33`) → keys silently ignored, credentials this way would **send unauthenticated**. Retry helpers (`sendWithRetry`/`nextDelay`/`delayThen`) duplicated verbatim across both senders; `delayThen` spawns a raw virtual thread per retry instead of a shared scheduler.

## Acceptance
- `close()` closes the underlying sender (SMTP event loop released); test it.
- Permanent failures (auth, 4xx/5xx-permanent) not retried.
- Example + `resource-reference.md` SMTP/HTTP key tables corrected to match `SmtpConfig`/`HttpEmailConfig` (folded into the docs ticket).
- Retry helper deduplicated, shared scheduler.

siy:
[...]

PR #898, opened 2026-09-06
fix: prune interceptor registries after release

### LINKED ISSUE #894: CacheInterceptorFactory and IdempotencyInterceptorFactory retain factory-level registries keyed by name that no release path prunes
Found during the full audit of all 23 non-test `ResourceFactory` implementations, 2026-09-06 (#268/#891 stream).

`CacheInterceptorFactory` and `IdempotencyInterceptorFactory` hold **factory-level** registries keyed by name — `cacheRegistry` and `storeRegistry`/`claimRegistry`, keyed by `cacheName` / `storeName` — that **no release path ever prunes**.

## Why this is not covered by the close-dispatch tickets

Every other leak on this release path concerns a *resource* not being closed. This one is different in kind: the leaked state lives on the **factory**, not on the resource, so it is untouched by all of:

- **#891** — the default close dispatch missing `AsyncCloseable` / `DatabaseConnector.stop()`
- **#268** — cache population, consumer refcounting and `releaseAll`
- **#892** — the release identity never matching a consumer
- **#271 / #893** — individual factories whose `close` no-ops

Even with every one of those fixed and a resource correctly closed, the factory's registry entry for its name remains. The growth is **per distinct cache/store name over the node's lifetime**, unbounded, and independent of how many slices are running at any moment.

## Why the audit classified these as "nothing to close"

The interceptor and stream/publisher factories provision plain in-memory objects, which legitimately have no close. That classification is correct **for the resource** and is what makes this easy to miss: the resource is genuinely disposable, and the leak is in the factory that made it. A close-dispatch audit looks straight past it.

## Verified vs inferred

- **Verified by reading:** the registries exist, are factory-level, are keyed by name, and no release path prunes them.
- **Inferred, NOT executed:** the growth rate and whether it is material on a real node. Nobody has measured it. It is plausibly slow — bounded by the number of *distinct names*, not by request or slice volume — so this may be a long-horizon leak rather than an urgent one. **Establish that before sizing the work**; the honest answer may be that it matters only for long-lived nodes with churning blueprint names.

## Acceptance

1. A registry entry keyed by a cache/store name is removed when nothing holds it any more, or the retention is deliberate and documented with its bound.
2. A test that provisions under a name, releases, and asserts the registry does not retain the entry.
3. Determine whether the two factories should share one mechanism — they have the same shape and the same defect, and fixing them differently invites the next one to be different again.

Cross-refs: #268, #891, #892, #271, #893 — all resource-close defects on the same release path; this is the only one where the leaked state is on the factory.

mikemikimike:
## Summary

- replace the factory-level cache, store, and claim maps with a shared reference-counted named registry
- retain shared resources while interceptors use them and remove each entry after the last release
- add release regression tests for cache and idempotency resources

## Verification

- baseline regression test: 2 failures, with both registries remaining at size 1 after the final release
- targeted Maven regression test: 2 tests passed
- full reactor test for `aether/resource/interceptors`: 25 modules and 82 interceptor tests passed
- JBCT format and lint check: 0 format issues and 0 lint errors; existing warnings remain

Closes #894


siy:
Thank you for this — it is the first external contribution to the repository, and it targets the right place: the two factory-level registries in `aether/resource/interceptors` really are never pruned (#894). The `acquire`/`release` mechanism is sound: both are single `compute` calls, so concurrent acquires and a release racing an acquire are serialized correctly, and the two tests go red under every mutation we tried (no-op release, deleted `close` override, prune-on-first-release, non-incrementing acquire).

Before it can merge, four things:

**1. `null` is not used in this codebase.** `CacheMethodInterceptor` and `IdempotencyMethodInterceptor` gain a fourth record component and a 3-arg constructor that passes `null` for it. This codebase follows Java Backend Coding Technology (JBCT), and absence is always expressed as `Option<T>`, never `null` — a record component that may be absent is `Option<String>`, and a constructor that has no value passes `Option.empty()`. The same applies inside `NamedResourceRegistry`. Please read the JBCT rules before revising: the JBCT course at https://pragmatica.dev/java/jbct/course/ is the free canonical; https://pragmatica.dev/java/jbct/course/four-return-types/ is the lesson on this exact rule (T / `Option` / `Result` / `Promise`), and https://pragmatica.dev/java/jbct/course/error-handling/ covers typed errors, which the registry's release path will need. The complete argument is the book at https://leanpub.com/jbct-book. In this repository, `CONTRIBUTING.md` explains the JBCT format+lint gate and `build.sh` runs it locally; `jbct/README.md` documents the linter and its rules.

**2. Tooling.** The `jbct-coder` agent under `ai-tools/` in https://github.com/siy/coding-technology is written for Claude Code; there is no Codex version. Its rule sections and its forbidden-patterns checklist are tool-independent and can be folded into an `AGENTS.md` or a system prompt. Ignore its header's library version; the course is the reference for what is current.

**3. A double release steals a sibling's reference.** The count carries no holder identity; the identity check is on the backend, which every holder of a name shares. Scenario: slices A and B both hold idempotency store `orders` (count 2). If `close` runs twice for A — a retried release, a node-shutdown close, a test — the count reaches 0 and the entry is pruned while B is live. Slice C then provisions `orders` and receives a new store *and* a new claims map, so B and C execute the same "idempotent" operation independently under one key. `close` is public API, so the registry has to defend this itself. Suggested shape: an identity-keyed set of holders per entry instead of an `int`, with release idempotent per holder. It must be identity-based, not `equals`: the interceptors are records with structural equality, and two provisions of the same config are `equals` to each other. Surfacing the `boolean` from `release` (both factories currently discard it) would make a wrong release visible.

**4. Changelog fragment and sequencing.** Every source-touching PR needs `changelog.d/894-<slug>.md` (format and examples in `changelog.d/README.md`; each bullet ends in `[verified: <TestClass#method>]` or `[mechanism: …]`); the `fragment` job fails deterministically without it. Also, on the current base the new `close` override cannot be reached: `releaseAll` only closes what the plain `provide` overload cached, and generated slice code uses the context overload, which caches nothing. PR #900 (in review) changes that. There is no file overlap and the two compose, but this PR should land after #900 and the fragment should say that "Closes #894" holds from that point.

Smaller, not blocking: #894's third acceptance item asked for one mechanism across both factories, which is half met — one class, but the idempotency factory still runs two independent registries for store and claims rather than one entry holding both.

Happy to answer questions here.

siy:
Re-reviewed at `0d08e5c23`. All three findings are closed, and the two subtle ones are closed properly:

- **Holder identity.** The holder set is `IdentityHashMap`-backed and the holder is the interceptor instance, so two interceptors that are `equals` to each other still count as two holders — which is exactly the trap here, since these are records with structural equality. The new registry test pins it with two `equals()` holders rather than asserting the happy path. That is the right test, not just a passing one.
- **`null` is gone.** `Option<String>` components with `Option.empty()` in the 3-arg constructors; zero `null` in the added lines.
- **Changelog fragment** is present and well-formed.

Scope and provenance re-checked at the new head: the two merge commits carry no contributor content (each compares clean against both its parents), and everything you authored sits inside `aether/resource/interceptors` and `changelog.d`. Nothing outside scope, nothing unexpected.

One item left, and it is a wording change rather than code:

**Fragment bullet 3 overstates what a deployed node will do.** It says pruning happens "with #900 in place". #900 is now merged, so the `close` override is reachable through `closeThroughOwningFactory` — that part is right. But #892 is still open: on a deployed node `releaseAll` computes a 2-segment slice coordinate while registration uses the 3-segment versioned artifact, so it matches no key and the close never runs. Your fix is correct and the tests prove the mechanism; it simply cannot be *observed* in production until #892 lands. Please name #892 in that bullet and say the pruning is proven by test and gated in production on #892. Everything else in the fragment reads accurately.

`Closes #894` is earned by mechanism and test on that understanding.

One process note so the wait is not a mystery: workflow runs on pull requests from forks need a maintainer to approve them before they start, which is GitHub's default for first-time contributors and not a judgement about the change. Approval is the repository owner's step, not something a reviewer can do.

Thanks for turning this around so quickly, and for taking the `Option` point rather than working around it.

siy:
Two small house-style points to fold into the same push, so you are not sent back a second time:

- **Doc comments use `///`, not `/** */`.** The tree has moved to JEP 467 markdown doc comments; the new `NamedResourceRegistry` uses the older block form. Any file in `aether/resource/` shows the convention.
- **`release` returns a `boolean` that both factories discard.** Not wrong, but a release that matched nothing is exactly the condition worth noticing, and dropping the return makes it invisible. Either use it or say in a comment why it is ignored.

Also worth knowing: the PR description still describes the first revision (the baseline-regression numbers and the plain `Map` implementation), which no longer matches the code. Updating it keeps the record straight for whoever reads this in the changelog later.

None of these block anything; they are the last of it.

siy:
CI has run. `build-and-test` and the changelog check passed. `forge-tests` failed, and **it is not your change** — I have re-run it.

The failure is `MultiSourceCommunitySmokeTest`, a class-level error at 515.7 s. That is a known hang in the multi-node cluster test suite, tracked as #915: two untimed lifecycle awaits that, when one does not settle, run out the 480 s job backstop instead of failing by name. It hit two of our own PRs earlier today with the same ~515 s signature, and it cannot be reached from `aether/resource/interceptors` — the module your PR touches — at all.

Two related tickets if you are curious about why a hang like that survives a test-framework timeout: #914 (`Promise.await()` never checks the interrupt flag, so an interrupted waiter spins instead of stopping) and #916 (a transient activation race classified as permanent). Both were found this week while chasing a different flaky test; neither is anything you need to act on.

So: nothing for you to do about the red. The one outstanding item on this PR is still the fragment bullet naming #892 as the production gate, plus the two style notes above whenever you next push.

siy:
The rerun is green — all three jobs pass at `0d08e5c23`, confirming the earlier red was the known hang and not your change.

That leaves only the fragment bullet naming #892 as the production gate, plus the two style points, whenever you next push.

siy:
Re-reviewed at `2d00b6127`. All three items are closed and there is nothing further from my side.

- The fragment's bullet 3 now names #892 and states it correctly: the pruning is proven by test, and observation on a deployed node is gated on #892.
- `///` doc comments in the registry; the swap is syntax-only, with `acquire`, `release` and the identity-keyed holder set untouched.
- The discarded `release` boolean is explained by comment, and the explanation is accurate.

Nothing else moved under cover of those changes — one comment line in each factory, the comment-style swap, and one fragment bullet.

The remaining step is not yours: workflow runs need a maintainer to approve them again after each push while a contributor is new to the repository, so CI on this head is waiting on that. Merging is the repository owner's decision.

Thank you for three quick, precise revisions, and for taking the `Option` and holder-identity points rather than working around them. This is the first outside contribution the project has had, and it has set a good standard for the next one.

siy:
Merged as `a1cccd27b`. Thank you — this is the first external contribution to the project, and it set a good standard for whatever follows.

Worth recording what it took: three revisions inside a few hours, each one addressing specific feedback rather than working around it. You took a house rule you had no way to infer from the code (absence is `Option`, never `null`), and you got the identity-versus-structural-equality point right — using an identity-keyed holder set so two interceptors that are `equals` to each other still count as two holders, pinned by a test with two structurally equal records. That subtlety is the one a reviewer expects to have to explain twice.

**One thing to know about what happens next, since it affects when this becomes observable.** The pruning is proven by your tests, but on a deployed node it is still gated by #892: `releaseAll` computes a 2-segment slice coordinate while registration uses the 3-segment versioned artifact, so no consumer matches. Until that lands, the close path does not run in production, which is why the changelog fragment says so rather than claiming a user-visible fix. Your change is what makes the fix correct when the gate opens.

If you want more of this, the tickets tagged against the resource lifecycle are the adjacent ones — #892 itself, #893 and #271 are all in the same release path. No obligation, and thanks again.

Issue #923, opened 2026-09-07
Consensus unavailability rolls back a deployment: exhaustion raises an untyped cause at 90s, inside the 120s chain timeout, so it classifies Fatal


siy:
## What

When consensus is genuinely unavailable, the deployment path exhausts its consensus retries after ~90 s (3 × 30 s) and raises a **plain `Causes.cause`**. That cause is not a `SliceLoadingFailure` and not a `CoreError.Timeout`, so `SliceLoadingFailure.classify` takes the catch-all and returns `Fatal.UnexpectedError`. The deployment is marked permanently failed and the blueprint is rolled back — for a condition that is, by definition, an outage rather than a defect in the deployment.

The 90 s exhaustion sits inside a 120 s chain timeout, so it always wins the race: the chain timeout (which would classify `Intermittent` via `CoreError.Timeout`) can never fire first.

## Where

`aether/aether-deployment/.../fsm/NodeDeploymentState.java:1752`, `:1839`, `:1862` — the consensus retry exhaustion sites. Classification at `aether/slice-api/.../SliceLoadingFailure.java` (`classify`'s catch-all).

## Why it matters

An operator who deploys during a consensus outage — or whose cluster loses quorum for 90 s mid-deployment — gets a permanently failed, rolled-back blueprint rather than a deployment that resumes when consensus returns. Recovery is manual and the record says the deployment failed, not that the cluster was unavailable.

**Not reachable under ordinary load.** A healthy cluster commits in milliseconds; this needs 90 s of genuine unavailability, not slowness. So it is an outage-recovery defect, not a source of everyday rollbacks, and it does **not** roll back deployments that would otherwise have succeeded under normal conditions.

## Acceptance

Consensus unavailability is classified intermittent (or is otherwise distinguished from a deployment-level fault) so the deployment is retried or held rather than permanently failed and rolled back. State what happens if consensus never returns — the answer must be a terminal state, not an unbounded retry (see #922, where retry exhaustion currently re-drives roughly every second forever). Pin it with a test that makes consensus unavailable for longer than the exhaustion window and asserts the deployment is not marked permanently failed.

## Evidence

[...]

Issue #916, opened 2026-09-07
A transient activation race is classified as permanent: classify()'s catch-all Fatal rolls back the blueprint under ALL_OR_NOTHING


siy:
## What

`SliceLoadingFailure.classify` treats every cause it does not recognise as **permanent**, so a *transient* activation race is converted into a fatal failure and the leader rolls the whole blueprint back under `ALL_OR_NOTHING`. The "deterministic" classification that drives the rollback is a catch-all default, not a judgement about the failure.

## Where

- `aether/slice-api/src/main/java/org/pragmatica/aether/slice/SliceLoadingFailure.java:166-180` — `classify(Cause)` returns the cause itself if it is already a `SliceLoadingFailure`, maps resource-capacity and `CoreError.Timeout` to `Intermittent`, and otherwise falls through to `return new Fatal.UnexpectedError(cause)`. `isFatal()` is true for `Fatal`.
- `aether/aether-deployment/.../fsm/NodeDeploymentState.java:139` — `SLICE_NOT_FOUND_FOR_ACTIVATION` is a plain `Causes.forOneValue("Slice %s state is ACTIVATE but not found in SliceStore")`. It is neither a `SliceLoadingFailure` nor a `CoreError.Timeout`, so it takes the fallthrough. Raised at `:412`.

## Why it matters

The race is reachable through the same public endpoints an operator uses: delete a blueprint, then re-apply the same artifact before the unload has completed. The new ACTIVATE crosses the in-flight unload, the node reports the slice is not in the store, the leader classes that as permanent and rolls back — so a **retryable timing collision permanently fails a deployment** and the operator sees a rolled-back blueprint with no indication that retrying would have worked.

Found while diagnosing #727: this is the mechanism behind that flake (reproduced under CPU load on a quiet 16-core host, 240.5 s), but the flake is only how we noticed it. PR #913 fixes the **test** so it no longer produces the race; this ticket is the product half and is deliberately separate.

## Acceptance

[...]

Issue #805, opened 2026-09-03
Schema activation gate pre-filters the in-memory blueprints map before the shared predicate; bestEffortFailureCommand merges by read-then-Put


siy:
Two non-blocking residuals from the round-3 review of the #760/#724/#759 branch (2026-09-03), recorded so they are not lost when that branch merges.

1. **Gate pre-filter on the in-memory map.** `blocksSliceActivation` now resolves `schemaRequired` from the KV store inside the shared predicate (`ClusterDeploymentState.java:145-154`), and both the gate and `SchemaRoutes.heldSlices` call it. But the gate still pre-filters the in-memory `blueprints` map before reaching the shared predicate, while the route does not, so a stale map can make the gate block a slice the route reports as not held: the reverse of the divergence #760's first round fixed. Fix: one path, no pre-filter, or the route applies the same pre-filter from the same source; pin with a test that plants a stale map entry.

2. **`bestEffortFailureCommand` read-then-Put.** `ClusterDeploymentState.java:1473-1482` merges the failing-slice list into the outcome record by reading the current value and putting a merged one, outside any compare-and-set. Two BEST_EFFORT slice failures landing concurrently can lose one slice id from the record. Fix: fold the merge into the consensus batch as an append-style command, or CAS on the record's version; pin with two concurrent failures.

Acceptance: both pinning tests red without the fix, green with it; no behaviour change for ALL_OR_NOTHING.

siy:
Closing as fixed and merged.

- **Fix commit:** `c6f0bcdd54939d839e5d11642fbc8027fcb23495`
- **PR:** #956
- **Ancestry:** confirmed an ancestor of `release-1.0.0-rc4` — `git merge-base --is-ancestor c6f0bcdd54939d839e5d11642fbc8027fcb23495 6fd0002362622a1da18ebd8a02c54d65f1849e7d` exits 0 (rc4 tip `6fd0002362622a1da18ebd8a02c54d65f1849e7d`, read from `git ls-remote` at 2026-09-11).

The schema activation gate no longer pre-filters the in-memory blueprints map ahead of the shared predicate — one owner source now decides — and `bestEffortFailureCommand`'s read-then-Put is fenced so the outcome merge cannot lose a concurrent update.

Issue #698, opened 2026-08-27
SliceTargetValue owner erased at reconstruction in ControlLoopContext.applyScaling and AbTestManager.targetPreservingOverrides — schemaRequired still reverts on real autoscale


siy:
Follow-up from #555. #555 fixed `ClusterDeploymentState`'s consumer-side resolution of `schemaRequired` (`handleSliceTargetChange`, `restoreAppBlueprint`, `restoreSliceTarget`) to resolve via the owning blueprint instead of hard-defaulting to `true`. That fix is correct and complete at its layer, but it only helps when the `SliceTargetValue` it consumes actually carries an owner.

## The bug

Two producers reconstruct a `SliceTargetValue` from scratch instead of using the owner-preserving `AetherValue.withInstances(...)`, and both hardcode `Option.none()` for the owner where the real owner is available or derivable:

- **`ControlLoopContext.applyScaling`** (module `aether-control`, `ControlLoopContext.java` ~line 516) — builds a fresh `SliceTargetValue.sliceTargetValue(artifact.version(), newInstances, newInstances, Option.none(), currentBlueprint.maxInstances(), ...)` on every autoscale decision. Call path: `applyDecisions` (:454) → `prepareGuardedChange` (:471) → `prepareChange` (:492) → `prepareChangeToBlueprint` (:498) → `applyScaling` (:516) → collected (:463) → `cluster.apply(commands)` (:465), which reaches consensus and persists.
- **`AbTestManager.targetPreservingOverrides`** (module `aether-invoke`, `AbTestManager.java` ~line 274) — same pattern. Its own javadoc claims overrides "must not evaporate," but the owner itself evaporates on every override application.

## Consequence

#555's consumer-side fix (`ClusterDeploymentState.handleSliceTargetChange`) resolves `schemaRequired` from the owner when one is present, and correctly preserves it for **manual scale**, because `SliceRoutes.java:427-453` and `RollbackManager.java:406-411` build their `SliceTargetValue` via `withInstances(...)`, which preserves the owner.

A **real autoscale event**, however, still reaches `handleSliceTargetChange` with `owner = Option.none()` (because `ControlLoopContext.applyScaling` erased it at the source), so the flag falls through to the historical unowned-slice default (`true`). `schemaRequired = false` still silently reverts to `true` on the first autoscale event after deploy — #555's original symptom, just moved one layer upstream.

## Class of bug

[...]

PR #924, opened 2026-09-07
fix: make deployment retry exhaustion terminal instead of re-driving forever (#922)

### LINKED ISSUE #922: Retry exhaustion on the transient deployment path abandons the blueprint without a rollback or a record (ALL_OR_NOTHING silently broken)
## What

When a deployment fails with an **intermittent** cause, `handleTransientFailure` retries it. If the retries are exhausted, the deployment is abandoned **without a rollback**: no `ROLLED_BACK` record is written and the partially-applied blueprint is left in place. `ALL_OR_NOTHING` is silently not honoured on that path.

## Where

`aether/aether-deployment/.../fsm/` — the transient-failure path (`handleTransientFailure` and its exhaustion arm), contrasted with `handleDeterministicFailure`, which does trigger `rollbackBlueprintForArtifact`. Precise lines in `oss/internal/rc4-AR-916-report-2026-09-07.md`.

## Why it matters

This is the failure mode that justified a design decision, so it should not stay hypothetical. While fixing #916 the question was whether `classify()`'s catch-all should default to intermittent instead of fatal. It was ruled to stay fatal **precisely because** an intermittent default would retry a genuinely fatal cause a few times and then abandon it with no rollback — silently breaking `ALL_OR_NOTHING`. That reasoning is sound, and it means the same hole already exists today for every cause that is *legitimately* classified intermittent (resource-capacity and `CoreError.Timeout`, plus the store-absence causes typed by #916).

So: a cluster genuinely short of capacity for long enough to exhaust the retries ends up with a partially-applied blueprint and no record saying so. Distinct from #916 — that one is about which causes are permanent, this is about what happens when a correctly-classified transient one runs out of attempts. Also related to ordering 2 of #916's race, which reaches "permanently failed, no rollback, no record" by a different route.

## Acceptance

Retry exhaustion on the transient path ends in the same terminal state as a deterministic failure: rollback attempted, outcome recorded, and the operator can see from the blueprint status that the deployment did not apply. Pin it with a test that exhausts the retry budget on an intermittent cause and asserts a `ROLLED_BACK` (or equivalent explicit) record rather than silence. State the retry budget and where it is configured, since the acceptance depends on it being finite and knowable.

## Evidence

Identified by the #916 stream while ruling on the catch-all default (`oss/internal/rc4-AR-916-report-2026-09-07.md`, residual). Read-verified against the FSM; **not** reproduced by exhausting a real retry budget — doing so is part of acceptance.

siy:
> ## ⚠️ STATUS — head `7941ed668`, round 3. DO NOT MERGE.
>
> **A fourth adversarial verification is in flight.** Two prior reviews BLOCKED this PR.
>
> **The diagnosis below is still accurate and worth reading. The section "The fix, and the scope the
> review made necessary" describes a SUPERSEDED approach** — it was rejected by round-2 verification
> and no longer matches the code.
>
> **What the code does now.** The present-tense predicate (*is an instance ACTIVE?*) is replaced by a
> past-tense one, `everReachedActive` — a fail-safe disjunction of three legs, settling only when all
> three say the artifact never reached ACTIVE, and gated on `coreMembershipResolved()` first:
>
> | Leg | Evidence | Lifetime |
> |---|---|---|
> | 1 | an instance is ACTIVE now | survives failover via the `sliceStates` rebuild |
> | 2 | `everActiveArtifacts` — this leader watched it start | in-memory; empty on a new leader |
> | 3 | the owning blueprint's durable `DeploymentOutcomeValue.SUCCEEDED` | durable KV; survives failover |
>
> **Why round 2 blocked the previous approach:** asking *"is an instance ACTIVE right now?"* cannot
> work, because `handleSliceFailure` removes the failing key from `sliceStates` **before** either
> failure branch runs. For a single-instance slice — the most common deployment shape — the guard was
> structurally unreachable, so every such slice hit by a >31 s transient was condemned permanently and
> cluster-wide, with its last outcome record still reading SUCCEEDED and reconcile emitting nothing.
> That case self-healed before this PR.
>
> **The reframing that produced round 3:** the livelock and the self-heal are the *same mechanism*.
> Before #922, budget exhaustion cleared the counter and re-drove without settling — which is how a
> transient recovered, and equally how `ArtifactNotFound` spun forever. The discriminator has to
> separate *was up and fell over* from *never came up*, which is a question about history, not about
> the present.
>
> **Four residues are stated open in the fix report, not claimed closed** — including one nothing
> covers (a re-used blueprint id whose attempt spans a leader failover reads the previous attempt's
> record). All four fail toward **re-driving** rather than condemning, which is the reversible side of
> a one-way door. That is a deliberate judgment call and is marked as one.
>
> **CI green on this PR is on the OLD head `8d9b8beb9`** and says nothing about `7941ed668`.
> Mergeability is not checks-green.
>
> Fix report: `oss/internal/fix-924-round3-2026-09-07.md` · Round 2: `verify-924-round2-2026-09-07.md`

Closes #922.

**Extracted from #920 after a CTO ruling that the terminal-state change belongs to #922, not #916.** It was written and verified inside #920 first, then moved here intact. #920 types two activation-path causes as `Intermittent`; this changes what happens when *any* intermittent cause exhausts its retry budget. Different tickets, different blast radius, so they ship separately with their own fragments.

**Updated after review (1 BLOCKING, 2 SHOULD-FIX, 4 NOTE).** The blocking finding was real and is fixed and pinned; the first version of this PR traded a livelock for a worse, silent failure. Details below.

**Why this is live, not theoretical.** An artifact coordinate that cannot be resolved — a typo in a blueprint is enough — makes the cluster re-drive the load roughly once a second for the life of the process, logging exhausted retries that nothing acts on. There is no terminal state to stop it and no record an operator can act on.

**Same-mechanism family.** #916 / #920 (activation-path causes misclassified as permanent), this ticket (the exhaustion arm never settles), and **#923** (consensus exhaustion raises an untyped cause at 90 s, inside the 120 s chain timeout, so it classifies `Fatal`) are three faces of one thing: what a deployment failure is *typed* as, and what the machinery does with that type. #923 is not addressed here.

## The defect: exhaustion never reached a terminal state

`handleTransientFailure` retries a non-fatal failure five times with exponential backoff. When the budget was spent, `logMaxRetriesExceeded` cleared the retry counter and routed a `DeploymentFailed` event — and returned. It never added the artifact to `permanentlyFailed`.

That is not a resting state:

1. `handleSliceFailure` issues an unload on **every** failure, before the fatal/transient branch;
2. the node's `handleUnloading` removes the `NodeArtifactKey`;
3. the leader's `handleSliceNodeRemoval` — finding the artifact *not* permanently failed — schedules a reconcile;
4. `reconcileBlueprint` is gated on `permanentlyFailed` and nothing else, so it redeploys the artifact and `retryCounters.merge` restarts at 1.

So an intermittent cause that never settles was re-driven at roughly **1 Hz for the life of the cluster**: no terminal state, no rollback, no outcome record, and consensus round-trips forever.

## The fix, and the scope the review made necessary

`handleDeterministicFailure`'s tail — mark `permanentlyFailed`, route `DeploymentFailed`, honour the declared atomicity — is extracted as `settleAsPermanentlyFailed`, and **both failure branches now converge on it**. `logMaxRetriesExceeded` is renamed `handleRetryBudgetExhausted`, because it no longer merely logs.

**But settling unconditionally was itself a defect, and a worse one than the livelock.** `settleAsPermanentlyFailed` came from the **deterministic** branch, where concluding cluster-wide from a single node is sound *because a deterministic failure is node-independent*. **That inference does not transfer to the transient branch** — a transient failure on node A says nothing about node B. The defect was not the choice of flag but a **scope inference carried across a branch where it does not hold**:

- `retryCounters` is keyed per artifact **and node** (`slices/<nodeId>/<artifact>`);
- `permanentlyFailed` is a cluster-wide `Set<Artifact>`, cleared at exactly one site, on blueprint apply.

So roughly **31 s** of trouble on **one** node condemned a slice that was deployed and healthy on every other node. And it did so **silently**: a fully-deployed blueprint has already left `inFlightBlueprints`, so `rollbackBlueprintForArtifact` matches nothing, **no FAILED outcome is written at all**, and the artifact's last recorded outcome still reads SUCCEEDED while reconcile quietly refuses to replace the lost instance. The livelock was noisy and kept trying; this was a permanent, invisible wedge on a deployment that had been working.

`handleRetryBudgetExhausted` now settles only when **`hasActiveInstanceElsewhere` is false**: a deployment **attempt** with nothing ACTIVE anywhere is abandoned, while a **running** workload suffering a node-local transient keeps being reconciled toward its desired instance count.

### Why that discriminator and not the blueprint being in flight

Gating on `inFlightBlueprints` was the cheaper test, and it is the wrong one. `ClusterDeploymentContext.newActive()` constructs **every** collection empty, and only a live `AppBlueprintPutReceived` populates that map — nothing repopulates it after a leader change. Gating on it would have **silently reopened #922 after any failover**. `sliceStates`, by contrast, is rebuilt from durable KV entries by `rebuildSliceStateFromKVStoreEntries` during `rebuildStateFromKVStore` in `Active.onEntry`, so the predicate reads the same answer on a new leader as on the old one. Verifying a condition's meaning is not verifying its lifetime.

Two smaller points, both deliberate: the check is **strictly ACTIVE** rather than `isLiveState`, because a sibling instance merely LOADING or LOADED belongs to the same unproven attempt and must not vote to keep it alive; and it filters on `activeNodes()`, the identical filter `getCurrentInstances` uses, so the predicate agrees by construction with what `reconcileBlueprint` counts as running.

### What this deliberately leaves open

- A transient affecting **every** node for longer than the budget still settles the artifact, since by then nothing is ACTIVE anywhere. The fix narrows the evidence required for a cluster-wide permanent verdict from one node's budget to every node's; it does not remove the case. Removing it entirely would mean never settling a previously-healthy artifact — the livelock with no terminal at all.
- A running workload that cannot reach its desired instance count keeps retrying at the reconcile cadence. That is convergence, not this defect, but it is a candidate for backing off rather than terminating.
- The retry budget is still not operator-configurable (review S1). Threading a configurable `MAX_RETRIES` touches ~20 construction sites or needs a late-wired setter; that is a feature, not a fix.

### An open design question this fix does not answer, and the reviewer should press on

**This interacts with #926, and the interaction is not resolved here.** #926 reports that cluster-failure events are leader-gated: measured over ten days on the internal test host, a five-node cluster produced 8 SWIM fault detections and **zero** `NodeFailed` emissions, with `LeaderLost` firing 70–86 thousand times while `LeaderElected` never fired at all. A cluster that cannot elect a leader therefore emits nothing about its own failure.

Put that together with this change. With no leader, deployments cannot progress, so a deployment started during such an outage drains its retry budget with nothing ever reaching ACTIVE. `hasActiveInstanceElsewhere` then reads "nothing ACTIVE anywhere" — correctly, but **for the wrong reason** — and settles the artifact permanently failed. Per #926, nothing is emitted about any of it.

That may well be the right outcome: the deployment genuinely did not succeed. But a permanent state cleared only by a fresh blueprint apply is a strong thing to reach through an infrastructure condition the operator cannot see, and distinguishing "this deployment cannot work" from "this cluster currently has no leader" needs a signal #926 says does not exist. It is an argument for fixing emission first and treating the two as one design question. **Deliberately not addressed here** — it is out of scope for #922 and #926 now carries it.

## Retry budget — corrected

`MAX_RETRIES` = 5, backoff 1/2/4/8/16 s jittered, about **31 s on one node**. Both it and `MAX_RETRY_DELAY_SECONDS` = 30 are private compile-time constants: finite and knowable, but **not operator-configurable**.

Two corrections the review was right about:

- **The 30 s cap never binds at the current budget.** `Math.min(1L << (retryCount - 1), 30)` with `retryCount` never exceeding 5 tops out at 16 s. It guards a larger budget; it does not shape this one.
- **The bound is per artifact and node, not cluster-wide.** The sixth reported failure on any one node spends that node's budget, but with placement rotating across N allocatable nodes the cluster-wide total can reach **N x (MAX_RETRIES + 1)** reported failures — and the bound holds only while the allocatable node set is stable, since fresh nodes get fresh counters.

**Operator recovery action:** the artifact stays undeployable until its blueprint is redeployed, which clears `permanentlyFailed` on apply.

## Verification

The pin drives **`CoreError.Timeout`**, deliberately: one of the two causes the ticket names as already legitimately `Intermittent`, independent of #920's typing work, so the fix cannot be said to depend on it.

**The reverse-risk pin asserts the RECOVERY, not the absence of a rollback.** "No rollback happened" is equally true of a cluster that has silently stopped doing anything — which is the bug itself — so such an assertion would pin nothing. `exhaustionOnOneNode_whileTheArtifactIsActiveElsewhere_stillRecoversTheLostInstance` deploys the slice on two nodes, spends the entire budget on one, then asserts that reconciliation still re-drives the artifact **and** that the lost instance comes back ACTIVE on both nodes once the transient clears. A companion test asserting the standing SUCCEEDED record is explicitly labelled non-discriminating in its own javadoc, and nothing is relied on it.

**Red-before, two probes.** Each production hunk is pinned by its own test, and neither masks the other. Both mutations compiled, so both are genuine test failures rather than broken probes. The change was committed *before* probing and each probe reverted as a hunk against the committed base; restores were verified by content hash, not by `git status`.

| Probe | Mutation | Result |
|---|---|---|
| A | `hasActiveInstanceElsewhere` guard removed — settle unconditionally, i.e. the reviewed defect | `Tests run: 960, Failures: 1` — only the new recovery pin. Both original #922 pins stayed green |
| B | `settleAsPermanentlyFailed` call replaced by the pre-#922 log-and-return tail | `Tests run: 960, Failures: 2` — only the two original #922 pins. Both new tests stayed green |

| Check | Result |
|---|---|
| `aether-deployment` tests | **960**, 0 failures |
| `RetryExhaustionTerminalTest` | 4, all pass |
| `jbct:check -Djbct.skip=false` | 0 format issues, 0 lint errors |
| Base | merged forward onto `f3abe5c5b`, which contains #916 (`2cb38bc2d`) and #921 |
| Root reactor `mvn clean install` | **BUILD SUCCESS** — 20:39, **13,153 tests** from surefire XML, **0 modules SKIPPED**, 0 `[ERROR]` lines. Base `f3abe5c5b`, head `8d9b8beb9` |

**Base merged forward before the root build, deliberately.** #916 routes both slice-store-absence causes onto the **transient retry path** — precisely the path this change governs. A root build against the old base would have been evidence about a base nobody will run.

**Residual, stated rather than implied:** I did not construct a test for a transient exhaustion racing an unload/activate crossing from #916. The harness drives the leader FSM through ordered KV notifications, so "mid-crossing" is not a state it can hold — the predicate reads whatever the last notification left. Recorded as a residual.

## Stale surface corrected here

`aether/docs/specs/stream-offheap-budget-spec.md` resolves an open design question ("`fatal` flag for deployment failure") with *"non-fatal/transient is correct; permanent failure still surfaces after retries"*. The second half was **false on disk** for as long as this loop existed. It is true again with this change: a stream budget that genuinely can never fit never reaches ACTIVE on any node, so exhaustion settles it.

### DIFF
--- a/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentContext.java
+++ b/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentContext.java
@@ -256,6 +256,7 @@ public final class ClusterDeploymentContext {
                                                  ConcurrentHashMap.newKeySet(),
                                                  ConcurrentHashMap.newKeySet(),
                                                  ConcurrentHashMap.newKeySet(),
+                                                 ConcurrentHashMap.newKeySet(),
                                                  new ConcurrentHashMap<>(),
                                                  new ConcurrentHashMap<>(),
                                                  new ConcurrentHashMap<>(),
--- a/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentState.java
+++ b/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentState.java
@@ -1689,26 +1689,60 @@ public sealed interface ClusterDeploymentState extends FsmState<ClusterDeploymen
             SharedScheduler.schedule(this::reconcile, timeSpan(jitteredMs).millis());
         }
 
-        /// #922 — retry exhaustion is TERMINAL. It previously logged, cleared the
-        /// counter and routed `DeploymentFailed`, then returned, leaving the artifact absent from
-        /// `permanentlyFailed`. That was not a resting state. [#handleSliceFailure] issues an
-        /// unload on every failure; the node's removal of the `NodeArtifactKey` arrives at
-        /// [#handleSliceNodeRemoval], which — finding the artifact not permanently failed —
-        /// scheduled a reconcile 1 s later; [#reconcileBlueprint] is gated by nothing else, so it
-        /// redeployed the artifact and `retryCounters.merge` restarted at 1. An intermittent cause
-        /// that never settles therefore looped at roughly 1 Hz for the life of the cluster: no
-        /// terminal state, no rollback, and consensus round-trips forever.
+        /// #922 — retry exhaustion is TERMINAL, but only for a deployment that has produced
+        /// nothing. It previously logged, cleared the counter and routed `DeploymentFailed`, then
+        /// returned, leaving the artifact absent from `permanentlyFailed`. That was not a resting
+        /// state. [#handleSliceFailure] issues an unload on every failure; the node's removal of the
+        /// `NodeArtifactKey` arrives at [#handleSliceNodeRemoval], which — finding the artifact not
+        /// permanently failed — scheduled a reconcile 1 s later; [#reconcileBlueprint] is gated by
+        /// nothing else, so it redeployed the artifact and `retryCounters.merge` restarted at 1. An
+        /// intermittent cause that never settles therefore looped at roughly 1 Hz for the life of
+        /// the cluster: no terminal state, no rollback, and consensus round-trips forever.
         ///
-        /// Settling here bounds every intermittent cause at `MAX_RETRIES` + 1 reported failures and
-        /// makes the `ALL_OR_NOTHING` promise hold on this branch as it already did on the
+        /// Settling makes the `ALL_OR_NOTHING` promise hold on this branch as it already did on the
         /// deterministic one. The counter is cleared first so a later redeploy of the same artifact
         /// — which clears `permanentlyFailed` when the blueprint is applied — starts from a clean
         /// budget rather than inheriting an exhausted one.
+        ///
+        /// #924 review, BLOCKING: settling UNCONDITIONALLY here was itself a defect, and a worse
+        /// one than the livelock. The terminal is cluster-wide and permanent; the retry budget it
+        /// was hung on is per artifact AND node. So a slice already deployed and healthy on three
+        /// nodes, whose instance on ONE node suffered a transient longer than ~31 s, was settled
+        /// permanently failed for the WHOLE cluster — and because its blueprint had already left
+        /// `inFlightBlueprints`, [#rollbackBlueprintForArtifact] matched nothing and not even an
+        /// outcome record was written. Auto-heal, rebalancing and scale-up for that artifact were
+        /// then dead until an operator re-applied the blueprint. Before the #922 fix that case
+        /// self-healed, so the fix traded a livelock on a deployment that never succeeded for a
+        /// permanent, silent, cluster-wide failure of one that had.
+        ///
+        /// [#hasActiveInstanceElsewhere] is the discriminator, and the two populations are the ones
+        /// the ticket and the review name: a deployment ATTEMPT with nothing running is abandoned
+        /// (bounded — #922's case), while a RUNNING workload suffering a node-local transient keeps
+        /// being reconciled toward its desired instance count (unbounded — convergence, which is
+        /// what an orchestrator owes a workload the operator believes is up).
         private void handleRetryBudgetExhausted(SliceNodeKey sliceKey, String failureReason) {
             var artifact = sliceKey.artifact();
 
             retryCounters.remove(sliceKey.asString());
             if (permanentlyFailed.contains(artifact)) {
+                return;
+            }
+
+            if (hasActiveInstanceElsewhere(artifact)) {
+                log.warn("Max retries ({}) exceeded for {} on {}: {} — NOT marking permanently failed: "
+                         + "the artifact is ACTIVE on another node, so this is a node-local transient and "
+                         + "reconciliation keeps converging toward the desired instance count",
+                         MAX_RETRIES,
+                         artifact,
+                         sliceKey.nodeId(),
+                         failureReason);
+                ctx.router()
+                   .route(DeploymentFailed.deploymentFailed(artifact,
+                                                            sliceKey.nodeId(),
+                                                            SliceState.FAILED,
+                                                            failureReason,
+                                                            ctx.nowMs()));
+
                 return;
             }
 
@@ -1718,6 +1752,44 @@ public sealed interface ClusterDeploymentState extends FsmState<ClusterDeploymen
                       sliceKey.nodeId(),
                       failureReason);
             settleAsPermanentlyFailed(sliceKey, failureReason);
+        }
+
+        /// Whether ANY instance of this artifact is ACTIVE on a live node — the discriminator
+        /// between a deployment ATTEMPT that has produced nothing and a RUNNING workload suffering a
+        /// node-local transient (#924 review, BLOCKING).
+        ///
+        /// It exists to match the EVIDENCE to the BLAST RADIUS. The terminal reached by
+        /// [#settleAsPermanentlyFailed] is cluster-wide (`permanentlyFailed` is a `Set<Artifact>`)
+        /// and permanent; the retry budget that triggers it is per artifact AND node
+        /// (`retryCounters` is keyed on `sliceKey.asString()`). A cluster-wide permanent verdict
+        /// needs cluster-wide evidence, and "nothing of this artifact is running anywhere" is it.
+        ///
+        /// `inFlightBlueprints` is the cheaper test and the WRONG one. [ClusterDeploymentContext#newActive]
+        /// builds it empty and only a live `AppBlueprintPutReceived` populates it, so it does not
+        /// survive leader failover: gating on it would silently stop settling — reopening #922 — for
+        /// every deployment whose leader changed mid-flight. `sliceStates` is rebuilt from durable
+        /// KV entries by [#rebuildSliceStateFromKVStoreEntries] during [#rebuildStateFromKVStore] on
+        /// activation, so this predicate reads the same answer on a new leader as on the old one.
+        ///
+        /// Strictly ACTIVE, not [#isLiveState]: a sibling instance merely LOADING or LOADED is part
+        /// of the same unproven attempt and must not vote to keep it alive, or a deployment stuck
+        /// short of ACTIVE would never settle at all.
+        ///
+        /// "Elsewhere" is implicit rather than filtered: [#handleSliceFailure] removes the failing
+        /// key from `sliceStates` before either failure branch runs, so the instance that just
+        /// failed cannot count itself. That ordering is load-bearing — moving this call ahead of
+        /// that removal would make every exhaustion look survivable.
+        private boolean hasActiveInstanceElsewhere(Artifact artifact) {
+            var liveNodes = activeNodes();
+
+            return sliceStates.entrySet()
+                              .stream()
+                              .filter(entry -> entry.getKey()
+                                                    .artifact()
+                                                    .equals(artifact))
+                              .filter(entry -> liveNodes.contains(entry.getKey()
+                                                                       .nodeId()))
+                              .anyMatch(entry -> entry.getValue() == SliceState.ACTIVE);
         }
 
         /// The activation gate for a slice's schema migrations, scoped to the slice's OWN blueprint.
--- a/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentState.java
+++ b/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentState.java
@@ -74,6 +74,7 @@ import org.pragmatica.aether.slice.kvstore.AetherValue.ActivationDirectiveValue;
 import org.pragmatica.aether.slice.kvstore.AetherValue.AppBlueprintValue;
 import org.pragmatica.aether.slice.kvstore.AetherValue.ClusterConfigValue;
 import org.pragmatica.aether.slice.kvstore.AetherValue.CommunityValue;
+import org.pragmatica.aether.slice.kvstore.AetherValue.DeploymentOutcomeStatus;
 import org.pragmatica.aether.slice.kvstore.AetherValue.DeploymentOutcomeValue;
 import org.pragmatica.aether.slice.kvstore.AetherValue.GovernorAnnouncementValue;
 import org.pragmatica.aether.slice.kvstore.AetherValue.NodeArtifactValue;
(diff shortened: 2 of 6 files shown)


siy:
## Round-2 adversarial verification: 1 BLOCKING — do not merge

Head `8d9b8beb9`. Full report: `oss/internal/verify-924-round2-2026-09-07.md`. **Round 3 is in flight.**

**B1 — the round-1 BLOCKING is narrowed, not closed.** `hasActiveInstanceElsewhere` asks *"is an instance ACTIVE right now?"*, but `handleSliceFailure` removes the failing key from `sliceStates` **before** either failure branch runs. The predicate therefore returns false whenever the transient reached every instance, and it protects exactly one shape: a transient confined to a strict subset of instances.

Two shapes it cannot protect, both proven by probe with **both polarity controls passing**:

| Probe | `permanentlyFailed` | Expected |
|---|---|---|
| shared/downstream transient, previously-healthy deployment | `[slice-a]` | no settle — **FAIL** |
| **single-instance slice, transient on its only node** | `[slice-a]` | no settle — **FAIL** |

Both end with the only `DeploymentOutcomeValue` still reading **SUCCEEDED** and `reconcile()` emitting nothing. Silent, terminal, operator-only. **Both self-healed before this change.** The second needs no correlated failure at all and is the most common deployment shape there is.

**The reframing driving round 3:** the livelock and the self-heal are the *same mechanism*. Pre-#922, exhausting the budget cleared the counter and re-drove without settling — which is how a transient recovered, and equally how `ArtifactNotFound` spun forever. The discriminator must separate *was up and fell over* from *never came up*, so round 3 uses a predicate over durable evidence that the artifact **ever** reached ACTIVE rather than that it is ACTIVE now.

**S2 (new).** The predicate reads `activeNodes()` unguarded on an **irreversible** path. In the boot window the supplier yields `MEMBERSHIP_NOT_WIRED` — an empty set distinguished only by reference identity — so every node fails `contains` and the artifact settles permanently; membership churn does the same with no boot window. `coreMembershipResolved()` exists for exactly this and `StaleEntryCleaner` gates on it at four sites.

**Reported closed but open:** S1 (the ~31 s non-configurable budget; deferring the operator knob is reasonable, reporting it closed is not) and N3 (the fragment's "every bullet carries a tag" is 9 of 11). N4 rotted again — `:1901` now lands on javadoc; the fourth call site moved to `:1972` under this branch's own +118 lines.

**Every mutation claim the stream made reproduced exactly** (960/1, 960/2, correct tests). The diagnosis was accurate to the line and the livelock is real — this is a discriminator problem, not a wrong diagnosis.

siy:
## PARKED pending a structural audit — round 4 found a third BLOCKING defect

Head `7941ed668`. Full report: `oss/internal/verify-924-round4-2026-09-07.md`.

**BLOCKING.** `AbTestManager.targetPreservingOverrides` (`AbTestManager.java:274`) and `ControlLoopContext.applyScaling` → `putBlueprint` silently drop `SliceTargetValue.owningBlueprint`. Leg 3 — the durable leg — resolves SUCCEEDED through the owning blueprint, so **from the first autoscale or A/B test event it is permanently silent**. With leg 2 empty after failover and leg 1 dead on the first failure, round-2's blocking finding returns intact. Verified independently on disk. Filed separately as **#933**, because it is a production bug with consequences beyond this PR: a wrong `schemaRequired` default after failover, and an under-counted blast radius in schema-failure reporting.

**The thread question is settled and is NOT blocking.** Leg 3 runs on the consensus apply executor, synchronously inside `KVStore.process(Batch)`. `KVStore.get` is `Option.option(storage.get(key))` on a `ConcurrentHashMap` — a local read of already-applied replicated state. No I/O, no lock, no consensus wait. An outage makes it **silent, not slow**, which is the opposite failure mode from #929.

**What round 4 confirmed as claimed:** all three legs ARE pinned in isolation; the mutation matrix and 969/0 reproduce exactly.

### Why this is parked rather than sent back for a fourth round

Three rounds, three blocking findings, each defeated by a *different* piece of state:

| Round | Discriminator | Defeated by |
|---|---|---|
| 1 | settle unconditionally on exhaustion | condemns any transient |
| 2 | is an instance ACTIVE **now** | the key is removed *before* the branch runs |
| 3 | did it **ever** reach ACTIVE (3 legs) | all three legs can be false for a healthy workload |

The predicate is trying to read a fact **the system does not durably record**. Every candidate source fails differently: `sliceStates` is cleared before the check; `everActiveArtifacts` is in-memory and empty on failover; `DeploymentOutcomeValue` is blueprint-scoped (a partially deployed blueprint never writes it) *and* reachable only via an owner that #933 drops; `activeRoutings` is version-blind and was correctly rejected.

Owner's ruling, 2026-09-08: zoom out and examine the subsystem's design — if this is whack-a-mole, the problem is structural. A design audit is now running across the deployment subsystem, deliberately briefed so it can conclude these are independent bugs.

**Nothing here is rejected.** `7941ed668` stands, and the round-3 work — particularly the structural argument for leg 2 and the reasoned rejection of `activeRoutings` as version-blind — carries forward whatever shape the fix finally takes. #922 remains open and live.

**Do not merge. Do not open a fourth round until the audit reports.**

siy:
## Closing unmerged — the question has no home, and a fifth round would move the discriminator to a fifth piece of state

Owner ruling 2026-09-08, on the structural audit at `oss/internal/audit-deployment-subsystem-2026-09-08.md`. Recorded as `know:` commit `78960de46`.

**#922 stays OPEN.** The defect is real and live. This PR is closed; the ticket is not.

### Why closed rather than given another round

Four rounds, each competently argued, each defeated by a **different** piece of state its author had no reason to suspect:

| Round | Discriminator | Defeated by |
|---|---|---|
| 1 | settle unconditionally on exhaustion | condemns any transient |
| 2 | is an instance ACTIVE **now** | the failing key is removed *before* the branch runs |
| 3 | did it **ever** reach ACTIVE (3 legs) | all three legs can be false for a healthy workload |
| 4 | — | leg 3's evidence is reached through a pointer two unrelated subsystems erase |

Base rate for that strategy in this ticket is **0 for 4**. That is the signature of a question with no home, not of four careless attempts.

### What the audit established

The deployment subsystem durably records slice **facts** and holds slice **judgments** only in leader memory. Sorting `ClusterDeploymentState.Active`'s 13 components splits cleanly with no exceptions: all five facts (`sliceStates`, `transitionalStateTimestamps`, `workerNodes`, `sliceDependencies`, `activeRoutings`) are rebuilt from durable records; all six judgments (`permanentlyFailed`, `retryCounters`, `inFlightBlueprints`, `restoringBlueprints`, `consecutiveImbalancedTicks`, `reportedSchemaHolds`) are rebuilt **empty**, and not one has durable backing.

Every round here was an inference over side maps, because the question — *"was this artifact ever healthy?"* — is a judgment about the past, and judgments are not stored.

**`permanentlyFailed` is itself a judgment, so it clears on every leader change.** This PR was therefore building a terminal state on a mechanism that cannot hold one: even a perfect predicate produces a verdict that evaporates at the next failover. That also lowers #922's urgency — today's livelock and today's condemnation are both term-scoped, which no document in this PR's four-round review chain had noticed.

### The work is NOT discarded

- **Round 3's six-mutation matrix**, with each of three legs pinned in isolation and disjoint red sets — independently reproduced by round-4 verification — is the acceptance test for whatever replaces this.
- **Both round-4 probe suites** are preserved: `oss/internal/verify-924-round4-probe-Round4UnownedSliceProbeTest.java.txt` and `…-Round4PartialBlueprintProbeTest.java.txt`.
- **A cold-start handoff** written for a reader with zero context: `oss/internal/fix-924-round3b-parked-2026-09-07.md`. It carries the three-round table, the leg-2 structural argument, the threading answer, and four conclusions that must not be re-derived — notably that `activeRoutings` is version-blind and that `inFlightBlueprints` fails the opposite way.
- Head `7941ed668` remains on `origin`; the branch is not deleted.

### Conclusions that must not be re-litigated

- **Leg 2 is justified.** `recordSucceededOutcome` has one call site, gated on `pendingSlices().isEmpty()`, so a **partially deployed blueprint never writes a SUCCEEDED record at all** — leg 3 is blind by construction to a slice that did come up. Measured, with a passing positive control.
- **The blocking finding has two independent doors.** Owner erasure needs the autoscaler; a partially deployed blueprint across a leader failover needs neither. **Fixing #698 does not close this.**
- **Leg 3's KV read cannot block** — `Option.option(storage.get(key))` on a `ConcurrentHashMap`, a local read of already-applied replicated state. A consensus outage makes it *silent*, not slow.
- **The residues fail toward CONDEMNING, not re-driving.** The settle fires when every leg is silent, so a leg that cannot vote is a vote to condemn. The opposite was asserted in this PR's own fragment and was wrong.

### Sequence this sits in

**#698 → #926 → #805 item 1 → #930 (the decide half, folding in #923) → the storage half (the judgment record, which dissolves #922 and this PR).**

#930 fixes how permanence is *decided from a cause*; it does not touch how permanence is *recorded and re-read*. Every round here lived entirely in the second half. The two changes are complementary and neither subsumes the other.

Issue #914, opened 2026-09-06
Promise.await() ignores interruption: an interrupted waiter busy-spins forever (bounded await spins to its deadline)


siy:
## What

`Promise.await()` and `Promise.await(TimeSpan)` park in a bare loop and never consult the thread's interrupt status. `LockSupport.park()` **returns immediately, without clearing the interrupt flag, whenever the calling thread's interrupt status is set**. So once a waiting thread is interrupted:

- **`await()` (unbounded) never returns and spins at 100% CPU.** The loop is `while (result == null) { LockSupport.park(); }` — every `park()` returns instantly, the condition is unchanged, and the thread re-parks forever. Interruption cannot break it, and the thread burns a core while it waits.
- **`await(TimeSpan)` busy-spins until its deadline.** Same mechanism; the deadline still ends it, so the damage is bounded to the remaining timeout, but the wait consumes a core instead of sleeping and the interrupt has no effect.

## Where

- `core/src/main/java/org/pragmatica/lang/Promise.java:3282-3306` — unbounded `await()`, park loop at `:3302`.
- `core/src/main/java/org/pragmatica/lang/Promise.java:3313-3345` — `await(TimeSpan)`, `parkNanos` loop at `:3337`.
- No `InterruptedException`, `Thread.interrupted()` or `Thread.currentThread().interrupt()` anywhere in the file (grep: zero hits for `Interrupt`).

## Why it matters

Interruption is how a supervisor stops a stuck worker. Today it does the opposite: it converts a parked, idle thread into a spinning one that still cannot be stopped.

Observed, not theorised: during #727's settling run a partially-started `EmberCluster` hung on an unbounded `await()` for **586 s straight through an 8-minute JUnit backstop**, because the framework's interrupt could not end the wait (`oss/internal/rc4-AO-727-report-2026-09-06.md`; fixed test-side in #913 by bounding that specific await). The same shape reaches production: any executor shutdown, request cancellation, or supervisor timeout that interrupts a thread sitting in `await()` gets an unkillable, CPU-burning thread. It also feeds back into the flake picture — a spinning thread raises the very load under which #727 reproduces.

## Acceptance

[...]

Issue #911, opened 2026-09-06
Flaky on CI: SwimDeathPathCoConfirmationTest.singlePeerFaulty_notIsolation_backlogNotExpiredOnReconnect failed on an empty commit's push-CI


siy:
## What

`SwimDeathPathCoConfirmationTest.singlePeerFaulty_notIsolation_backlogNotExpiredOnReconnect` failed on the release branch's own push-CI (`build-and-test`) for an **empty** commit, which cannot change behaviour. Green on the three pushes that followed within eight minutes.

## Where

- Run 34055558733, branch `release-1.0.0-rc4`, head `89243fc8b` (a `know:` commit, no file diff), 2026-09-06 19:39 UTC — `build-and-test` FAILURE, `Tests run: 3, Failures: 1`.
- Runs 34055939777 / 34055958335 / 34055961005 on `c37a38cb1` / `a39cdb911` / `305780c96` (19:46-19:47 UTC) — SUCCESS.

## Why it matters

Decision 3 (branch protection requiring `build-and-test`, `forge-tests`, `fragment` to conclude) is gated on the known flakes being landed first. A flake that fires on an empty commit is a pure instrument failure and will block legitimate merges once protection is on. This ticket is the flake ledger for this test, on the #727 model; occurrences append here.

## Evidence

Observed by the CTO session 2026-09-06 from `gh run list --branch release-1.0.0-rc4` and `gh run view 34055558733 --log-failed`. Not reproduced locally. Not read-verified against the test source yet.

## Acceptance

Identify the timing assumption the test makes (the name suggests a backlog-expiry window versus a reconnect); either make the wait event-driven, give it honest load margin, or make the failure diagnostic (name what it timed out waiting for), and state which. Two green reruns are not acceptance.

siy:
## Occurrence, 2026-09-08 — local, and it adds a load-dependence signal this ledger did not have

Observed by `verify-934` during round-2 verification of PR #934 (tree `pragmatica-stream-h` @ `aba3f7330`). Posted by the CTO session, which holds this session's tracker-write authorization; the observing agent prepared the text and correctly declined to post it itself.

| Condition | Runs | Failures |
|---|---|---|
| The class **in isolation** | 6 | **0** |
| Full `swim` module | 4 | **1** |
| Full `swim` module, PR #934's round-2 test additions **reverted** | 4 | **0** |
| Full 4-module reactor | 1 | **1** (skipped the three downstream modules) |

[...]

PR #934, opened 2026-09-08
fix(aether): publish MembershipFsm transitions outside the per-member monitor (#929)

### LINKED ISSUE #929: Production deadlock: MembershipFsm publishes transitions under a per-member monitor, deadlocking two event loops at shutdown (surfaces as the MultiSourceCommunitySmokeTest teardown hang)
## What

`MultiSourceCommunitySmokeTest`'s `@AfterAll` teardown does not complete. Its bound fires with `TimeoutException: tearDown() timed out after 8 minutes`, the class reports a class-level ERROR at ~515 s, and the job fails — while the test bodies themselves finish in roughly 35 s.

**This is the residue of #915, not a regression of it.** #915's fix is present and working: the class has **zero** unbounded awaits, the wait is bounded, the bound fires, and `HangDiagnosticExtension` captures a 55-thread dump naming `tearDown()` as the failing phase. Before that fix this same failure ran silently into the job backstop with nothing naming what hung. So the diagnosis half succeeded and the underlying defect is now visible: **the cluster does not shut down.**

## Where

- `aether/forge/forge-tests/src/test/java/org/pragmatica/aether/forge/MultiSourceCommunitySmokeTest.java` — `@AfterAll tearDown()`.
- Diagnostic that made this legible: `HangDiagnosticExtension` (`:82`, `:113`), added under #749/#750.

## Evidence

Release-branch push-CI run **34112221517**, head `173cbdbc7` — which is an **empty `know:` commit**, so nothing in the change can have caused it. `build-and-test` passed; `forge-tests` failed.

```
=== HANG DIAGNOSTIC (#749/#750) === @AfterAll failed in MultiSourceCommunitySmokeTest:
TimeoutException: tearDown() timed out after 8 minutes
--- 55 threads ---
```

Timeline in that run: class starts 10:37:14, cluster forms 10:37:32 ("5-core cluster formed, committed cap=5"), first worker settles 10:37:33, diagnostic fires 10:45:49. The dump shows 55 live threads at teardown including `replica-set-controller` and `stream-partition-backfill` pairs parked on `AbstractQueuedSynchronizer`, and Netty event-loop threads still selecting.

## Why it matters

1. **The release branch's own CI is red on this**, on a commit that cannot have caused it. That is the fifth occurrence of a ~515 s class-level failure in about 24 hours, now on `main`-line CI rather than a PR.
2. **Branch protection (decision 3) requires `forge-tests` to have concluded successfully.** With this live, enabling protection blocks legitimate merges on a teardown that hangs, with nothing to revert.
3. **The 8-minute bound is far too long to be useful.** The test body completes in ~35 s. A bound roughly 14× the healthy duration converts a hang into a slow failure rather than a fast one, and each occurrence costs a full ~25-minute job.

## Acceptance

Teardown completes, or fails fast and names what did not stop. Concretely: identify which component's shutdown does not return — the thread dump points at replica-set/stream-partition workers and Netty loops still live — and either fix the shutdown path or bound each teardown step individually with a bound derived from its healthy duration rather than a single 8-minute umbrella.

Pin it with a test that stalls one component's shutdown and asserts the failure arrives quickly and names that component. Note the interaction with **#914**: `Promise.await()` ignores interruption and busy-spins, so a teardown built on it cannot be ended by a framework timeout — check whether that is the mechanism here before assuming it is a component bug.

## Related

#915 (bounded the awaits and added the diagnostic — this is what it revealed), #914 (uninterruptible await), #911 and #727 (the flake ledger this joins), decision 3 (branch protection precondition).

siy:
Fixes #929 — an AB/BA monitor deadlock between per-member `MembershipFsm$MemberTracking` locks that
wedges two production event loops of a live node. Implements owner ruling parts **1, 3, 4** (not
part 2, not #914).

## The defect

`MemberTracking.dispatch` was `synchronized` and published to `transitionSink` while holding the
dispatching member's monitor. The listener runs synchronously into
`AetherNode.onFsmTransition` → `propagateMemberCount` → `strictCoreObservedMemberCount`, which
streams the member map calling the `synchronized` `isStrictCoreMember()` on **every** member. So a
thread held one member's monitor and requested all the others', in `ConcurrentHashMap` iteration
order. One node's SWIM loop (`onSwimSuspect`) and QUIC loop (`onLivenessGone`) dispatching on two
different members each held one monitor and wanted the other's. That order is a hash order, so lock
ordering cannot repair it.

## What changed

**Part 1 — root fix.** `dispatch` splits into a `synchronized applyEvent` that mutates state and
*stages* the listener calls, and an unsynchronized `dispatch` that runs them under a new per-member
`transitionGuard` with the monitor released. All six sinks are staged, not just `transitionSink` —
the class documents the "cheap and non-blocking" contract that all of them were violating.

The guard preserves both properties the old `synchronized dispatch` gave: one member's transitions
stay totally ordered, and a listener still observes the member in exactly the state its record names.
Only the incidental blocking of *readers* is dropped — the deadlock's one edge. Lock order is always
guard → monitor; the four check-then-dispatch entry points were converted to take the guard first so
there is no inversion. The six ingress drivers now run under the same guard, which makes each driver
atomic against other transitions of that member — strictly stronger than before, since each step
previously took the monitor separately.

**Part 3 — bounded transport shutdown that stops lying.** `NettySwimTransport`'s bare `.sync()` calls
become bounded `await(5s)` (Netty's own timeout is enforced by `confirmShutdown()` *on* the event
loop, so a wedged loop cannot enforce it). The unconditional `LOG.info("SWIM transport stopped")` is
gone: success logs from `onSuccess`, and a shutdown that timed out logs a failure *instead of* it,
returning `SwimError.ShutdownTimeout`. The 2 s quiet period is dropped — SWIM is fire-and-forget UDP
with nothing to drain.

**Part 4 — bound where it can see the blocking code.** `EmberCluster.stop()` submits each
`node.stop()` so the synchronous prologue runs off the caller's thread and inside `NODE_TIMEOUT`.
Landed after part 1, per the ordering constraint.

## Evidence

- **Deterministic deadlock probe** driving the two production ingress points on two members of one
  FSM, with a barrier forcing the interleaving. **Mutation-tested:** restoring `synchronized (this)`
  around the fan-out turns it and the monitor pin RED; the fixed code runs 3/3 in 0.072 s.
- **Ordering/atomicity pin** goes RED when the guard is removed from both `dispatch` and
  `inTransition` (the naive "publish after release" fix).
- **Transport honesty pin** against a genuinely wedged event loop, with a positive control that the
  appender *can* see the success line. Reinserting the unconditional log turns it RED; removing the
  bound turns the failure-report assertion RED.
- **2346 module tests green** (aether-deployment, swim, ember, node), counted from surefire XML.
- **`jbct:check` green** on all four modules (invoked explicitly — it is not lifecycle-bound).
- **`MultiSourceCommunitySmokeTest` teardown: 6.25 s**, node stops concurrent, all 9 SWIM transports
  stopped within one millisecond, zero `HANG DIAGNOSTIC` / timeout hits. Previously this teardown was
  released at exactly T+480 s by the JUnit backstop; the healthy baseline was ~16.6 s.

One green run of an intermittent test is not proof the race is gone — the deadlock evidence is the
deterministic probe plus its mutation, not that run.

Full report, including the `transitionSink` consumer audit and two weak pins the mutations caught:
`oss/internal/fix-929-2026-09-07.md`.

**Do not merge — adversarial verification runs first.**

### DIFF
--- a/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/membership/fsm/MembershipFsm.java
+++ b/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/membership/fsm/MembershipFsm.java
@@ -4,6 +4,7 @@
 // See LICENSE in the repository root for full terms.
 package org.pragmatica.aether.deployment.membership.fsm;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -436,9 +437,13 @@ public final class MembershipFsm {
     }
 
     /// Register the Wave-1 transition-journal listener invoked once per ACTUAL per-member state
-    /// change, at the central dispatch chokepoint. The listener is called under the per-member
-    /// monitor — implementations must be cheap and non-blocking (the journal ring-buffer append
-    /// is). Diagnostic-only; a `null` argument resets it to the no-op.
+    /// change, at the central dispatch chokepoint. Since #929 the listener runs AFTER the
+    /// per-member monitor is released (still serialised per member by
+    /// [`MemberTracking#transitionGuard`], and still while that member's state is exactly what the
+    /// record names) — it may therefore read the member map without deadlocking, which is what
+    /// `AetherNode.propagateMemberCount` does on the Member-boundary edge. Cheap and non-blocking
+    /// is still the right shape: a slow listener stalls every further transition of THAT member.
+    /// Diagnostic-only; a `null` argument resets it to the no-op.
     @Contract
     public void onTransition(Consumer<MembershipTransitionRecord> listener) {
         this.onTransition = listener == null
@@ -456,9 +461,10 @@ public final class MembershipFsm {
 
     /// Register the Wave-4 membership-delta listener invoked once per counted-set lifecycle
     /// edge (JOINED / REMOVED, see [`MembershipDeltaEdge`]) at the central dispatch chokepoint.
-    /// The listener is called under the per-member monitor — implementations must be cheap and
-    /// non-blocking (the production [`MembershipDeltaProjector`] enqueues into its FIFO and
-    /// returns). A `null` argument resets it to the no-op.
+    /// Since #929 the listener runs AFTER the per-member monitor is released (still serialised per
+    /// member by [`MemberTracking#transitionGuard`]); cheap and non-blocking remains the right
+    /// shape (the production [`MembershipDeltaProjector`] enqueues into its FIFO and returns), but
+    /// it is no longer load-bearing against deadlock. A `null` argument resets it to the no-op.
     @Contract
     public void onMembershipDelta(Consumer<MembershipDeltaEdge> listener) {
         this.onMembershipDelta = listener == null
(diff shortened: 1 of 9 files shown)


siy:
## Merging — standing criteria met, and the check that has bitten before passes

**All three jobs green ON THE VERIFIED HEAD**, `aba3f7330`: `build-and-test` 19m40s, `forge-tests` 8m46s, `fragment` 7s. Stating the head explicitly because all three of PR #924's greens were on heads reviews subsequently rejected — mergeability is not checks-green, and a green on a superseded head is not a result.

**Round-2 adversarial verification: 0 BLOCKING, 0 SHOULD-FIX, 7 NOTE** (`oss/internal/verify-934-2026-09-08.md`). Every round-1 finding closed **against re-run evidence rather than the report's word**:

| Round-1 finding | Evidence it is closed |
|---|---|
| **BLOCKING** — mutation D wrongly retired as invalid; the guard acquisition in `dispatch` unpinned | Mutation D now **RED** at `aDirectDispatchIngress_…:284`; the reviewer's own probe had put **225 of 800** assertions stale under D while all 959 module tests stayed green |
| SHOULD-FIX 1 — negative result rested on an unretained log with blind controls | Console log retained (2124 lines); controls now **non-zero**: `INFO` 1930, both teardown markers 9, six failure patterns 0. Last round those controls were 0, which is what made the zeros worthless |
| SHOULD-FIX 2 — `await()` returned true for a *completed* future, not a *succeeded* one, so a failed close logged the success line | `isSuccess()` consulted; new `SwimError.ShutdownFailed` carries the cause; mutation F **RED** with its control green |
| SHOULD-FIX 3 — `doStop` lost `Result.lift` | Lifting restored, inner `Result` flattened |
| SHOULD-FIX 4 — `stopThen` let a cleanup failure gate an already-resolved address | Logged and `recover`ed before `flatMap` |
| SHOULD-FIX 5 — gate claim unearned; the PR *added* a format violation | Gate forced with non-zero counts **88 / 150 / 6 / 18**; `NettySwimTransport.java` gone from the format list; the surviving `SwimProtocol.java` issue proven pre-existing by `cmp` against `git show 4607e86d0:…` |
| NOTE 1 — leaf invariant unstated and unpinnable | **Closed, and the reviewer records that it was wrong**: `Fsm` invokes the observer under the monitor, so it *is* observable. Pinned; mutation G **RED** |

**Tests: 2350** (961 / 178 / 6 / 1205), counted from surefire XML with stale reports deleted first — and reproduced by the reviewer on a box carrying **four concurrent external Maven builds**, a harsher condition than the reporting run.

### What this fixes

An AB/BA monitor deadlock between per-member `MembershipFsm$MemberTracking` locks. `dispatch` published to `transitionSink` while holding member X's monitor; the listener walked every member calling the `synchronized` `isStrictCoreMember()`. Two event loops dispatching on different members wedged in `ConcurrentHashMap` iteration order — a hash order, so no lock-ordering discipline could repair it. The class's own javadoc stated the contract this broke.

The fix splits one lock into two: `applyEvent` mutates under the monitor and **stages** the fan-out, which `dispatch` then runs under a separate guard with the monitor released. It preserves both properties the old code gave — per-member transition ordering, and a listener seeing the state its record names — and drops only the incidental blocking of readers, which was the deadlock's sole edge.

### Carried forward, deliberately

Seven NOTEs, none blocking. **NOTE 7 is filed as #942** — `EmberCluster.stop()` skips `clearClusterState` when a node stop fails, a pre-existing shape that part 3's bounded transport makes materially more reachable. Not exercised by anything today, which is why it is a reading rather than a measurement.

**The `build-and-test` failure earlier on this branch was `SwimDeathPathCoConfirmationTest`, and it is #911's, not this PR's** — that ticket's own occurrence is on an **empty `know:` commit with no file diff**, so the test flakes with zero code change. A new occurrence with a load-dependence signal is appended there.

**#929 closes with this.**

Issue #515, opened 2026-07-24
jbct init: no cloud deployment story — deploy-prod.sh calls nonexistent `aether artifact push --env`, no bootstrap config scaffold; forge /data/aether absolute default WARNs on laptops


siy:
Getting-started dry-run (2026-07-24), scaffold cloud-story gaps (sibling of #511/#513, same template-drift family):
1. `deploy-prod.sh` invokes `aether artifact push --env prod` — **no such command/flag exists** in the current CLI (repo-wide grep: zero matches for `--env` in aether/cli). The script also references the stale `target/blueprint.toml` file remnant (#513).
2. `jbct init` scaffolds NOTHING for cloud provisioning: `aether.toml` is only a commented database section; no bootstrap TOML template, no pointer to `aether cluster bootstrap`. Combined with the missing docs (see the bootstrap-docs issue) the cloud path is a dead end from the scaffold.
3. Minor but trust-eroding: forge/node defaults write `/data/aether/*` (absolute system path) → on a laptop the first forge run prints a wall of read-only-filesystem WARNs and silently disables the WAL ("non-crash-durable"). A local-dev simulator should default under the project dir or $AETHER_HOME.

Fix: regenerate deploy-prod.sh against the real CLI surface; scaffold a minimal bootstrap TOML (or `jbct add-cloud` command) once the docs issue lands its reference example; forge data-dir default relative/homed. Covered by the same structural init-output smoke gate as #511/#513.

siy:
Fixed in 92f9dbe30 (release-1.0.0-rc3): deploy-test.sh/deploy-prod.sh regenerated against the real CLI (aether artifacts push + aether blueprints deploy --wait); forge homes node data under $AETHER_HOME/forge-data - stream WAL crash-durable in local dev, one loud line instead of the WARN wall + silent WAL-off; production node /data/aether default untouched. Point 2 scope note: aether.toml gained a commented pointer to the new bootstrap-config reference (docs #514) rather than a scaffolded bootstrap TOML template - if a full scaffolded template (or jbct add-cloud) is wanted, say so and it becomes a follow-up.

PR #944, opened 2026-09-08
chore: gate CLI documentation against the live picocli CommandSpec trees

siy:
Makes CLI documentation drift fail the build instead of rotting silently.

## What

A new test-only module `cli-docs-gate`, last in the root reactor. It builds the **real
`picocli.CommandSpec` trees** for both shipped CLIs (`aether`, `jbct`), extracts every CLI invocation
written in this repository's Markdown, and diffs them both ways.

Follows the `aether/dead-surface-gate` pattern: owns no production code, declares test-scoped
dependencies on both CLI modules so no reactor order can run it against a CLI that has not been built.

Not a text scrape, deliberately — a scrape cannot see `@Mixin` contributions (`OutputOptions` gives
the root `--format`/`--output`), the options `mixinStandardHelpOptions` synthesises, or picocli's own
`AutoComplete.GenerateCompletion`, whose command name `generate-completion` appears nowhere in this
repository's source. That last one is a real shipped undocumented subcommand this gate found.

## The counts are part of the result

Every run prints them, and the gate **fails rather than passes** if the corpus collapses (floors:
100 docs / 300 invocations / 100 commands):

```
documents scanned ....... 230 Markdown file(s)
fenced shell blocks ..... 1060 read, 1526 skipped by language
inline code spans ....... 29334
invocations extracted ... 919
commands enumerated ..... 219 (aether 200, jbct 19)
```

The corpus is **sweep-everything-minus-an-explicit-exclusion-list**, not an include-list of globs: a
glob that stops matching after a directory rename would turn this gate green forever. Exclusions
(changelogs, `.internal`, `archive`, `specs/future`, build output) each carry a stated reason — a
changelog's job is to name commands as they *were*.

Blind spots are printed on every run rather than left unknown: one command declares both subcommands
and positionals (`aether deploy`); option values, positional shapes and clustered short options are
not validated.

## Drift

**Fixed here (21 edits, each verified against the CommandSpec):** `aether blueprint {deploy,apply,list,status,delete}`
→ `blueprints` (14), `aether artifact push` → `artifacts push` (6), `aether-cli deploy` → `aether deploy` (1).

Enumerating before substituting mattered: `aether blueprint set` and `aether artifact upload` look
like the same typo but have **no plural counterpart either**, so they were left alone rather than
"fixed" into a second wrong name.

**Waived, not absorbed — 91 findings across 35 files** (34 absent commands, 57 absent options), plus
**11 undocumented commands** baselined. These are feature claims, not typos: `aether scale
--min/--max/--target-cpu` describes autoscaling the CLI does not have, and `jbct init --slice` is the
*inverse* of the shipped `--no-slice`. Rewriting them is a product call, not a rename.

Both lists are **self-cleaning**: a finding not on the list fails; a list entry matching nothing fails;
a baseline entry that is no longer a command fails. They can only shrink. Keyed by path plus the drift,
never by line number.

## Proof it can fail

All 8 probes run against this head, tree clean before and after. Each demands a non-zero exit **and**
its own specific message.

| Probe | Mutation | Exit |
|---|---|---|
| P0 | none (clean tree) | **0** |
| P1 | fake command added to a doc | **1** |
| P2 | real option renamed **in situ** in a doc | **1** |
| P3 | corpus collapsed (`.md` → `.mdx`) | **1** |
| P4 | a doc fixed, its waiver line left behind | **1** |
| P5 | baseline entry removed | **1** |
| P6 | wrong binary name (`aether-cli`) | **1** |
| P7 | baseline entry that is not a command | **1** |

P2 is in situ rather than appended, so it proves the gate reads real content and not only text the
probe just wrote. P3 covers the "green having examined nothing" class this repo has been burned by.

## Verification

- `mvn -pl cli-docs-gate test` — 5/5 green at this head.
- `mvn install -DskipTests -pl '!examples'` — CI's own module scope; **143-module reactor, BUILD
  SUCCESS**, `cli-docs-gate` at position 135. CI runs `mvn install -B -pl '!examples'`, so the gate is
  inside the scope that gates merges.

Two defects in the gate itself were found by measuring rather than assuming: `Files.walk` descended
into sibling modules' `target/` before filtering, racing CI's `-T 1C` module-parallel reactor (now
pruned at `preVisitDirectory`; all four corpus counts identical before and after), and dotted tokens
like `aether cluster ...` were reported as missing commands (fixed after confirming no command in
either CLI has a dot in its name).


siy:
## Update — both BLOCKINGs from adversarial verification are fixed

Head `db86381c590eb5b1d3ce2cf1c2a9b380d5343001`. Verification report: `oss/internal/verify-944-2026-09-08.md`.

### BLOCKING 1 — the ratchet stopped appending but not swapping

The size-only ratchet asserted the key set's **cardinality, never its content**. Delete one waiver
line for a document you fixed, add one for drift you introduced: count unchanged, nothing stale,
nothing unwaived, **exit 0**. New drift silenced with no constant edited, and the diff reads like a
virtuous commit.

Both lists are now pinned twice, and the halves do different jobs:

- **COUNT** blocks growth, and is the half a human reads — `91` → `85` in a diff says "drift was
  fixed" at a glance, which a digest never can;
- **DIGEST** (SHA-256 over the sorted entries) blocks rewriting at constant size.

New probe **P10** performs exactly the verifier's swap and the build goes red with `NOT THE SAME
ENTRIES`. Matrix is now **11/11 at this head**, tree clean before and after each:

| | mutation | exit |
|---|---|---|
| P0 | clean | **0** |
| P1 | fake command | 1 |
| P2 | real option renamed in situ | 1 |
| P3 | corpus collapse (`.md`→`.mdx`) | 1 |
| P4 | doc fixed, waiver left behind | 1 |
| P5 | baseline entry removed | 1 |
| P6 | wrong binary | 1 |
| P7 | baseline entry that is not a command | 1 |
| P8 | new drift + **appended** waiver line | 1 |
| P9 | shrink without lowering the constant | 1 |
| **P10** | **swap at constant size** | **1** |

### BLOCKING 2 — ten fixable typos, and the method that missed them

Ten of the 91 waived entries were the identical singular/plural error I fixed 20 times, each carrying
an option between the binary and the subcommand (`aether -c "$AETHER_URL" artifact push …`). **The
first fix pass was driven by a text search for the literal string — in the PR whose argument is that
text searches cannot reach what a parse can.**

Pass two was driven by the gate's own extraction. **How the new method was shown to reach what the old
could not**, rather than assumed: both were run over the same corpus and disagree by exactly the ten —
the text search matches none of them. Each replacement's plural target was checked against the
enumerated `CommandSpec` paths first, which is what stopped `aether blueprint set` being "fixed" into
`aether blueprints set`, a command that does not exist either.

Waiver **91 → 85** (29 absent commands, 56 absent options, 34 files). After both passes the only
`blueprint`/`artifact` command tokens among the 919 extracted invocations are two deliberate ones,
each waived with a reason.

### Also fixed

- **NOTE 8.2, my own regression** — one first-pass edit rewrote *historical* prose: `getting-started.md`
  describes a past bug as "the broken `aether artifact push --env prod` call (#515) is gone", and I
  renamed it to `artifacts push`, naming a call that was never broken. Reverted; the restored line is
  waived. It shows the changelog exclusion is by directory, so historical prose inside a living
  document is unprotected.
- The changelog's `[verified:]` tags no longer claim more than their mutations prove.
- Reactor numbers corrected: CI's scope `-pl '!examples'` is **141 modules with the gate LAST at 141**,
  not the "143 / position 135" I reported (143 is Maven's pre-exclusion inspect count).

### Not fixed here, deliberately

- **SHOULD-FIX 7.2** — `resolve()` discards accumulated option findings on hitting an unknown
  subcommand, so options on the 29 command-drift lines go unvalidated. Real, same family as the
  under-matching bug I self-reported. It did **not** materialise on these ten: those lines carried only
  `-c` and `--api-key`, both valid, and the one key added to the waiver came from the 8.2 revert.
- **`jbct.skip`** — pre-existing, not a regression. `jbct.includeTests=false` takes both this module
  and `dead-surface-gate` to zero files examined, so setting `jbct.skip=false` here would only move it
  from a goal that does not run to a goal that runs on nothing.
- **`aether stream create`** in `in-memory-streams-spec.md` — that spec is written against the singular
  `stream` tree throughout; renaming one line to the plural leaves it internally inconsistent while
  four sibling commands remain absent under either tree. Belongs to #951.

Issue #760, opened 2026-08-30
fix(deployment): unresolved schema_required silently holds every slice in LOADED — diagnosis is DEBUG-only


siy:
## Summary

When a blueprint's `schema_required` cannot be resolved, it defaults to **required**. Any blocking schema record then holds **every slice in that blueprint** in `LOADED` indefinitely — no routes publish, the whole application surface 404s — and the only explanation emitted anywhere is at `DEBUG`.

The fail-closed default is deliberate and defensible. **This ticket is about the silence, not the gate.**

## The mechanism

`ClusterDeploymentState.resolveSchemaRequired` (`:1165`) reads the persisted blueprint's `resourcesConfig`, parses it, and reads `deploymentConfig().schemaRequired()`. When any link is missing it ends `.or(true)`:

```java
.onEmpty(() -> log.debug("schemaRequired unresolved for {}, defaulting to true (missing blueprint entry, resourcesConfig, or unparsable/incomplete resources.toml)", blueprintId))
.or(true);
```

`areSchemasReady` (`:1417`) then returns false while any record owned by that blueprint sits in `BLOCKING_SCHEMA_STATUSES` — `PENDING`, `MIGRATING`, `FAILED` (`:174`). `tryActivateIfDependenciesReady` (`:1448`) returns early with another `log.debug`, so `ACTIVATE` is never issued and `publishRoutesIfPresent` is never reached.

Note this is a genuine **cluster-wide, all-or-nothing** gate — unlike ordinary activation, which is per-slice and gated only on a slice's own dependencies. One unresolved config key takes down the entire application surface.

## Evidence

Local 5-node forge cluster deploying the `ticketing` example (24 slices). With DEBUG enabled on the deployment FSMs:

```
110 x  "waiting for schema migrations"
  0 x  "activating immediately"
  0 x  "Issuing ACTIVATE command"
```

```
schemaRequired unresolved for org.pragmatica.example:ticketing:1.0.0-SNAPSHOT,
defaulting to true (missing blueprint entry, resourcesConfig, or unparsable/incomplete resources.toml)
```

`ticketing`'s `resources.toml` has no `[deployment]` section at all, so resolution yields nothing and the default engages.

Operator-visible state at that moment — **this is the whole problem**:

```
management API : 68 instances LOADED, 7 LOADING, 0 ACTIVE
deploy API     : HTTP 200 {"status":"deployed","slices":24}     (#759)
GET /                              -> 404
[...]

Issue #570, opened 2026-08-03
CTM setDesiredSize: unguarded read-modify-write on ClusterConfigKey loses concurrent scale requests


siy:
## Problem

`ClusterTopologyManagerRecord.setDesiredSize` performs an unguarded read-modify-write on the replicated `ClusterConfigKey.CURRENT` key. Two concurrent scale requests silently lose one of the updates, leaving the cluster at a size nobody asked for.

## Mechanism

`ClusterTopologyManagerRecord.java:227-247`:

```java
private Promise<Unit> writeDesiredCoreCount(int size) {
    var existing = clusterConfigReader.get();                      // READ
    ...
    var updated = withCoreCount(existing.unwrap(), size);          // MODIFY (bumps configVersion)
    var command = new KVCommand.Put<>(ClusterConfigKey.CURRENT, updated);   // WRITE — plain Put
    return commandApplier.apply(List.of(command)) ...
                         .onSuccess(_ -> log.info("... (configVersion={})", updated.configVersion()));
}
```

`configVersion` is incremented into `updated` and then **only logged** — it is never used as a precondition. The write is a plain `Put`, not a compare-and-set. Two callers that both read version *N* both write *N+1*; Rabia serialises them, last writer wins, and the first request is discarded with no error to either caller.

## Why the usual gate doesn't cover this

CTM's other entry points check activation:

- `onClusterConfigChanged` (`:291`), `onMembershipDecision` (`:301`), and siblings at `:316`, `:1025` all begin `if (!active.get()) { return; }`.

**`setDesiredSize` (`:217-225`) does not.** It validates only `size >= MINIMUM_CLUSTER_SIZE`, then calls `writeDesiredCoreCount` directly.

That matters because CDM and ControlLoop are leader-gated *structurally* — CDM only handles write-triggering events in its `Active` FSM state, ControlLoop's write path only exists while its timer is scheduled, and both are driven by `LeaderChange`. CTM's reconcile paths follow the same discipline via `active.get()`. The scale path is the one write that enters **behind** that gate.

## Reachability — not theoretical

`setDesiredSize` is called from the operator HTTP path, `ClusterConfigApplier.applyScaleUp` / `applyScaleDown` (`ClusterConfigApplier.java:96,103`). "Leader" appears in that route only for diagnostics, never as an admission check. `ManagementServer` runs on **every** node, so any node can accept a scale request.

[...]

PR #956, opened 2026-09-08
fix(aether): one owner source for the schema gate, and fence the best-effort outcome merge (#805)

### LINKED ISSUE #805: Schema activation gate pre-filters the in-memory blueprints map before the shared predicate; bestEffortFailureCommand merges by read-then-Put

siy:
Closes #805 — both residuals from the round-3 review of the #760/#724/#759 branch.

## 1. Gate pre-filtered a stale in-memory map

`blocksSliceActivation` is shared by the activation gate and `SchemaRoutes.heldSlices`, but the two fed it from different sources: the gate took the slice owner from the node-local `Active.blueprints` mirror *and* pre-filtered on that mirror's `schemaRequired` flag; the route read the committed `SliceTargetValue` and pre-filtered on nothing. The mirror lags the KV notifications that rebuild it, so a stale entry made the gate hold a slice the route reported as not held.

Both call sites now go through one new static, `ClusterDeploymentState.resolveSliceOwner(KVStore, Artifact)`, reading the committed record; the mirror pre-filter is gone. It removes no check — `blocksSliceActivation` already resolves `schemaRequired` per candidate record from the same store.

## 2. `bestEffortFailureCommand` read-then-Put

The read happens at command-build time, the Put applies after consensus, so two BEST_EFFORT failures in flight before either applies read the same base and each Put the other's slice id away. Not a narrow window: `RabiaEngine.pendingBatches` is a `ConcurrentSkipListMap` keyed by a SHA-256 content hash (`"batch-" + sha256Hex(...)`, lexicographic `Id.compareTo`) and all three proposal sites take `firstEntry()`, so the survivor is chosen by content hash, not submission order — a coin flip on the happy path, no failed round required.

`DeploymentOutcomeValue` now carries `outcomeVersion` and implements `VersionFenced` (RFC-0018, #570), so the applier rejects a Put that is not the committed value's immediate successor. Rejection alone still drops the id, so the merge path confirms after its own apply resolves and retries against current committed state, bounded at 5 attempts with an ERROR naming the slice on exhaustion. The re-read is sound because `RabiaEngine.commitChanges` calls `stateMachine.process` before `promise.succeed`.

**Guarantee, per operation:** a BEST_EFFORT deterministic slice failure adds its artifact id to the committed `failingSlices`, or logs at ERROR that it did not. Not atomicity — the applier's successor-version fence plus a bounded read-after-apply retry. Convergence is bounded, not guaranteed.

## Verification

- 3946 tests green across `slice`, `aether-deployment`, `node`, `aether-control`, `aether-invoke`, `cli` (counted from surefire XML); 0 failures, 0 errors.
- Gate: `Running JBCT check on 80 / 88 / 150 Java file(s)`, `JBCT check passed.` on all three changed modules.
- **Mutation probes**, each a revert of one production hunk against a committed base:
  - Item 1 hunk reverted → all 4 `StaleMirrorDivergence` tests red, nothing else moved.
  - `VersionFenced` removed → 6 of 9 red.
  - Confirm-and-retry removed, fence intact → exactly the 4 `ConcurrentFailures` red, `ApplierFence` green.
  - `failedOutcomeCommand` reverted to a blind version stamp → the ALL_OR_NOTHING acceptance test red.

An earlier version of the concurrency test passed with the fence removed — the retry happened to cover for it. The test now drains each confirm before releasing the next batch, which is both the adversarial order and the one production reaches.

## No behaviour change for ALL_OR_NOTHING

Fencing the record makes every writer fenceable, so all four production write sites derive their version via `Active.nextOutcomeVersion`. Pinned by `AllOrNothingUnaffected#theTerminalOutcome_isAccepted_overAnAlreadyCommittedRecord`, which asserts on committed state after a real apply — the pre-existing `ClusterDeploymentStateTransactionalTest` coverage cannot speak to this, as it inspects proposed commands from a recording node that never applies anything.

## Wire format

`deployment-outcome` TOML goes 4 fields → 5. Consistent with RFC-0018 O1's rc-line posture: rc releases do not support mixed-version co-application.

### DIFF
--- a/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/BlueprintService.java
+++ b/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/BlueprintService.java
@@ -77,8 +77,10 @@ public interface BlueprintService {
     Option<ExpandedBlueprint> get(BlueprintId id);
     /// Durable terminal outcome of `id`'s last deployment attempt (#759 review, BLOCKING 3). Bounded
     /// to exactly one record per blueprint id: the FSM writes via `KVCommand.Put` at
-    /// `AetherKey.DeploymentOutcomeKey.deploymentOutcomeKey(id)`, and a Put at the same key overwrites
-    /// the prior value, so the store holds only the latest outcome — cardinality is the number of
+    /// `AetherKey.DeploymentOutcomeKey.deploymentOutcomeKey(id)`, and a Put at the same key replaces
+    /// the prior value — subject, since #805 item 2, to the `VersionFenced` successor check, which
+    /// rejects a Put built on a stale read rather than letting it overwrite. The store therefore holds
+    /// only the latest outcome — cardinality is the number of
     /// distinct blueprint ids ever deployed, not the number of attempts. Survives
     /// `unloadBlueprintSlices`'s ALL_OR_NOTHING rollback, which removes only `AppBlueprintKey`, never
     /// this key — the intended read path for the node's blueprint-status route after a rollback
--- a/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentState.java
+++ b/aether/aether-deployment/src/main/java/org/pragmatica/aether/deployment/cluster/fsm/ClusterDeploymentState.java
@@ -188,6 +188,31 @@ public sealed interface ClusterDeploymentState extends FsmState<ClusterDeploymen
         return resolveDeclaredSchemaRequired(kvStore, blueprintId).or(true);
     }
 
+    /// #805 item 1: the SINGLE slice-owner resolution path for both callers of
+    /// [#blocksSliceActivation] — the activation gate (`Active.blockingSchemaRecords`) and
+    /// `SchemaRoutes.heldSlices` — reading the committed `SliceTargetKey`/`SliceTargetValue` record
+    /// that both already treat as the ownership authority.
+    ///
+    /// #760's first round made the two share this PREDICATE; they still fed it from two different
+    /// sources. The gate resolved the owner from [Active#blueprints], a node-local mirror rebuilt
+    /// from KV notifications and therefore lagging them, and additionally pre-filtered on that
+    /// mirror's `schemaRequired` flag; the route read the KV record directly and pre-filtered on
+    /// nothing. A stale mirror entry naming an owner the committed record no longer names made the
+    /// gate hold a slice the route simultaneously reported as NOT held — the same divergence #760
+    /// closed, running the other way. Sharing a predicate is not sharing a decision while its INPUTS
+    /// disagree.
+    ///
+    /// The dropped `schemaRequired` pre-filter is not a lost check: [#blocksSliceActivation] already
+    /// resolves `schemaRequired` per candidate record from this same [KVStore] via
+    /// [#resolveSchemaRequired(KVStore, BlueprintId)], so the mirror's copy was strictly a second,
+    /// divergent answer to a question the shared predicate was already asking.
+    static Option<BlueprintId> resolveSliceOwner(KVStore<AetherKey, AetherValue> kvStore, Artifact artifact) {
+        return kvStore.get(SliceTargetKey.sliceTargetKey(artifact.base()))
+                      .filter(SliceTargetValue.class::isInstance)
+                      .map(SliceTargetValue.class::cast)
+                      .flatMap(SliceTargetValue::owningBlueprint);
+    }
+
     record Dormant(ClusterDeploymentContext ctx) implements ClusterDeploymentState {
         @Contract
         @Override
@@ -1710,8 +1735,8 @@ public sealed interface ClusterDeploymentState extends FsmState<ClusterDeploymen
         /// permanent failure. The hold now clears only via `/api/schema/{ds}/retry`
         /// (FAILED -> PENDING -> COMPLETED) or a redeploy that republishes the record.
         ///
-        /// A slice whose owning blueprint cannot be resolved — no `Blueprint` entry, or an entry
-        /// carrying no owner — is reported READY. No record can be attributed to it, so blocking
+        /// A slice whose owning blueprint cannot be resolved — no committed `SliceTargetValue`
+        /// record, or one carrying no owner — is reported READY. No record can be attributed to it, so blocking
         /// would be an unclearable hold: nothing that ever completes could match it, and the slice
         /// would sit in LOADED forever. Records only ever exist because some blueprint declared
         /// migrations, and that blueprint's own slices do carry its owner, so the safety property is
@@ -1723,12 +1748,15 @@ public sealed interface ClusterDeploymentState extends FsmState<ClusterDeploymen
         /// Named records rather than a boolean so the hold can be reported with detail (#760) — the
         /// prior `noBlockingSchemaRecords` collapsed the same scan into a single flag, which is all
         /// [#areSchemasReady(SliceNodeKey)] needs but nothing an operator-facing log could name.
+        /// #805 item 1: owner resolution goes through the shared
+        /// [ClusterDeploymentState#resolveSliceOwner(KVStore, Artifact)] — the committed
+        /// `SliceTargetValue` record — instead of the node-local [#blueprints] mirror, and the
+        /// mirror's `schemaRequired` pre-filter is gone. Both were inputs the route did not share,
+        /// so the gate and `SchemaRoutes.heldSlices` could disagree about the same slice even while
+        /// calling one predicate. See that method for why dropping the pre-filter removes no check.
         private List<SchemaVersionValue> blockingSchemaRecords(SliceNodeKey sliceKey) {
-            return Option.option(blueprints.get(sliceKey.artifact()))
-                         .filter(Blueprint::schemaRequired)
-                         .flatMap(Blueprint::owner)
-                         .map(this::collectBlockingSchemaRecords)
-                         .or(List.of());
+            return resolveSliceOwner(ctx.kvStore(), sliceKey.artifact()).map(this::collectBlockingSchemaRecords)
+                                                                       .or(List.of());
         }
 
         private List<SchemaVersionValue> collectBlockingSchemaRecords(BlueprintId owner) {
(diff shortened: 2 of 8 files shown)


siy:
## Update: relayed #698 owner-path finding — determined CLOSED at this head, with a new regression sensor

`fix-930-922` relayed that `recordBestEffortFailureOutcome` reaches `DeploymentOutcomeKey` via `Blueprint::owner`, which traces to `SliceTargetValue.owningBlueprint` — the field #698 found being erased. An erased owner means **no record is written at all**, so the CAS fix here would have been correct and irrelevant for every autoscaled or A/B-tested slice.

**Determination: closed at this head — enumerated, not assumed.**

`Active.blueprints` has four write sites; two pass `Option.some(expanded.id())` and can never be empty, two read `SliceTargetValue.owningBlueprint()`. So the exposure reduces to that record's writers. All five audited:

| Writer | Owner handling |
|---|---|
| `ControlLoopContext.applyScaling` | fixed by #940 |
| `AbTestManager.targetPreservingOverrides` | fixed by #940 |
| `DeploymentManagerImpl.addSliceTargetCommand` | `existing.map(withX).or(owner-less fallback)` |
| `RollbackManager.updateSliceTargetForRollback` | `existing.map(withVersion).or(owner-less fallback)` |
| `SliceRoutes.applyDeployCommand` | `existing.map(applyScaleToExisting).or(owner-less fallback)` |

The last three were **not** named by #698, so they were the real risk. They are safe: every `withX` helper copies `owningBlueprint` through, and the owner-less fallback fires only when no record exists. This branch's base *is* #940 (`a75e6af42`), and both fixed producers were read at this head rather than trusted from the commit message.

**Mutation-proved.** New `BestEffortOutcomeMergeTest$OwnerResolutionFromTheSliceTargetRecord` drives the real `SliceTargetPutReceived` notification instead of seeding the mirror. Erasing the owner as it enters the mirror turns the positive case red with exactly the relayed symptom (`no deployment-outcome record was committed`) while the owner-less negative case stays green — a discriminating pair.

**Not closed, and deliberately not fixed here:** the owner lookup still reads the in-memory mirror rather than the committed record — mirror *staleness/absence* is a separate path to a dropped record, distinct from #698's *erasure*. A naive source switch would regress: in `handleAppBlueprintChange` the mirror gets its entry in the same pass that only queues the `SliceTargetKey` Put, so the mirror legitimately leads the store in that window. Correct shape is mirror-first with a committed-record fallback. Awaiting a scope ruling.

Re-verified: 3948 tests green, gate `Running JBCT check on 80 / 88 / 150 Java file(s)` + `JBCT check passed.` on all three.

siy:
## Scope ruling taken: mirror lookup fixed in this PR

**Final SHA: `3accd72e57fa87156d263f8d8ae12a70f0cfdcad`** — pinned, no further pushes.

`Active.resolveOutcomeOwner` now reads the in-memory mirror first and falls back to the committed `SliceTargetValue`. #698/#940 closed *erasure*; this closes *mirror unavailability* — an artifact whose entry `removeNonTargetVersions` dropped, or the window after failover before `rebuildSliceStateFromKVStoreEntries` runs. Either way the old result was no record at all, not a degraded one.

**Mirror-first is load-bearing, and the obvious fix is the wrong one.** Two orthogonal probes:

| Probe | Result |
|---|---|
| mirror-only (fallback removed) | `whenOnlyTheCommittedRecordNamesTheOwner` RED; mirror test GREEN |
| committed-record-only (mirror ignored) | `whenOnlyTheMirrorNamesTheOwner` RED — **plus 6 more**, including every item-2 concurrency pin |

The second is the one worth keeping. Resolving from the committed store alone is exactly the change a reader would call a simplification, and it breaks 7 of 15 tests, because `handleAppBlueprintChange` populates the mirror in the same pass that only *queues* the `SliceTargetKey` Put — the mirror legitimately leads the store there. `OwnerResolutionFallback#theOutcomeRecord_isWritten_whenOnlyTheMirrorNamesTheOwner` pins that window (driving the real `AppBlueprintPutReceived` notification, with slice-target writes dropped to model it), so the simplification cannot land silently.

Consulting both strictly widens what is recorded: worst case is attributing a failure to an owner about to change, never losing the record. For a ticket about lost records that is the right asymmetry.

**Verification at the pinned SHA:** 3950 tests green (surefire XML count) across the six modules, 0 failures / 0 errors; gate `Running JBCT check on 80 / 88 / 150 Java file(s)` followed by `JBCT check passed.` on all three changed modules. An earlier run had `aether-deployment` failing format and `node` SKIPPED — the whole gate was re-run after `jbct:format`, never resumed from the failed module, and the test sweep was re-run against the formatted source.

**Symbols, not line numbers** (per the ticket's own stale-citation trap — #805's `1473` now sits inside `capturePriorActiveBlueprint`): `ClusterDeploymentState.blocksSliceActivation`, `.resolveSliceOwner`, `Active.blockingSchemaRecords`, `Active.recordBestEffortFailureOutcome`, `Active.resolveOutcomeOwner`, `Active.bestEffortFailureCommand`, `Active.submitBestEffortFailureOutcome`, `Active.nextOutcomeVersion`, `SchemaRoutes.sliceOwner`, `AetherValue.DeploymentOutcomeValue.fenceVersion`, `KVStore.staleSuccessorWrite`.

siy:
## SHOULD-FIX round complete — **final SHA `87364c91d3b2952f41bccb458a2972271c5f7a6e`**, no further pushes

### SHOULD-FIX 1 — false comment, corrected with a measurement rather than a rewrite

The review is right: removing `awaitQuiescence()` alone leaves the fence-removed red set byte-identical. Rather than swap one unverified explanation for another, I ran the attribution experiment.

**Fence removed + drain removed + `applyBatch` unsynchronized**, five consecutive runs of `$ConcurrentFailures`:

| Run | failing of 4 |
|---|---|
| 1 | 2 |
| 2 | 3 |
| 3 | 1 |
| 4 | **0 — entirely green** |
| 5 | 2 |

Race-shaped, and run 4 reproduces the original near-miss. Restore the `synchronized` alone and the red set is deterministic. **The serialized applier is what removes the race**, modelling production's single applier thread where `KVStore.process`'s read-then-write fence never interleaves with another apply. The drain fixes which *scenario* runs — A's confirm completes before B's merge lands, the adversarial order production reaches — not whether the bug is *detected*.

`$ApplierFence` was deterministic across all five runs (stale-read and version-skip red every time). It is the timing-independent detector, and the comment now says so.

Drain kept, comment corrected in place on `releaseHeld` and `applyBatch`, with the measurement recorded so the next reader can act on it. Two failures here, not one: the near-miss, then a confident wrong explanation of its fix.

### SHOULD-FIX 2 — the deferral reason was wrong, and the fix is already in

`recordBestEffortFailureOutcome` is reached only from the `else` of `if (ctx.atomicity() == DeploymentAtomicity.ALL_OR_NOTHING)`, so it is BEST_EFFORT-only by construction and the ALL_OR_NOTHING risk I cited never existed. The mirror-first-with-committed-fallback fix landed earlier in this branch (`Active.resolveOutcomeOwner`); the bad justification is removed from the report so it cannot be carried forward. The separate conclusion — that a naive source switch regresses because `handleAppBlueprintChange` lets the mirror lead the store — stands on its own evidence.

### Verification at the pinned SHA

Delta from `3accd72e5` is javadoc-only in one test file. `BestEffortOutcomeMergeTest` 15/15 green across 6 nested classes (XML testcase count), and the full gate re-run on all three changed modules: `Running JBCT check on 80 / 88 / 150 Java file(s)` each followed by `JBCT check passed.` The six-module 3950-test sweep at `3accd72e5` still stands for this SHA — a javadoc-only edit in a test file cannot reach the other modules, and I am stating that reliance rather than implying it.

siy:
## Attribution corrected — **final SHA `5a38d6695aba984ba6ee4f2f4cce545a153db098`**

The interleaved experiment settles it and my correction was inverted. Comment-only change; production code untouched.

**What the shipped comment now says**, citing the distributions and their sample size:

| configuration | reds per round (out of 4) |
|---|---|
| drain ON, applier synchronized (shipped) | 4, 4, 4, 4, 4 |
| drain OFF, applier synchronized | 4, 4, 3, 1, 4 |
| drain OFF, applier unsynchronized | 1, 4, 2, 2, 3 |

Rows 1–2 isolate the drain: **it is what makes detection deterministic.** Rows 2–3 isolate `synchronized`: **no measurable effect.** The mechanism agrees — the drain orders A's confirm before B's apply; `synchronized` can only order what happens inside one apply. `ApplierFence` stays named as the timing-independent detector (exactly 2/3 red in all fifteen runs).

**The "may be removed without weakening detection" sentence is gone.** That was the actual defect: not an inaccuracy but an instruction, and following it would have degraded this pin from always-catching to sometimes-catching. `synchronized` is now documented as kept on modelling grounds rather than measured effect, so the next reader knows which of the two is load-bearing and why.

**How the wrong correction happened**, recorded in the comment because it is the transferable part. My five-run measurement was real but covered ONE configuration — I never ran the second. I took the review's single "byte-identical" observation as establishing it, and treated two data points from different sources as a controlled comparison. Both were small samples of a flaky distribution that happened to look clean. More care at any step would not have caught it; interleaving the configurations within rounds did, because it made a load spike unable to explain the result.

Three failures on this one test, not one: the near-miss, a confident wrong explanation, and a confident wrong *correction*. The rule earned: **a claim about a flaky mechanism is not reportable without its sample size and its control.**

**Verification at this SHA:** delta from `87364c91d` is javadoc-only in one test file. `BestEffortOutcomeMergeTest` 15/15 green across 6 nested classes; full gate re-run on all three modules — `Running JBCT check on 80 / 88 / 150 Java file(s)`, `JBCT check passed.` each. The markdown table in the javadoc did not trip the gate. As before I am relying on the six-module 3950-test sweep at `3accd72e5` for the other modules, and stating that reliance rather than implying it.

Issue #952, opened 2026-09-08
start-postgres.sh prints 'PostgreSQL ready' and exits 0 for a container that died 295ms earlier; the failure surfaces 10s later misattributed to mvn install


siy:
Found by a clean-room run of rc4's documented Quick Start on a Hetzner box with no prior state, 2026-09-08. Cost €0.043. Server destroyed, verified.

## What

`./start-postgres.sh` pins `postgres:18-alpine` but mounts the **pre-18 data path**. The container dies immediately — it lives **295 ms** and exits with `ExitCode=1`.

The script then enters its readiness loop. **That loop has no failure branch.** It prints **`PostgreSQL ready`** and exits **0**.

Ten seconds later Forge fails, blaming **`mvn install`**. The Maven artifacts were all present and correct. The real cause — a database container that had been dead for ten seconds — is nowhere in the error.

## Why this is worse than a broken script

A script that fails is a nuisance. **A script that reports success for a thing that is dead is an instrument that lies**, and it does two kinds of damage:

1. **It misattributes.** The next person debugs `mvn install`, which is working. The evidence points away from the defect, so effort is spent proving a healthy component healthy.
2. **It is indistinguishable from working.** `PostgreSQL ready` + exit 0 is exactly what a correct run prints. Nothing downstream can tell the two apart, and neither can CI.

This is the same shape as a gate that exits green having examined nothing: **a readiness check with no failure branch cannot report unreadiness, so its success carries no information.**

## Acceptance

- The mount path matches the pinned image version. Either pin an image whose data path matches, or derive the path from the version rather than hardcoding it.
- **The readiness loop must be able to fail.** It needs a bounded timeout and a non-zero exit, and it must check the container is *running* — not merely that a loop counter elapsed. Prove it: kill the container deliberately and show the script exits non-zero with a message naming the database.
- Forge's error should not attribute a database failure to `mvn install`. At minimum it should report what it actually checked.
- Verify from a machine with **no prior state**. A developer box with a warm Docker cache and an existing volume will pass for reasons a new user does not have.

## Provenance

Not reproducible from reading; found by running the documented path verbatim on a virgin machine. The runtime itself is sound — once repaired, a 7-node cluster booted and the README's `curl` returned 200 with correct arithmetic.

siy:
Closing as fixed and merged.

- **Fix commit:** `2fb21244af8064518cbad3d471336c1266f6f526`
- **PR:** none (committed directly to the release branch)
[...]

Issue #727, opened 2026-08-28
SliceInvocationTest.invokeAfterSliceUndeploy_returnsNotFound: intermittent class-level ERROR at a fixed ~240s on CI runners — three occurrences in one day


siy:
Elevating from #718's observation list to its own actionable ticket: three CI failures of the SAME test at near-identical elapsed times — run 33185195246 (240.5s, docs-only diff), run 33197516374 (240.5s, MAILBOX-only diff), run 33216313569 (240.4s, config-parse diff that cannot plausibly affect undeploy) — each green on rerun, each costing a full forge-tests cycle (~25 min) of branch latency.

The fixed ~240s signature suggests a hard timeout being hit under runner load rather than uniform randomness: the test's wait for the undeploy-then-404 condition presumably has a ~240s ceiling that CI runners intermittently exceed, or the condition needs an event-driven kick rather than a poll (compare the resume-kick class of fix in the #386 dead-letter hold — a hold released only when something re-kicks the loop). Not machine-interference (#715 doesn't apply to isolated GitHub runners).

Acceptance: identify the 240s ceiling's source; either make the wait event-driven, give it honest load margin, or make the failure mode diagnostic (name what it timed out waiting for) — and state which. DAG integration lines: invalidates nothing; no true deps; related #718 (observation home, cross-refs these runs).

Occurrences append here as they happen; the ticket is the durable flake ledger for this test.

siy:
Family ledger addition: run 33216566706 (`9838ecd85`, #386 publisher tier switch) — `ClusterProvisioningDiagnosticsProbeTest` class-level ERROR at **497.6s, identical to the decisecond with its morning occurrence** (run 33129716820, flake-confirmed by identical-tree rerun). Two distinct forge tests now each repeat at their own fixed ceiling (240.5s and 497.6s) — strengthens the hard-timeout-under-runner-load hypothesis over uniform randomness. Caveat on this occurrence: the diff touches publisher PROVISIONING paths, so a real link is not excluded — identical-tree rerun in flight is the discriminator; if it fails again this stops being ledger material and becomes a defect signal against the #386 tier switch.

siy:
[...]

Issue #718, opened 2026-08-28
A per-attempt retry WARN starves cluster formation: stale forge-data triggers backpressure, the log volume exceeds the 60s start deadline, and System.exit(1) kills the node (verified-quiet 2/2 fatal vs 2/2 healthy)


siy:
Observation ticket split out of #704 per CTO instruction; milestone suggestion rc4 (nothing here is a proven correctness failure at rf=1 — the #704 slice failures were functional and independent of this noise — but the volume on an IDLE host is a signal worth its own investigation).

**Observed (single-host 5-node forge, ticketing at rf=1 = 24 instances, otherwise idle machine, 2026-08-28; full log preserved by stream D):**
- ~2965 `Retry.handle()` WARNs and 864 `QuicClusterNetwork.writeIfWritable()` WARNs + 47 `refuseBackpressured` in ~10 minutes.
- 10 x `system stream system:cluster-events:1.0.0 local-partition create FAILED (fail-soft; SystemStreamRegistrar will retry): Stream config consensus commit failed` during startup.
- Repeated `Stale-epoch stream append rejected for system:cluster-events:1.0.0[0]: presented owner epoch 0:0 is older than the partition high-water 1:1` + `applied 0/1 new events ... triggering backfill repair`.
- `RabiaEngine SLOW-APPLY` warns (queueDepth up to ~251 observed).
- Late-run CTM churn: `drainNode requested (target=si-2, reason=JOIN_GRACE_REAP)` — an id from a different naming scheme than the run's node-1..5, suggesting stale persisted forge state on the host contributes; then node-5 removed. Environmental caveat stated in #704.

**Cross-reference, CI scale:** run 33129716820 attempt 1 failed forge-tests via a class-level ERROR in ClusterProvisioningDiagnosticsProbeTest after 497.6s with sustained `Stream config consensus commit failed` WARNs; attempt 2 green on the identical docs-only sha c6de6f55e — flake-shaped, same signature family as the local observations.

**Questions this ticket owns:** why does a 5-node single-host cluster generate thousands of retry/backpressure events with 24 slice instances and zero load; is the stale-epoch append storm on system:cluster-events expected during formation; does stale forge-data on disk (the si-* reap) deserve a startup guard; and is the CI probe flake the same mechanism surfacing under CI resource contention.

siy:
[...]

Issue #491, opened 2026-07-18
Stream backfill catch-up stalls: promoted owner has no QUIC peer connection to the survivor after owner-kill churn (post-#487 WARN evidence)


siy:
Exposed by #487's new rate-limited WARN during the #457 acceptance run (2026-07-19). With self-send loopback ACTIVE (so this is definitively NOT the #467/#487 class): the `stream-partition-backfill` thread still drops sends — `WARN No peer state for NodeId[id=sof-2] — dropping message (dead or never-connected peer)` — where sof-2 is the LIVE survivor holding all data. The promoted owner's catch-up-from-survivor stalls at watermark -1 indefinitely; RF restoration never converges in-JVM. 14 WARNs in the run, ALL peerId != self (sof-2 ×6, sof-5 ×5, sof-1/3/4 ×1) → the churned post-failover forge mesh is missing peer connections broadly.

Two hypotheses to discriminate BEFORE fixing (observe-then-fix):
1. **Product race:** backfill/catch-up starts before the mesh (re)dials the catch-up source after ownership churn — no ensure-connection/await-dial in the backfill path, no retry-on-drop. Would exist in cloud too, masked by timing/retries (cloud 02-chaos passes RF restoration).
2. **Forge/in-JVM mesh limitation:** the single-JVM Ember mesh doesn't re-establish or never established certain peer pairs after a kill.

Prior art (read first — rich history of in-JVM QUIC dial gaps): the #94-roots arc (three QUIC dial mechanisms fixed), the QUIC reconnection-storm fix (2-part PEERS root + guards/dedup), connectedPeerCount memories. Also likely related: the chronic forge slice-ROUTING "cross-node acks not received" force-transitions seen on multiple unrelated tests — same missing-connection signature, now observable via the #487 WARN.

Repro: StreamOwnerFailoverTest phase 9 (soft-observed pointing here; harden to a hard assert when this closes — the machinery is retained). Relates: #457 (test), #487 (loopback + the WARN that exposed this), #445 arc.

siy:
# Diagnosis: #491 — RF-restoration stall after owner kill (in-JVM forge)

## VERDICT: BOTH
Forge/in-JVM-amplified membership+transport churn cascade (H2, dominant) on top of a
genuine product gap (H1). NOT "backfill fires before the mesh forms." Real mechanism:
the graceful leader-kill tears down LIVE survivor<->survivor QUIC links, which re-form
[...]

PR #1010, opened 2026-09-11
fix: make Forge's QUIC base port configurable and fail loudly on a collision (#1008)

### LINKED ISSUE #1008: Forge's QUIC base port is hard-coded, so a second instance silently never reaches quorum instead of failing to bind
**Forge's QUIC base port is a hard-coded constant, so two Forge instances cannot coexist on one host — and the second one fails SILENTLY.** This makes this project's own concurrency protocol unachievable for Forge runs, and it cost real debugging time today.

## The defect

`EmberCluster.java:93` defines `DEFAULT_BASE_PORT = 6000`, and `ForgeServer.java:243` passes it **as a constant**. Meanwhile `ForgeConfig` exposes the **management, app and dashboard** ports — and **never references the QUIC base port at all.**

So three of four ports are configurable and the fourth is not. That asymmetry is what makes this easy to miss: an operator or agent who carefully reconfigures the ports `ForgeConfig` exposes will still collide on 6000–6004.

## CORRECTED 2026-09-11: the failure is silent ONLY WHEN YOU FOLLOW THE CONCURRENCY PROTOCOL

The first version of this issue claimed a second Forge fails silently in all cases. **That is wrong, and the corrected mechanism is a sharper argument for fixing this, not a weaker one.**

**Default against default is LOUD and FAST.** Measured twice: a second stock Forge dies in **~2 s** (one arm at 1.24 s) with **12× `Address already in use`**, **zero** retry WARNs and **zero** `STREAM_LIMIT_ERROR`. `SO_REUSEADDR=true` is set, but **TCP 5150 fails first** — the management-port collision catches the process before QUIC is ever reached.

**The silent mode appears when the configurable ports are changed.** `ForgeConfig` lets you move management, app and dashboard. Move them — which is exactly what this project's own concurrency protocol instructs — and the loud TCP guard is gone, while the QUIC base port **stays hard-coded at 6000**. The second instance now binds its management port happily, collides invisibly on QUIC, and never reaches quorum (`activePeerCount=1`) with no message naming a port.

**So following the protocol converts a loud, immediate failure into a silent, slow one.** An operator who changes nothing is protected by accident; an operator who does the careful thing is not. That inversion is the defect.

**And `activePeerCount=1` is a shared symptom** with genuinely stale on-disk state, so once you are in the silent mode the two causes are indistinguishable from the output.

## Measured harm, 2026-09-11

Two agents working in parallel on this repository collided on it within one hour:

1. **An investigation into #718 had its experimental arms contaminated.** It had taken the suite lock; a second Forge started afterwards on the default ports and held 6000–6004 for roughly 22 minutes. Arms run in that window were discarded.
2. **A causal conclusion was published and then retracted.** The second agent measured an 8s-vs-62s-vs-125s cluster-formation spread and attributed it to stale `forge-data` — a plausible, well-evidenced-looking before/after. It was a confounded measurement: its own second Forge produced `activePeerCount=1` directly. Its own diagnosis of the error: *"`activePeerCount=1` was the shared symptom of both hypotheses and I attributed it to the one I'd been reading about. I never asked what else produces that line; the answer was a constant in the file I already had open."*
3. A related harm in the same window: `pkill -f aether-forge.jar`, used as cleanup, matches by command line **across every working tree** and SIGTERMed the investigation's process. A 143 exit read as a product failure until it was traced. (The runbook that recommended `pkill` has since been fixed — see PR #1007 — but the underlying reason anyone reached for a name-matching kill is that there is no way to tell two Forge instances apart by port.)

## Expected

1. **Make the QUIC base port configurable through `ForgeConfig`, like the other three.** Symmetry alone would have prevented this.
2. **Fail loudly when the QUIC port is unavailable.** Configurability is necessary but not sufficient. Note the default path already fails loudly *by accident*, because TCP 5150 is checked first — so the guard must be on the QUIC bind itself, not inherited from whichever port happens to be checked earliest. If 6000–6004 (or the configured range) is already bound, say so and exit non-zero rather than forming a one-node cluster that looks like a consensus problem.
3. Consider surfacing `activePeerCount=1` with its likely causes enumerated, since it is a genuinely ambiguous symptom.

## Note on scope

This is infrastructure for *working on* the product rather than a product-facing defect, which is why it is filed separately from #718 rather than bundled into its fix. But it is the reason #718's diagnosis took four wrong mechanisms before the right one, so its cost is not hypothetical.

siy:
Fixes #1008.

## The correction that shaped this fix

The issue asks for a guard "on the QUIC bind itself". **There is no bind failure to guard.** Measured
on this box (Darwin 25.5.0), all four holder/probe combinations:

| holder | probe | result |
|---|---|---|
| `SO_REUSEADDR` | `SO_REUSEADDR` | **binds — collision invisible** |
| `SO_REUSEADDR` | plain | refused, `Address already in use` |
| plain | `SO_REUSEADDR` | refused |
| plain | plain | refused |

`QuicClusterServer` sets `SO_REUSEADDR` (deliberately — a restarting node must rebind its own port),
so Forge-versus-Forge is exactly the one row that succeeds. `QuicTransportError.BindFailed` already
names the port and `EmberCluster.start()` already aborts the whole start on it — that path simply
never fires. Propagating bind failures harder would have changed nothing.

So the guard is a **preflight that binds each port WITHOUT `SO_REUSEADDR`**, which the kernel does
refuse while any holder exists, run before a single node is created.

## What changed

1. **`[cluster] base_port`** is now a `forge.toml` setting (default `6000`, unchanged). `EmberConfig`
   gains a `basePort` component and validates the whole range (`base_port + nodes - 1 > 65535` is
   refused). `ForgeServer` and `EmberInstance` read it instead of passing the constant positionally.
   Every existing `emberConfig(...)` overload keeps its signature and defaults the new value.
2. **`ForgePortPreflight`** refuses startup and exits non-zero, naming every occupied port. The
   message **discriminates the two producers of `activePeerCount=1`**, stating that no node was
   started so it is not stale `forge-data` state — the ambiguity that cost the debugging time.
3. Startup banner prints the QUIC range; forge-guide documents `base_port`; the
   lifecycle-verification runbook said no configuration overrides the QUIC base port — true when
   written, false now, corrected.

## Verification

**Root build** `env -u HCLOUD_TOKEN mvn clean install` from ROOT, no `-pl`, no `-DskipTests`:
BUILD SUCCESS, 145 modules, 0 `[ERROR]`, 0 SKIPPED/FAILURE. **13,748 `<testcase>` elements**
(counted as elements, not the `<testsuite tests=>` attribute), 0 failures, across surefire **and**
failsafe reports newer than a pre-build marker; the 5 files the marker excluded are 4 days old.

**Gate**, run separately per module, quoting file counts:
`aether/ember` → `Running JBCT check on 6 Java file(s)` / `JBCT check passed.`;
`aether/forge/forge-core` → `Running JBCT check on 3 Java file(s)` / `JBCT check passed.`
(The gate caught a format violation and a `JBCT-EX-01` lint error the lifecycle build did not.)

**Mutation probes** — each reverts one hunk against the committed base; red sets are named:

| probe | reverted | RED set |
|---|---|---|
| P1 | probe sets `SO_REUSEADDR` | 3: `whenHolderSetsReuseAddr`, `messageDiscriminatesFromStaleForgeData`, `namesEveryOccupiedPort` |
| P2 | `base_port` no longer read from TOML | 5 in `EmberConfigBasePortTest` |
| P3 | range-overflow check removed | 1: `basePort_isRejected_whenRangeForNodesExceedsPortSpace` |
| P4 | **both `ForgeServer` wiring hunks** | **0 — unpinned by unit tests** |

P1's set is disjoint from P2/P3's. **P3's set is a strict subset of P2's** (removing the TOML read
means `base_port` never reaches validation) — stated rather than claimed as independence.

**P4 is an honest gap**, so the wiring is pinned by a live Forge run instead (suite lock held; killed
by PID, never `pkill -f`):

- **Arm A** — a holder owns UDP 6000-6004 with `SO_REUSEADDR` (confirmed by `lsof`), second Forge on
  the default range: **exit 1**, ~15 ms after the banner, message naming all five ports.
- **Arm B** — same holder still on 6000-6004, `base_port = 7300`: Forge starts, **binds exactly
  7300-7304** (`lsof` on its PID), reaches **full membership 5/5, `quorumSafe=true`**, leader
  elected, zero bind errors, no collision message.

Arm B proves `base_port` reaches `EmberCluster` end-to-end; Arm A proves the collision is loud.

## Limitations, stated

- The preflight is **time-of-check/time-of-use**. Each socket closes before the cluster binds for
  real, so a process claiming a port inside that window is still silent. It converts the common case
  (another Forge already running) from silent to loud; it is not a lock.
- The duplicate-bind measurement is from **Darwin 25.5.0**. On Linux a duplicate UDP unicast bind
  under `SO_REUSEADDR` is refused anyway, so the preflight is correct there but less load-bearing.
  Consequently `ensureQuicPortsFree_fails_whenHolderSetsReuseAddr` is a genuine discriminator on
  macOS and passes vacuously on Linux.
- Test sources are not gated (`jbct.includeTests=false`, repo-wide, pre-existing).
- **Issue item 3 (enumerating causes on a live `activePeerCount=1`) is NOT done.** It belongs to the
  membership/status surface, not Forge startup. This fix removes the port collision from that
  symptom's candidate set at the only point where it can be ruled out cheaply.

### DIFF
--- a/aether/docs/operators/runbooks/lifecycle-verification.md
+++ b/aether/docs/operators/runbooks/lifecycle-verification.md
@@ -20,15 +20,20 @@ not reach the router at all. It falls through to the static file handler, which 
    ls aether/cli/target/aether.jar
    ```
 
-3. **Confirm no other Forge is running on this machine first.** Only one Forge instance can run
-   per host. Its management, app and dashboard ports come from `ForgeConfig`, but the cluster's
-   QUIC base port does not: `ForgeServer` passes the `EmberCluster.DEFAULT_BASE_PORT` constant
-   (`6000`, so nodes occupy `6000`–`6000+nodes-1`), and no configuration overrides it. There is
-   therefore no port-offset arrangement that lets two instances coexist — a second instance's nodes
-   cannot form a cluster, and the symptom is step 2 never returning rather than a bind error.
+3. **Confirm no other Forge is running on this machine first.** All four of Forge's port settings
+   are configurable in `forge.toml` — `base_port` (the cluster's QUIC/consensus UDP range,
+   `base_port`–`base_port+nodes-1`, default `6000`), `management_port`, `app_http_port` and
+   `dashboard_port`. Moving all four lets a second instance coexist; moving only the other three
+   does not, because the QUIC range is the one whose collision the kernel does not report (the
+   sockets carry `SO_REUSEADDR`, so a duplicate bind succeeds and the datagrams are split silently).
+
+   Since #1008 Forge checks its whole QUIC range at startup and exits non-zero naming the occupied
+   ports, so that collision is loud rather than appearing as step 2 never returning. A start that
+   fails this way is a port collision and not stale `forge-data` state — no node was started.
 
    ```bash
    lsof -nP -iTCP:5150 -sTCP:LISTEN    # expect no output before you start
+   lsof -nP -iUDP:6000                 # QUIC base port; expect no output before you start
    ```
 
 4. Start from clean simulator state. Forge keeps per-node data under `$AETHER_HOME/forge-data`,
--- a/aether/docs/slice-developers/forge-guide.md
+++ b/aether/docs/slice-developers/forge-guide.md
@@ -153,7 +153,14 @@ base_port = 6000             # Base QUIC/consensus UDP port (node i binds base_p
 management_port = 5150       # Base management port
 dashboard_port = 8888        # Dashboard port
 app_http_port = 8070         # Base app HTTP port (load target)
+start_timeout_seconds = 60   # How long to wait for the cluster to finish forming before exiting
 ```
+
+If the cluster does not finish forming inside `start_timeout_seconds`, Forge exits non-zero and
+reports what it had reached at the deadline — how many nodes were consensus-active, the leader (or
+`none`), each node with its state and QUIC port, and any per-node start failure. Raise the value if
+formation on your host is merely slow; a stale `AETHER_HOME/forge-data` directory is a common reason
+for formation to consume the whole budget.
 
 ### Running two Forge instances on one host
 
--- a/aether/docs/slice-developers/forge-guide.md
+++ b/aether/docs/slice-developers/forge-guide.md
@@ -149,10 +149,23 @@ Three things worth knowing:
 
 [cluster]
 nodes = 5                    # Number of nodes to simulate
+base_port = 6000             # Base QUIC/consensus UDP port (node i binds base_port + i)
 management_port = 5150       # Base management port
 dashboard_port = 8888        # Dashboard port
 app_http_port = 8070         # Base app HTTP port (load target)
 ```
+
+### Running two Forge instances on one host
+
+All four port settings must be moved, `base_port` included. It is the cluster's QUIC/consensus range
+(`base_port` through `base_port + nodes - 1`), and it is the one that does **not** announce a
+collision by itself: the QUIC sockets are bound with `SO_REUSEADDR`, so a second instance left on the
+default 6000 binds successfully, silently splits the range's datagrams with the first instance, and
+never reaches quorum — surfacing only as `activePeerCount=1`, which is also what genuinely stale
+`forge-data/` state looks like.
+
+Forge therefore checks the whole range at startup and refuses to start when any of it is held,
+naming the ports. If you see that message it is a collision, not stale state — no node was started.
 
 ### Auto-Healing
 
--- a/aether/ember/src/main/java/org/pragmatica/aether/ember/EmberConfig.java
+++ b/aether/ember/src/main/java/org/pragmatica/aether/ember/EmberConfig.java
@@ -20,6 +20,13 @@ import org.pragmatica.lang.Result;
 /// collide invisibly: relocating the three configurable ports removed the loud TCP guard while
 /// leaving the QUIC range fixed. Configurability alone does not make that collision visible (the
 /// duplicate UDP bind SUCCEEDS under `SO_REUSEADDR`); `ForgePortPreflight` is the guard.
+///
+/// #718 shape 4 — `startTimeoutSeconds` is how long Forge waits for the cluster to finish forming
+/// before giving up and exiting. It was a hard-coded 60-second literal at the single `await` site,
+/// so a host on which formation is merely slow had no way to buy more time. Deliberately NOT derived
+/// from, and not coupled to, the 60-second higher-id grace in `QuicClusterNetwork` (#491): the two
+/// are numerically equal and serve unrelated purposes, and no evidence of a real relationship was
+/// found.
 public record EmberConfig(int nodes,
                           int basePort,
                           int managementPort,
@@ -29,7 +36,8 @@ public record EmberConfig(int nodes,
                           ObservabilityConfig observability,
                           boolean lbEnabled,
                           int lbPort,
-                          int coreMax) {
+                          int coreMax,
+                          int startTimeoutSeconds) {
     public static final int DEFAULT_NODES = 5;
     /// Single source of truth stays [EmberCluster#DEFAULT_BASE_PORT]; mirrored here so config
     /// callers need not reach into the cluster class for a default.
@@ -40,6 +48,7 @@ public record EmberConfig(int nodes,
     public static final boolean DEFAULT_LB_ENABLED = true;
     public static final int DEFAULT_LB_PORT = 8080;
     public static final int DEFAULT_CORE_MAX = 0;
+    public static final int DEFAULT_START_TIMEOUT_SECONDS = 60;
 
     public static final EmberConfig DEFAULT = new EmberConfig(DEFAULT_NODES,
                                                               DEFAULT_BASE_PORT,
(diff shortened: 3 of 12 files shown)


siy:
## Scope extension: #718 shape 4 folded in (CTO ruling 2026-09-11T14:20Z)

Same class of defect in the same file, so it lands here rather than fighting for `ForgeServer`.

### 1. The start budget is configurable
`[cluster] start_timeout_seconds` (default `60`, unchanged). It was a hard-coded literal at the
single `await` site, so a host where formation is merely slow had no way to buy more time.
`EmberConfig` gains `startTimeoutSeconds` and rejects a non-positive budget — which would expire
before the cluster could possibly form. Both consumers read it (`ForgeServer.startCluster` and
`EmberInstance`, which carried the same literal).

### 2. The timeout message names what did not finish
Old text: `"Failed to start cluster: " + cause.message()` → `Promise is not resolved within specified
timeout`. **Identical whether nothing formed or four of five nodes were consensus-active.**

Real output from a live run with `start_timeout_seconds = 1`:

```
Cluster did not finish forming within the 1s start budget (cluster.start_timeout_seconds): Promise is
not resolved within specified timeout — exiting rather than leaving a cluster that never formed
looking healthy. Reached at the deadline: 4 of 5 node(s) consensus-active, leader=none. Per node:
node-1=active(quic 7300), node-2=active(quic 7301), node-3=active(quic 7302), node-4=active(quic
7303), node-5=inactive(quic 7304). No node reported a start failure, so the budget expired with the
start still in progress rather than with a node erroring out. 'consensus-active' is
AetherNode.isReady(), NOT a general health verdict. Forge's startup preflight verified UDP 7300-7304
was free immediately before this start, so a QUIC port collision at that moment is ruled out. Forge
did NOT check, and any of these is consistent with what is reported above: stale per-node state under
AETHER_HOME/forge-data (a retry/backpressure storm during formation can consume this whole budget);
host load; or a genuine consensus fault. If formation on this host is merely slow, raise
cluster.start_timeout_seconds.
```

Every value is read from `EmberCluster.status()` at the deadline; none is inferred. It names the one
thing Forge **did** verify (the #1008 preflight) so the reader can rule it out, and lists unchecked
candidates without ranking them — the #952 discipline. It never calls a node "healthy": `state` is
`AetherNode.isReady()`, labelled as such, because #727's defect was an unconditional `"healthy"`
literal that made every failing cluster report healthy nodes beside `leader=none`.

### Non-goal honoured, plus a finding
**Not coupled to the 60 s higher-id grace, and not derived from it.** I looked but did not act: that
grace lives in `QuicClusterNetwork.higherIdGracePeriodElapsed` and its own comment scopes it to
avoiding the cold-boot dual-Hello race for a **never-connected** peer (#491). Different mechanism,
different lifecycle. **Finding, reported not acted on: the two are numerically equal (both 60) and I
found no evidence of a real relationship in either direction** — no derivation, no shared constant,
no reference. Equal literals with unrelated documented purposes is exactly the coincidence the ruling
fences off, so both remain independent.

### Verification
Root build `env -u HCLOUD_TOKEN mvn clean install` from ROOT: **SUCCESS, 145 modules, 0 `[ERROR]`, 0
SKIPPED/FAILURE, 13,762 `<testcase>` elements, 0 failures** (element count across surefire + failsafe,
marker-relative; the 5 excluded files are 4 days old). Gates re-run separately per module after the
change: `aether/ember` → `Running JBCT check on 6 Java file(s)` / passed; `aether/forge/forge-core` →
`Running JBCT check on 3 Java file(s)` / passed. **Both failed on format first and were fixed** — the
green reactor caught neither.

Mutation probes (baseline 32 testcases, RED 0):

| probe | reverted | RED |
|---|---|---|
| P5 | `start_timeout_seconds` not read from TOML | 4 in `EmberConfigStartTimeoutTest` |
| P6 | budget hard-coded back to 60 **in the message** | 1: `clusterStartFailure_readsBudgetAndPortRangeFromTheConfig` |
| P7 | drop active-count + per-node detail | 5 `ClusterStartFailure` tests |
| P8 | fabricate `"healthy"` (the #727 defect) | 2, incl. `clusterStartFailure_neverCallsANodeHealthy` |
| P9 | **both `await` sites back to the literal 60** | **0 — unpinned by unit tests** |

**P6 is worth calling out as an instrument finding.** My first two message assertions read the budget
and port range from `EmberConfig.DEFAULT`, whose values *are* 60 and 6000 — so a hard-coded literal
would have satisfied them **vacuously**. I added `clusterStartFailure_readsBudgetAndPortRangeFromTheConfig`,
driving the same message from a non-default config (`nodes=3, base_port=7300, start_timeout_seconds=180`).
P6 reddens **that test and only that test**, which is the proof the other two could not have caught it.
P8 likewise proves the "never says healthy" pin is not vacuous — it can actually fail.

**P9 is an honest gap**, the same class as P4: the `await` sites are not unit-testable. Pinned by a
controlled live run instead (suite lock claimed and released; killed by PID, never `pkill -f`;
`AETHER_HOME` a fresh `mktemp -d` per arm):

- **T1** — `start_timeout_seconds = 1`: exit 1, the message above, reporting a real partial formation.
- **Control T60** — byte-identical config except `start_timeout_seconds = 60`: **cluster forms,
  `reachedFullMembership=true quorumSafe=true`, `Forge server running`, zero timeout diagnostics.**

Only the budget differs between the arms, so T1's failure is attributable to the configured value
rather than to a broken cluster or a loaded host.

### Instrument note for whoever reads the #1008 evidence
The CTO flagged that `lsof -iUDP` and `netstat -p udp` fail their own control on this box and cannot
see docker-published UDP. My `lsof` use was confined to **local JVMs**, where it did report the
sockets, and every load-bearing detection was a **real bind attempt** (the preflight itself, and the
port-release check above). So that half of the evidence does not rest on `lsof`.

Issue #1032, opened 2026-09-12
cluster destroy cannot complete a graceful drained teardown: draining every node necessarily violates the cluster's own disruption budget


siy:
Measured on a live 5-node Hetzner cluster, 2026-09-12, while verifying the #1023 fix.

## What happens

`cluster destroy` drains nodes one at a time. The cluster refuses those drains with **HTTP 409**:

```
Disruption budget exceeded: draining would leave 2 core-scoped operational nodes, minimum is 3
```

Measured outcome on a 4-node surviving cluster: `Drains succeeded: 0/3`, `Shutdowns succeeded: 3/3`. Two nodes were refused with 409; a third accepted the drain but never reached `DECOMMISSIONED` within 120 s.

## Why it is a design gap, not a bug in either component

The disruption budget is **correct**. It exists to stop an operator draining a cluster below quorum, and it did exactly that.

`cluster destroy` is also correct to drain before shutting down.

But **draining every node necessarily violates the budget** — by the time enough nodes are drained, the remainder is below the minimum by construction. There is no way for `destroy` to express *"I am tearing the entire cluster down, the budget's purpose does not apply."* So a graceful drained teardown of a whole cluster is **not currently achievable at all**.

A 409 is an application-level answer, which is itself informative: **TLS, authentication, routing and leader-forwarding all worked.** The drains reached the cluster and were deliberately refused. Nothing below this layer is at fault.

## Why it was invisible until now

Before #1028, `destroy` never got far enough to reach a drain — it failed earlier on a dead registry endpoint or on PKIX. It reported `Drains succeeded: 0/0` **under a success line**, which reads as "nothing needed draining" rather than "drain never ran".

With honest reporting (`0/3`, plus an explicit warning, and no "destroyed successfully"), the refusal became findable on the first real run. **The legibility change is what surfaced this.**

## Explicitly NOT done

The guard was **not** lowered to make the numbers look better. Any fix must preserve the budget's protection for ordinary drains.

## Suggested shape

A teardown-scoped intent that the budget honours — e.g. draining under an explicit whole-cluster-destroy context — so ordinary drains stay protected while a deliberate full teardown can proceed. The 120 s `DECOMMISSIONED` timeout on the node that did accept is a separate question worth its own measurement.

siy:
[...]

Issue #1023, opened 2026-09-11
cluster destroy cannot run a graceful drain on a TLS-auto_generate cloud cluster: registry endpoint is a SPOF, then PKIX path building fails


siy:
Observed on a live 5-node Hetzner cluster with `[operations.tls] auto_generate = true`, 2026-09-11.

`cluster destroy` refused twice — **correctly, the refusal is #998's gate working** — but in doing so revealed that a graceful drain is unreachable on this configuration. Final summary: `Drains succeeded: 0/0, Shutdowns succeeded: 0/0`. **`0/0` is a skip, not a pass.**

## Two independent failures

**1. The registry stores ONE node's address as the cluster endpoint — a single point of failure.**

The registry entry held `endpoint = "https://138.199.236.244:8080"`. That node had been deliberately killed as part of the test, so `destroy` got a `ConnectException` — **while three healthy nodes were available to serve the same enumeration.** A cluster endpoint that names one node cannot survive that node, which is the situation teardown most often runs in.

**2. With the endpoint repaired to a live node: `SSLHandshakeException … PKIX path building failed`.**

The CLI cannot build a trust path to the cluster's own auto-generated CA. So for any cluster using `auto_generate` TLS, drain and shutdown cannot run at all and the operator must pass `--force-undrained` to tear down.

*The PKIX mechanism is OBSERVED, not traced — the destroy path's trust setup was not read.*

## Why it matters

Drain exists to let a node hand off work before it disappears. An operator following the documented teardown path on a TLS-auto_generate cloud cluster never drains anything, and the only signal is a `0/0` that reads like success.

**Suggested shape:** the registry should carry all known member addresses and `destroy` should try them in turn rather than trusting one; and the CLI needs a trust path to a cluster-generated CA, or an explicit, documented reason why it cannot have one.

PR #1035, opened 2026-09-12
docs: record the #1032 blocker and live verification in the #1023 changelog

siy:
Changelog-only follow-up to #1028, which merged before its author's final changelog commit landed. That was my error as the merging party, not the author's.

#1028's merged changelog fragment describes the drain claim as *"[design intent — unverified]"* and carries **zero** references to #1032. This adds what the live cloud run established and — more importantly — what it did **not**:

- **"This does NOT make drain work, and #1023 stays open."** The live 5-node run built the trust path and fell through from a deleted endpoint to a live member, then the cluster **refused every drain with HTTP 409, "Disruption budget exceeded"**. Tracked as **#1032**. Without this line the fragment reads as though drain now works.
- **The live verification**, with its mechanism: a CA derived from the recorded secret validated a certificate issued by a real node at first leadership (`curl` without it `ssl_verify=20`, with it `ssl_verify=0`).
- **An operator-facing cost this introduces**: a fully unreachable N-node cluster now takes roughly **(N−1) × 130 s** before refusing — about 17 minutes for a dead 9-node cluster, in an outage, which is when `destroy` is actually run.
- **A scope limit the live run exposed**: `collectedAddresses` is a bootstrap-time snapshot written once by `BootstrapPhaseCollect` and never updated, so **a node that auto-heal replaces later at a new address is not a fallback candidate at all.** It happened to be covered on this run because the node that answered was ledger-recorded.

No code change. One file.

**Note for anyone tempted to merge the original branch instead:** `fix/1023-destroy-drain-spof-2026-09-12` is based on an rc4 predating #1016 and #1026, so merging it would revert both. This branch carries the single changelog file onto the current tip.

### DIFF
--- a/changelog.d/1023-destroy-drain-endpoint-spof-and-cluster-trust.md
+++ b/changelog.d/1023-destroy-drain-endpoint-spof-and-cluster-trust.md
@@ -56,12 +56,38 @@ persists** — no new field was added to the registry or the bootstrap state.
   [mechanism: the installed anchor is the derived cluster CA and nothing else, so a certificate from a
   different cluster's secret has no path to it — #209's MITM property, preserved on the destroy path]
 
+- **This does NOT make drain work, and #1023 stays open.** Verified on a live 5-node Hetzner cluster
+  with the registry-named node deleted at the provider: the trust path was built, the CLI fell through
+  to a live member, and enumeration succeeded — then the cluster **refused every drain with HTTP 409,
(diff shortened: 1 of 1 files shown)


siy:
Closing as a duplicate of #1036, which carries the identical file (both resolve to sha256 d1fcda2068c99992, one changed file each) and preserves the author's commit. The author and I independently caught the same gap within minutes and each opened a fix; that redundancy is fine, the duplication was my failure to say I was doing it. #1036 is the one to merge.

Issue #675, opened 2026-08-27
[operations.auto_heal] tunables are dead beyond enabled — three config surfaces parse auto-heal knobs that nothing reads; collapse them


siy:
Evidence from streams C (#575 work) and E (docs audit), both code-verified.

**Surface 1 — AutoHealSpec → discarded.** Main.resolveAutoHeal (Main.java:344-350) always builds runtime AutoHealConfig from AutoHealConfig.DEFAULT, overriding only maxNodes — which it sources from [cluster] max_nodes, a DIFFERENT TOML key, not [operations.auto_heal] at all. So retry_interval, startup_cooldown, stale_observation_ttl, quic_miss_promotion_threshold, provisioning_timeout, provision_stability_window, decommissioned_retention, swim_hints_ttl all parse into AutoHealSpec and are never read again. The duplicated types also disagree on decommissionedRetention's default (24h vs 60s). enabled=false is already loudly rejected (PF-25, ClusterBootstrapConfigValidator.java:81-91, #575).

**Surface 2 — TimeoutsConfig, distinct record, same fate.** [timeouts.scaling] auto_heal_retry / auto_heal_startup_cooldown (TimeoutsConfig.java:231-232, parsed at ConfigLoader.java:804-809) have ZERO call sites for their accessors outside parsing.

**Scope:** decide ONCE which auto-heal knobs the runtime actually honors and where they live, then collapse: wire or PF-25-style-reject every field on every surface (silent-discard is ruled out by the config-honesty standard, see #575/#578), eliminate the AutoHealSpec/AutoHealConfig duplication, and reconcile the divergent default. Fix sites span Main.java/AetherNodeConfig.java (cluster-core territory) and TimeoutsConfig/ConfigLoader (operator-surface territory) — execution may split per territory, but the design decision is one decision.

Docs: bootstrap-config.md and timeout-configuration.md are being corrected now to describe current truth citing this ticket (docs stream).

siy:
Docs half done — `bootstrap-config.md` and `timeout-configuration.md` now describe current truth for all three affected surfaces:

[...]

PR #1176, opened 2026-09-14
test: rc4 base break — ClusterTopologyManagerRoleMismatchTest uses the collapsed four-knob AutoHealConfig (#675)

siy:
rc4's own push-CI has failed on its last two pushes (`c55933cf1`, `e4c697c12`): `build-and-test` and `forge-tests` both fail identically on an `aether-deployment` **testCompile** error.

**Cause — two same-batch merges, each green alone.** #1128 collapsed `AutoHealConfig` to the four knobs the runtime actually reads, deleting `DEFAULT_STALE_OBSERVATION_TTL`, `DEFAULT_QUIC_MISS_PROMOTION_THRESHOLD` and the wider factory overloads. `ClusterTopologyManagerRoleMismatchTest` was last touched by #1120 in the same 11:00Z merge batch and still calls the 6-arg form. Same shape as the `ArtifactStore` break that #1160 repaired.

**The fix is positional, and the positions moved.** The old signature began `(retryInterval, startupCooldown, ...)`; the new one begins `(startupCooldown, provisioningTimeout, ...)`. So the test's leading `timeSpan(60).seconds()` was **retryInterval** — one of the reader-less fields #1128 removed — and its real `startupCooldown` was the second argument, `timeSpan(1).millis()`. Mapping the arguments by position would have set a 60-second startup cooldown and stalled the test's formation check. The replacement preserves the two values that survive the collapse, and matches every sibling test in this family (`ClusterTopologyManagerAutoHealDurabilityTest`, `...RenderUserDataTest`, `...WorkerReconcileTest`, `...ZoneRotationTest` all use `autoHealConfig(timeSpan(1).millis(), DEFAULT_PROVISIONING_TIMEOUT)`).

**Blast radius.** Search space: `git grep` over the whole tree at `e4c697c12`; positive control `DEFAULT_PROVISIONING_TIMEOUT` = 9 files. `DEFAULT_STALE_OBSERVATION_TTL` and `DEFAULT_QUIC_MISS_PROMOTION_THRESHOLD` each resolved to **exactly this one file**. `DEFAULT_RETRY_INTERVAL` hits only `QuicTransportTuning` (an unrelated constant on another class); `DEFAULT_DECOMMISSIONED_RETENTION` only CHANGELOG.md and an archived handover. No production code is affected — this is test-only.

Test-only repair of a broken base; no changelog fragment.


siy:
Superseded by #1193, which merged as `8fee60080`.

This PR's two commits — the `ClusterTopologyManagerRoleMismatchTest` repair (#1128 × #1120) and the `EmberClusterTeardownWiringTest` `isLeader` repair (#1070 × #1126) — were stacked into #1193 alongside the `SmtpSession` channel race (#1185), because **none of the three could go green alone**: a PR branched from the tip inherited break 1, while this PR inherited break 3. Stacking made one CI run able to prove all three.

Verified by content at the merged tip `8fee60080`: stale `AutoHealConfig` symbol **0** (positive control: file readable, 479 lines), `"isLeader"` in the ember fake **1**, SMTP `setChannel` **1**.

The incident itself is tracked as **#1177** — nothing gates the *merge result* of a symbol-removal PR against a concurrent PR that still uses the symbol. Three instances in one day.

Issue #685, opened 2026-08-27
auto-heal disable (#603) is in-memory only and silently reverts on leader failover


siy:
Follow-up from #603 (auto-heal disable now actually gates provisioning). The gate itself is real and verified end-to-end, but its state does not survive the one event an operator disables auto-heal *because of*.

## The gap

`aether cluster topology auto-heal disable` flips an in-memory flag on the current leader. It suppresses provisioning **for the current leader term only** — it does not survive a leader failover. A new leader starts with auto-heal enabled regardless of what the previous leader was told, silently re-arming a suppression the operator explicitly requested, exactly when the cluster is already having a bad time (a leader failover is itself often the reason someone reached for the kill switch).

## Interim recovery action (needed now, independent of any code fix)

After any leader failover, an operator who disabled auto-heal must re-check `aether cluster topology status` and re-issue the disable if needed. This needs to be documented in both places the quad requires it:
- `cli.md` — `auto-heal disable` section
- The corresponding management-api.md guarantee row

This is not optional cleanup — without it the documented behavior (disable = disabled) silently stops being true after a failover with no operator-visible signal.

## Eventual fix direction (not a commitment — rc4 triage decides)

Make the disable state replicated/consensus-backed instead of leader-local in-memory, so it survives a leader change the same way other operator intents do. Direction only; sizing and priority against rc4 scope is a separate call.

## Acceptance

- Recovery action documented in cli.md + management-api.md (independent of the code fix, can land immediately).
- Either the disable state survives leader failover, or the status endpoint/CLI makes the leader-local, term-scoped nature of the flag explicit enough that an operator can't be surprised by it.

siy:
Fixed by PR #867, merged into release-1.0.0-rc4 on 2026-09-05. Merged-base push-CI concluded SUCCESS on b16ba5688 across build-and-test, forge-tests and fragment, so the eight-PR combination is verified, not just each PR individually. Closing manually because GitHub does not auto-close linked issues when a PR merges into a non-default branch.

Issue #603, opened 2026-08-17
aether cluster autoheal off is cosmetic: the flag has one reader and it is the status route


siy:
## Summary

`aether cluster autoheal off` reports success and changes the status endpoint, but **does not disable auto-heal**. The flag it flips has exactly one reader, and that reader is the status route.

## Evidence (read at HEAD `15e1dfd2c`)

`autoHealEnabled` is an `AtomicBoolean` on `ClusterTopologyManagerRecord`:

- written by `setAutoHealEnabled(boolean, String)` — `ClusterTopologyManagerRecord.java:498`
- read by `isAutoHealEnabled()` — `ClusterTopologyManagerRecord.java:493`

`isAutoHealEnabled()` has **one** consumer in the whole tree:

```
aether/node/src/main/java/org/pragmatica/aether/api/routes/ClusterTopologyRoutes.java:120
    return ctmOpt().map(ctm -> new AutoHealStatusResponse(ctm.isAutoHealEnabled()))
```

`LeaderReconciler` — which drives replacement provisioning — contains **zero** references to it. No provisioning, reconcile, or node-lifecycle path consults the flag.

## Impact

An operator turning auto-heal off during an incident gets:

- a success response from the CLI,
- a status endpoint that agrees it is disabled,
- and auto-heal still provisioning replacements underneath.

That is worse than the knob not existing, because the status surface actively confirms the wrong thing. The most likely time to reach for this switch is a provisioning storm or a suspected over-provisioning loop — exactly when believing it worked is expensive.

## Note on the shape

This is the mirror image of the defect class documented in the 2026-08-17 handover §4: there, values were read by code that should not have trusted them; here, a value is trusted by an operator and read by nothing that matters. Both come down to a signal whose provenance nobody checked — the question worth asking of any new flag is **"who reads this, and does the answer change behaviour?"**

## Suggested direction (not prescriptive)

Either gate the replacement-provisioning decision in `LeaderReconciler` on the flag, or — if the intended semantics are narrower than "stop all auto-heal" — make the CLI and the status response say precisely what the flag does control. Silence is the only unacceptable option.

Not covered here: whether the flag should be cluster-wide consensus state rather than a per-leader `AtomicBoolean` (today a leader change loses the setting).

siy:
Closing: the flag now gates the thing it claims to gate.

## Evidence (verified at `04717dc0c`)

[...]

Issue #575, opened 2026-08-04
[operations.auto_heal] enabled is silently ignored — AutoHealSpec has zero runtime consumers and AutoHealConfig has no enabled field


siy:
## Problem

`[operations.auto_heal] enabled = false` cannot disable auto-heal. The key parses into a type nothing reads, and the type the runtime actually uses has no `enabled` field at all.

In a system with a documented over-provisioning death-spiral history, "the operator's off switch is not connected" is the wrong failure to carry into GA.

## Verified

**Two parallel, hand-duplicated types:**

| | config type | runtime type |
|---|---|---|
| | `AutoHealSpec` (`aether-config`) | `AutoHealConfig` (`environment-integration`) |
| `enabled` | **present** (`AutoHealSpec.java:7`) | **absent** — 8 components, all timings/thresholds (`AutoHealConfig.java:14-21`) |
| parsed from TOML | yes (`ClusterBootstrapConfigParser.java:536-562`) | n/a |
| read by runtime | **no** | yes (`AetherNode.java:2004,2568,2898`) |

`AutoHealSpec` has **zero production consumers outside `aether-config/src/main`**. The only reference anywhere else is a comment in `AutoHealConfig.java:30` — *"Kept in sync with AutoHealSpec."*

**They have already drifted.** Reported: `decommissioned_retention` is 24h in the spec vs 60s in the runtime default (`AutoHealConfig.DEFAULT_DECOMMISSIONED_RETENTION = timeSpan(60).seconds()`). A comment is not a synchronization mechanism.

## Consequence

An operator who sets `enabled = false` to stop replacement provisioning — say, while diagnosing a churn incident — gets a config that parses, validates, diffs cleanly, and changes nothing. Note this is distinct from #509: that concerned deficit-fill not being gated by auto-heal-off; this is auto-heal-off not existing at runtime in the first place.

## Options

1. **Collapse the duplication.** One type, read by the runtime, with `enabled` honoured. Removes the drift class permanently rather than re-syncing two records by hand.
2. Wire `AutoHealSpec.enabled` through to the provisioning decision and add a test that a disabled config provisions nothing.
3. If auto-heal genuinely cannot be disabled by design, **reject the key at validation** rather than accepting it silently — and remove it from the docs and from `cluster init` scaffolding.

(1) is the structural fix; (3) is the minimum honest position.

Found during a GA readiness sweep for present-but-inert surfaces. Related: #509, #574.

siy:
[...]

PR #1210, opened 2026-09-14
fix: integration-harness cluster-name isolation and S19 answering-node identity (#1186, #1187)

### LINKED ISSUE #1186: Integration-harness cluster names are fixed per runtime: two concurrent arms share one label and arm 1's teardown reaps arm 2
### What

`run-tests.sh` assigned the two cluster names as plain literals, and the `--runtime jvm` branch overwrote
both **unconditionally**. There was no way to give a run its own cluster name. Every `--runtime jvm` run on
cloud was therefore named `cloud-test-b-jvm`, and every container run `test-b`.

Three things are keyed on that name, and all three collide when two same-runtime arms run at once:

1. the Hetzner label `aether-cluster=<name>` that `cloud-reaper.sh` scopes on;
2. the CLI state dir `~/.aether/clusters/<name>/bootstrap-state.json`, which the harness reads node IPs from;
3. the `--cluster` argument the harness passes to its own teardown.

Consequence: **arm 1's normal, successful teardown destroys arm 2's live cluster mid-run.** The failure
presents as nodes disappearing under a running test — indistinguishable from a runtime/membership defect,
and the natural response is to investigate the product. It also silently bills for VMs that are reaped and
re-provisioned.

This blocks any concurrent cloud campaign. It is not hypothetical: the 2026-09-14 campaign plan needs two
concurrent arms, and this is the single thing that made it un-fireable.

### Where (symbols at SHA)

At `2005ea7d2`, all in `aether/tests/integration/run-tests.sh`:

- `CLUSTER_A_NAME="test-a"` and `CLUSTER_B_NAME="test-b"` — plain assignments, not `${VAR:-default}`
- the `jvm)` arm of the `case "$CLOUD_RUNTIME"` block inside `case "$ENV_TYPE" in cloud)`, which sets
  `CLUSTER_A_NAME="cloud-test-a-jvm"` / `CLUSTER_B_NAME="cloud-test-b-jvm"` unconditionally
- consumers: `bootstrap_cloud_cluster_a` / `bootstrap_cloud_cluster_b` (`aether cluster bootstrap … --cluster
  "$CLUSTER_B_NAME"`), the teardown safety-net's `cloud-reaper.sh --cluster "$CLUSTER_B_NAME" --destroy
  --force`, and `reap_cloud_cluster "$CLUSTER_A_NAME"`

Selector behaviour that makes the cross-reap total, in `tools/cloud-reaper.sh` — verified in the code, not
its comment: with `--cluster X` and without `--strict-cluster`, `SELECTORS+=("aether-cluster=${CLUSTER}")`
plus `aether-node-id`.

### Why it is separate

It is a defect in **shared tooling**, not campaign scaffolding. It predates the campaign, it affects anyone
who runs two clusters for any reason, and its symptom points at the product rather than the harness. Folding
it into a campaign branch would bury a reusable fix inside a one-off piece of work and leave the next person
to rediscover it from a confusing runtime symptom.

### Fix (implemented)

Branch `fix/cluster-name-isolation-2026-09-14`, commit **`a3edd1e29`**, based on `2005ea7d2`. 3 files,
+169/-4.

- New pure helper `resolve_cluster_name <a|b> <container|jvm>` in `aether/tests/integration/lib/cluster.sh`.
  Precedence: an explicit `CLUSTER_A_NAME`/`CLUSTER_B_NAME` from the environment always wins, **including
  over the jvm rename**; otherwise the historical defaults are returned byte for byte.
- `run-tests.sh` captures `CLUSTER_{A,B}_NAME_EXPLICIT` **before** assigning any default over them, and all
  four assignments now route through the resolver. The helper deliberately never reads
  `CLUSTER_{A,B}_NAME` itself, so it cannot mistake a default the script just set for an operator's choice.
- The resolver validates the name against the Hetzner label-value grammar and returns rc 2 on a bad one, so
  an invalid name fails **before** provisioning rather than as an opaque API error once VMs are billed.

Backward compatibility: a run that sets neither variable resolves exactly the previous names (pinned by U1/U2).

### Evidence

`bash aether/tests/integration/test/test-cluster-names.sh` → **15 passed, 0 failed**.

Green alone proves nothing, so each pin was mutation-probed against the committed base. **Four mutations,
four *different* named pins, none a superset of another:**

| Mutation | Predicted | Reddened |
|---|---|---|
| M1 reintroduce a literal jvm assignment in `run-tests.sh` | S2 | `S2 1 literal cluster-name assignment(s) reintroduced` |
| M2 delete the explicit capture (ordering bug) | S1 | `S1 capture/use ordering wrong — capture='' use='55'` |
| M3 jvm branch ignores the override (the original defect) | U3 | `U3 … expected 'arm2-a', got 'cloud-test-a-jvm'` |
| M4 drop the label validation | U7 | `U7 invalid name rejected — expected rc=2, got rc=0` (+1) |

Each mutation was verified to have **landed by content** (`git diff HEAD`, not `git diff` — a staged change
reads zero on the latter), and each restore was verified clean afterwards. An earlier probe round whose
mutations failed to apply reported `MUTATION DID NOT LAND — probe invalid` rather than a false green; that
guard is why M2-M4's first attempt was not mistaken for a passing result.

**S1 is the pin worth keeping.** It catches a capture placed *below* the defaults — which would read back the
default the script just assigned and turn every override into a silent no-op. No unit test of the pure
function can see that, because the function would still be correct.

### Known limitation — NOT fixed here, and it still blocks single-checkout concurrency

`RESULTS_JSON="${SCRIPT_DIR}/test-results.json"` is **not** per-arm. Two arms launched from the *same*
checkout still clobber each other's results file even with distinct cluster names. Concurrent arms must
therefore run from **separate clones** until that is addressed separately. Stated explicitly so nobody reads
this fix as making one checkout safe for two arms.

### LINKED ISSUE #1187: S19 quorum-loss arbitration resolves survivor IPs cache-first and never checks which node answered, so a recycled Hetzner IP makes it poll a healthy newcomer and report quorum held indefinitely
### What

`test_confirm_quorum_loss_race` decides whether the S19 quorum-loss branch was exercised by polling
`GET http://<survivor-ip>:8080/api/v1/cluster/membership` every 3s and reading `belowThreshold`.

The IP comes from `_s19_resolve_survivor_ip`, which is **cache-first and unconditional**: if a pre-kill
cache entry exists it is returned immediately, with no liveness check and no identity check. The caller
`_s19_survivor_membership_direct` then curls that IP and parses `belowThreshold` **without verifying which
node replied**.

Hetzner reuses public IPs aggressively across recreated VMs. When CTM auto-heal deletes a node and
provisions a replacement, the replacement can land on a **cached survivor's IP**. From that moment the
arbitration is reading a *healthy newcomer* in a refilled cluster, which reports `belowThreshold=false`
forever — while the harness believes it is reading a survivor that is below quorum.

The failure is silent and self-reinforcing:
- the read **succeeds**, so the "no usable membership reading" degrade path never triggers;
- `[ -n "$membership" ]` is satisfied, so it **never falls back to survivor 2** (the fallback only fires on
  an *empty* read);
- the loop therefore spends its whole 540s budget interrogating the wrong machine and concludes
  `quorum-held`, i.e. "auto-heal outran the kill window" — an honest-looking verdict for a measurement that
  never observed its subject.

This is the live-impostor variant of the corpse problem: not a dead subject returning nothing, but a
**live stranger returning something plausible**.

### Where (symbols at SHA `2005ea7d2`)

`aether/tests/integration/suites/02-chaos/test-self-drain-quorum-loss.sh`:
- `_s19_resolve_survivor_ip` — returns `_s19_cached_survivor_ip` immediately when set; live
  `cloud_public_ip` is consulted **only** when the cache misses
- `_s19_survivor_membership_direct` — `curl -sfk … "${ip}:${CLOUD_MGMT_PORT}/api/v1/cluster/membership"`,
  no identity assertion on the response
- `test_confirm_quorum_loss_race` — the 3s poll loop; its s2 fallback is gated on an **empty** `membership`,
  so a successful read from the wrong node suppresses it

### Evidence

**IP reuse is present and severe in the run this was found in** (run 3, 2026-09-14, 18 streamed journals
named `<serverId>-<ip>`):

| IP | distinct VMs |
|---|---|
| 91.98.86.221 | **6** |
| 91.98.152.142 | **6** |
| 167.233.217.34 | 3 |
| 46.225.150.249 | 2 |
| 2.28.22.255 | 1 |

Survivor s1 (`…01m2esbvb273fa99atz0fqq0d8`) was cached at **167.233.217.34** (run log: *"Pre-kill: cached
survivor … public IP (167.233.217.34)"*). That IP hosted **three** distinct VMs during the run. Survivor s2
(`hetzner-eu-core-4`, 2.28.22.255) was the only stable IP — and s2 is polled only if s1's read comes back
empty, which it would not.

**The observed signature matches this mechanism and not the "slow race" story.** Measured from the same
journals: SWIM suspicion→FAULTY p50 8s / p90 12s (n=51), replacements booting 60-110s apart. Since
`strictCoreMemberCount` excludes SUSPECT (see below), the strict count should fall to 2 within ~12s and
`belowThreshold` should go true well before any replacement joins. The arbitration nonetheless saw
`belowThreshold=true` in **0 of ~180 polls**, with no near misses — consistent with never having read the
survivor at all.

### Not proven

The mechanism is confirmed in code and its precondition (IP reuse) is confirmed in data. **What is not
proven is the timing alignment** — that s1's IP was occupied by a replacement *during* the arbitration
window specifically. The run log carries no timestamps, so the window cannot be aligned against the
journals. Confirming it is cheap and needs no cloud spend on its own: have the arbitration log the node id
it actually read.

### Suggested fix

1. **Assert identity, not reachability.** Read the node's own id from the membership response and compare
   it to the expected survivor; on mismatch, treat the read as invalid (not as `belowThreshold=false`).
2. **Re-resolve rather than trust the cache** during the poll loop, or expire the cache entry as soon as
   the cluster's VM set changes.
3. **Log the node id read on every poll.** Silence about *who answered* is what made a 540s wrong-subject
   poll indistinguishable from a lost race.

### Relation to #1171 — same shape, different system, NOT the same defect

#1171 (*sync adoption could not tell a live responder from a cold one*) is the identical **abstract shape**:
a reader that cannot distinguish which kind of subject answered it, and therefore accepts a plausible answer
from the wrong one. But it is product code (`RabiaEngine.adoptionThresholdMet` / `adoptCollectedState`),
this is test-harness shell, and they share no code path. **Not the same defect wearing two faces** — the
same failure *pattern* arising independently in two places, which is worth recording precisely because it
suggests the pattern, not either site, is the thing to design against.

### Superseded hypothesis (recorded so it is not re-formed)

An earlier hypothesis held that evicting a dead peer from `state==MEMBER` is quorum-gated, so two
sub-quorum survivors could never register the deaths — making the race structurally unwinnable.
**FALSIFIED by the code.** `MembershipFsm` gates eviction on `swimFaulty ∧ livenessGone` — both planes, but
both **local**, and explicitly *"death stays the manager's own SWIM/liveness decision"*. A single plane
moves a member MEMBER→SUSPECT, and `strictCoreMemberCount` counts *"members whose current state is exactly
MEMBER (SUSPECT excluded — strict)"*. So the strict count falls locally, with no consensus required, and
there is **no quorum gate and no product defect of that shape.**

siy:
Harness fixes found while planning the cloud campaign. **No product code is touched** — 10 files, all under
`aether/tests/integration/`, all shell. Nothing here enters the Maven reactor.

Closes #1186. Closes #1187.

## What and why

**#1186 — cluster names were fixed per runtime (`a3edd1e29`).** `CLUSTER_A_NAME`/`CLUSTER_B_NAME` were plain
assignments and the `--runtime jvm` branch overwrote both unconditionally, so two concurrent same-runtime arms
shared one cluster name, one `~/.aether/clusters/<name>/` state dir and one `aether-cluster` label — and
**arm 1's teardown reaped arm 2's live cluster.** That presents as nodes vanishing mid-run, i.e. as a runtime
fault. Now resolved through `resolve_cluster_name` (`lib/cluster.sh`); an explicit env name always wins,
including over the jvm rename, and the historical defaults are returned byte-for-byte when nothing is set.
The name is validated against the Hetzner label grammar so a bad one fails *before* provisioning.

**#1187 — the S19 race arbitration polled whoever answered (`f5a7fc5ea`).** `_s19_resolve_survivor_ip` is
cache-first and unconditional, and the membership read never checked which node replied. Hetzner recycles
public IPs hard — in the 2026-09-14 run two IPs hosted **six distinct VMs each**, and survivor s1's cached IP
hosted three — so a replacement landing on that IP answers with `belowThreshold=false` indefinitely. The read
*succeeds*, so neither the empty-read degrade path nor the survivor-2 fallback (gated on an **empty** read)
ever fires.

The identity was already in the response and simply not read: `ClusterMembershipResponse(String nodeId, …)`,
documented as *"`nodeId` is the answering survivor"*, `LOCAL` and never leader-forwarded. This needed no new
endpoint, field or product change. A mismatch is now **loud and terminal — the round is VOID, not lost**;
`rc 1` (mismatch, we provably polled a stranger) is kept distinct from `rc 2` (indeterminate, we cannot tell).

**`76a64b2be`** adds suite `02s-selfdrain` so one S19 round costs one test file instead of seven (2115s for the
full 02-chaos). The test is a **symlink** into `02-chaos`, not a copy — a copy would drift and the two suites
would disagree silently about what S19 asserts. Safe because `resolve_suite_dir` globs `"suites/<prefix>-"*`
**with the hyphen**, which is why `02y`/`02w` already coexist.

**`2354eb493`** adds the Arm D auto-heal suppression hook, **off by default**. It does not contradict the
ruling that a TOML-declared `enabled = false` is unbuildable — it cites it: PF-25 (#575) refuses that key
because the runtime never read it and an operator setting it *"gets silent no-op"*. The only real mechanism is
the imperative toggle (#603), durable in the consensus KV (#685). The S19 file is symlinked into both suites,
so the hook is gated and **fails closed**: `TRUE`, `True`, `1`, `yes`, `on` and empty all leave the stock path
untouched; only literal `true` arms it.

## Evidence

Four new test files for the harness itself — **50 assertions, 0 failures**, every one mutation-probed, each
mutation reddening a *different named* pin:

| File | Pins | Probes |
|---|---|---|
| `test-cluster-names.sh` | 15 | M1→S2 · M2→S1 · M3→U3 · M4→U7 |
| `test-s19-identity-guard.sh` | 9 | N1→I2 · N2→I6 · N3→I4+I6 |
| `test-suite-resolution.sh` | 7 | P1→R3 · P2→R5 · P3→**R0 only** |
| `test-arm-d-gate.sh` | 19 | Q1→D2+D3×5 · Q2→D6 · Q3→D7 |

Two probes are worth reading:

- **N2** (drop the `members` truncation) made the extractor return the **survivor's** id — pulled from the
  `members` array — while a **stranger** had answered, and the guard accepted. The whole defect class
  reproduced inside the fix meant to prevent it.
- **P3** (drop the hyphen from `resolve_suite_dir`) reddened `R0` **only**; the prefix-collision pin `R1`
  stayed green because `suites/02*` still returns `02-chaos` first alphabetically. A test that replicates
  production logic must pin its replica against the original, or it silently tests a stale copy.

Several pins assert **ordering**, not output — `S1` (the explicit-name capture must sit above the first
resolver call) and `D6` (the Arm D guard must sit above the first API call). A capture below its defaults, or
a guard below its first side effect, leaves the pure function perfectly correct while making the feature a
silent no-op. No unit test of the function can see that.

## What this does NOT establish

- **Every pin here is offline.** They are unit and structural tests over synthetic fixtures. The identity
  guard has never seen a live membership response, so its field-order and presence assumptions are unvalidated
  against a real cluster. If the live body differs, the guard returns INDETERMINATE and VOIDs every round —
  it fails safe, but it would produce no measurement.
- **Arm D still cannot verify its own precondition.** All three auto-heal routes are `LEADER`-scoped
  (`CLUSTER_AUTO_HEAL_STATUS(GET, …, LEADER)`), and S19 destroys the quorum that elects a leader. The arm can
  prove the flag was SET and that it SURVIVED, never that it HELD across the window. Bracketing is not
  covering, and any readout must say so.
- Base is `2005ea7d2`, which **does not build**. Acceptable only because this is shell that never enters the
  reactor. Zero upstream drift in all 10 touched files across the 47 commits to `567f08bec`, so a
  merge-forward is clean; not rebased, to keep the SHAs already cited in the sweep valid.
- `RESULTS_JSON` is still shared, so concurrent arms continue to need **separate clones**. This fix isolates
  cluster names, state dirs and reaper labels — not the results file.


siy:
## Live validation result — the identity guard MATCHED, and one residual it cannot cover

The `_s19_resolve_survivor_ip` identity guard (`f5a7fc5ea`, closing #1187) has now met a real cluster for the first time. Run: `2026-09-14-run4-02s-instrument-validation`, completed 17:25:21Z against `v1.0.0-rc4-candidate` = `2005ea7d2`.

**Result: 0 violations, 0 indeterminate**, and this is a positive result rather than an absence of complaint. The check has two-sided controls: `rc2` fires on an **empty** observed id and `rc1` on a **differing** one, so `rc0` on every poll positively proves that top-level `nodeId` was present, non-empty, and exactly equal to the addressed survivor — with `belowThreshold` parsed from that same body.

The quorum-loss branch was also entered on the first attempt (*"Race-arbitration: belowThreshold=true observed directly on a survivor — quorum-loss confirmed"*), where run 3 saw it in **0 of ~180 polls**. **This is not a rate.** One round. Run 3's 180/180 loss and this first-attempt win are *consistent with* #1187's diagnosis and measure nothing about how often the branch is reachable.

## ⚠ The residual, recorded here because it will not be visible later

**The live body serialises `nodeId` BEFORE `members`.** Therefore a naive first-match-over-whole-body extractor would have returned the **identical value** on every poll of this run.

Consequences, stated plainly:
- **This live validation cannot distinguish the correct extractor from a first-match one.** A future refactor to first-match would pass this exact validation silently.
- **Members-array truncation is validated by the N2 OFFLINE mutation only.** A members-first body has never been seen by anything.

So the guard's live evidence is **field-order-dependent**. It is strong evidence that the field is present, non-empty and correct; it is **no evidence at all** that the extractor is robust to a body whose field order changes.

**Suggested:** put that note at the extractor itself, in the same greppable form as the guarantee it scopes — a limitation a reader has to notice is one that gets skipped exactly when it matters. The N2 offline mutation is the only thing currently holding truncation, and it should say so where someone editing the extractor will read it.

## Two other findings from the same run

- **A pre-registered acceptance rule went unsatisfied and is reported as such.** The rule required QUORUM_LOSS drain **intent** in the journals; no journal watcher was running (an acknowledged omission). What exists instead is the drain's **consequence** on both survivors — `ActiveState=failed`, `ExecMainStatus=2` i.e. `Runtime.halt(2)`, reachable only via the designed drain halt, plus independent per-VM confirmation. Arguably stronger than a log line, but **not the evidence that was promised**, and it is recorded as unsatisfied rather than quietly upgraded. Journal streaming is now a precondition for future rounds.
- **#1053's JVM S19/S20 fixes are exercised for the first time** — the `systemctl show` drain read that replaced `docker inspect`, and S20 full-drain recovery to 5 healthy cores in 122s of a 600s budget.

**Teardown proved by consequence, not by the reaper's word:** independent post-check reports `servers=1, firewalls=1` — `165747131` and `11618919`, both expected to remain. Nothing leaked; nothing that had to survive died.

Evidence: `oss/internal/cloud-run-evidence/2026-09-14-run4-02s-instrument-validation/`

Issue #1253, opened 2026-09-19
OffHeapRingBufferCloseRaceTest flakes under CPU load: its non-vacuity check can see 0 entered race windows in 200 rounds


siy:
## Summary

`OffHeapRingBufferCloseRaceTest.concurrentReaders_seeNoEscapingThrowable_whenArenaClosedUnderThem` (added by #999, `478ab67ba`) fails under CPU contention. It fails on its **non-vacuity** assertion, not its safety assertion:

> `reads refused because the arena closed UNDER an in-flight reader (0 would mean these 200 rounds never entered the race window)` — expected positive, was 0.

Observed independently by two streams on 2026-09-19, both in files neither touched:
- s1246, in a full `aether-stream` run;
- s1241, in a root reactor at load average 15–21.

It passed 3/3 when re-run in isolation. CI runs a module-parallel reactor (`-T 1C` from the committed `.mvn/maven.config`), so it can redden unrelated PRs there too.

## Why it flakes (from reading the test at `ccba0dba5`)

`runOneCloseRace` starts 3 reader threads. It waits only until each has counted down `started`, which happens *before* its first read. Then it calls `close()`.

On a loaded box the readers may count down and be descheduled before entering a guarded read. Then `close()` lands with nobody in flight, so `closedUnderReaderCount()` stays 0. Across 200 rounds that can total 0. `ROUNDS = 200` was sized for an unloaded machine (see the comment at `:33-35`).

The safety assertion ("nothing escapes") is unaffected, and the non-vacuity check is doing its job: it refuses to call a run that never reached the race green. The defect is that entering the window is **probabilistic**.

## Fix

Make entering the window deterministic or adaptive. **Do not weaken the non-vacuity check.** Either option works:
- **Adaptive:** run rounds until at least K races are observed or a wall-clock budget expires (e.g. K=5, 20 s). Fail with the observed count only if the budget expires.
- **Deterministic:** add a package-private test seam in the guarded read path that parks one reader inside the access until `close()` has begun.

## Acceptance

- Green in a loop of 20 runs under an artificial CPU hog (N busy threads ≥ cores).
- Still RED if the guard is removed, i.e. the safety assertion keeps its teeth. Show the mutation.

Related: #999.

siy:
Fixed by PR #1275, squash-merged into `release-1.0.0-rc4` as `914303ac8` (session 25, 2026-09-20). Closed by the CTO session; GitHub does not auto-close on a non-default base.

Issue #1247, opened 2026-09-19
OffHeapRingBuffer reports index-corruption bugs as BUFFER_CLOSED, its try/catch comments are wrong, and JBCT lint has no try/catch rule at all


siy:
## Summary

An external review flagged "explicit try/catch remains in these modules despite the JBCT prohibition". Verification found every site in aether-stream at a genuine JDK boundary, with no business-exception catch. It found three real defects instead:

1. **`guardedAccess` masks real bugs.** It catches `IllegalStateException | IndexOutOfBoundsException` and maps **both** to `BUFFER_CLOSED`. ISE is the legitimate closed-arena race. IOOBE also covers a genuine index or offset-arithmetic bug, which is then reported as "buffer closed" with no log.
2. **The comments are false.** `OffHeapRingBuffer.java:217` and `:516` each claim to be "the single justified try/catch in this file". The file has 4.
3. **The lint cannot see the construct.** JBCT-EX-01 (`jbct/jbct-lint/.../CstNoBusinessExceptionsRule.java`) flags exception classes, `throw` and `throws`. **No rule in `jbct-lint/src/main` matches `try` or `catch`.** The `@SuppressWarnings("JBCT-EX-01")` annotations on try-only methods are therefore decorative; only `writeSynced` (which `throws IOException`) needs one. aether-stream is inside the gate (`aether/pom.xml:93` sets `jbct.skip=false`), but the gate cannot detect this construct, so a green gate is not evidence about try/catch.

References are at `ccba0dba5`. `S/` = `aether/aether-stream/src/main/java/org/pragmatica/aether/stream/`.

## Inventory

Search space: `try` followed by `{` or `(` in the 96 `*.java` files under `aether/aether-stream/src/main/java`. Positive control: the same pattern hits 16 files in `aether/node/src/main`.

- `S/OffHeapRingBuffer.java:237`, `:519`: `catch (OutOfMemoryError _)` around `Arena.allocate`. Native allocation boundary; legitimate.
- `S/OffHeapRingBuffer.java:536` (`guardedAccess`) and `:675` (`guardedRead`, new in rc4): ISE | IOOBE. Defect 1 above.
- `S/replication/ReplicationBatcher.java:128,148`: `lock()/try/finally unlock()`. Lock idiom with no catch.
- `S/wal/PartitionWal.java:544`: try-with-resources in `writeSynced`. I/O leaf.
- `StreamEntityLogSubstrate` and the projection package contain none.

[...]

PR #1275, opened 2026-09-19
fix(streams): make the close-race test enter its window every round (#1253)

siy:
Closes #1253

## Change
- `OffHeapRingBuffer`: adds a package-private test seam, `readWindowProbe(Runnable)`. `guardedRead` runs it after the `closed` fast-path check and before the native read. The default is a static no-op `Runnable` held in a plain field, and production code never sets it. The append path is not touched. The four production hunks are the constant, the field, the one call in `guardedRead`, and the setter.
- `OffHeapRingBufferCloseRaceTest.concurrentReaders_seeNoEscapingThrowable_whenArenaClosedUnderThem`: in each round, the first reader to reach the probe parks until `close()` has completed. Its native read is then refused by the JDK, so every round enters the race window regardless of scheduling. The two free-running readers still race the same `close()`.
  - The safety assertion (nothing escapes a reader) is unchanged.
  - Non-vacuity is tightened, not weakened. It was "total refusals > 0" and is now "total refusals ≥ ROUNDS" (at least one per round).
  - A new per-round check asserts that a reader actually parked. It runs after the readers are stopped, so an unwired seam fails on round 1 in about 5 s instead of timing out 200 times.

## Evidence
Load was generated by a separate JVM running 32 busy threads on the 8-core shared box, with load average 100–225 during the runs. Each iteration was one fresh JVM invoking the test method.
- **RED on the unmodified test (base `ccba0dba5`):** 4 of 40 iterations failed on the non-vacuity assertion ("Expecting actual: 0L to be greater than: 0L"). Refusals per iteration ranged from 0 to 15. With 16 hog threads, 20 of 20 passed but some iterations saw as few as 1 refusal.
- **GREEN with the fix (final HEAD):** 40 of 40 passed under the same 32-thread hog. The minimum was 200 refusals per iteration, i.e. one per round.
- **Module tests:** `aether-stream`, run as `-pl aether/aether-stream -am test -Dtest=org.pragmatica.aether.stream.**`. The fresh surefire XML holds **754 `<testcase>`** (290 files), with 0 failures, 0 errors and 0 skipped. This matches the surefire total of 754.
- **Mutation P1 (guard removed):** the `guardedRead` catch was changed to `ArithmeticException`, confirmed by content (the original catch count went from 2 to 1). The test went RED on the safety assertion: `Expecting empty but was: [java.lang.IllegalStateException: Already closed …]`, 1 of the 8 tests in the class. Restored and confirmed by content.
- **Mutation P2 (seam call removed):** RED on "a reader parked in the close-race window" after 5.3 s. Restored and confirmed by content.
- **Gate:** `mvn jbct:check -Djbct.skip=false -pl aether/aether-stream` printed `Running JBCT check on 96 Java file(s)` and `JBCT check passed.` Tests are outside the gate (`jbct.includeTests=false`).
- `scripts/changelog-check.sh`: ok (3 changed paths, 1 fragment).

[unverified: full reactor. No local root build was run, per the stream brief. CI `build-and-test` on this head is the reactor evidence.]
[unverified: CI-runner reproduction. The RED was reproduced only on the local shared box under an artificial hog.]


siy:
Merge-order note: #1247 (s1246) also edits the catch in `OffHeapRingBuffer.guardedRead`. Whichever PR merges second will need a mechanical rebase; this PR's hunk in that method is the single `readWindowProbe.run();` line added before the `try`.

Issue #1245, opened 2026-09-19
Stream publishBatch serializes every event through its own fsync and replication round-trip, and re-resolves each keyless event's partition inside its group


siy:
## Summary

**Performance.** For EVENTUAL streams, `DefaultStreamPublisher.publishBatch` groups events by partition, then publishes each group's events **one at a time**. Each event waits for a blocking WAL fsync plus a full replication acknowledgement round (up to 5 s) before the next is appended. A batch is therefore not a storage batch: it gets no group commit and no replication pipelining. `OffHeapRingBuffer.appendBatch` exists but the publisher never calls it.

**Correctness (found in verification).** The in-order chain calls `resolvePartition(event)` a second time. For keyless streams that call advances the round-robin counter again, so events are published to partitions other than the ones they were grouped under. The grouping is meaningless for keyless batches, and the "published in order within its partition" javadoc does not hold for them.

References are at `ccba0dba5`. `S/` = `aether/aether-stream/src/main/java/org/pragmatica/aether/stream/`.

## Evidence

- `S/DefaultStreamPublisher.java:232-240` `publishGroupInOrder`: `chain = chain.flatMap(_ -> publishEventual(resolvePartition(event), serializer.encode(event), timestamp))`.
- `resolvePartition` (same file): `partitionKeyExtractor.map(...).or(() -> (int)(roundRobinCounter.getAndIncrement() % partitionCount))`. This is stateful for keyless events.
- Each `publishEventual` → `publishLocal` → `durablyLog` blocks on the fsync (`S/StreamPartitionManager.java:1186-1189`, `.await()`) → `awaitReplication`.
- `matchesAwait` already treats acks cumulatively (`key.offset() <= confirmedOffset`, `S/replication/DefaultReplicationManager.java:251`), so awaiting only the last offset of a group is sound.
- Scope: app streams only. The durable-topic `Publisher<T>` exposes no batch API.

## Required fix

- Reuse the partition computed during grouping; never call `resolvePartition` twice for one event.
- Per group: one `appendBatch`, one WAL group write (see #1244 / #1232), then one `awaitReplication(lastOffset)`.
- Preserve the group's ordering guarantee; that is what the per-partition critical section in #1231 provides.

## Acceptance tests

- Keyless stream with 4 partitions: `publishBatch` of 8 events lands exactly 2 in each partition, matching the grouping. RED today.
[...]

Issue #1236, opened 2026-09-19
DurableTopicPublisher's failure contract is false: NOT_ENOUGH_REPLICAS is detected after the local append, and REPLICATION_TIMEOUT is ambiguous, not failed


siy:
## Summary

`DurableTopicPublisher` documents that a failed publish means "the event is NOT in the log". The code checks the replica count only **after** it has appended locally, fsynced the WAL and sent replication. So a publish reported as `NOT_ENOUGH_REPLICAS` has already been written, is visible to local consumers (see #1235), and may be on some peers. `REPLICATION_TIMEOUT` has the same shape and is an **outcome-unknown** result, not a failure.

This matters most because the natural caller reaction to a clean failure is to retry, and every retry mints a new message ID (see #1237). Retries therefore produce duplicates that dedup cannot collapse.

References are at `ccba0dba5`. `S/` = `aether/aether-stream/src/main/java/org/pragmatica/aether/stream/`.

## Evidence

- The contract (`S/topic/DurableTopicPublisher.java:19-21`): *"A failure means the append genuinely failed (e.g. `NOT_ENOUGH_REPLICAS` under the floor) and is actionable — the event is NOT in the log."* `aether/docs/specs/durable-pubsub-spec.md:69` (§5) says the same.
- The check (`S/replication/DefaultReplicationManager.java:155-160`): `if (targets.size() < minAcks) return NOT_ENOUGH_REPLICAS.promise();`. It sits inside `awaitReplication`.
- `awaitReplication` is reached only via `.flatMap` **after** `publishLocal` has appended to the ring, fsynced the WAL and sent replication (`S/DefaultStreamPublisher.java:257-262`). Forwarded publishes behave the same on the owner (`S/forward/StreamForwardHandler.java:143-152`).
- `minSyncReplicasFor` has only its two post-append callers. **There is no pre-append check anywhere.**
- `REPLICATION_TIMEOUT` (`DefaultReplicationManager.java:262-264`, 5 s) likewise leaves the event on the owner, in its WAL, and possibly on some peers.
- `SyncReplicationProvisioningTest.minSyncExceedsClusterPeers_failsNotEnoughReplicas` tests the await in isolation and never checks the log.

## Interaction with #411 / P8

[...]

PR #1287, opened 2026-09-19
fix(streams): publishBatch appends each partition group as one storage batch under its grouped partition (#1245)

siy:
Closes #1245

EVENTUAL batch publication previously performed a separate storage/replication operation for every event and resolved keyless partitions twice. A local partition group now uses one ordered append section, a group durability barrier, and a cumulative replication acknowledgement. Grouped events retain their original partition. This PR targets `release-1.0.0-rc4` and remains last in the reviewed merge sequence.

The batch goes through the shared owner-aware write router. It preserves the epoch fence, committed-owner admission, configured replica floor and current STRONG refusal. Remote owners are still forwarded to one event at a time, in order. Safe pre-append refusals can fall back to the single-event path; ambiguous partial appends are never automatically retried.

A failed cumulative barrier or ambiguous mid-run refusal reports every submitted event as `OutcomeUnknown`, including events already appended. The API does not provide atomic multi-event publication. The new no-WAL `SEALING_BEHIND` regression checks the retained appended prefix, unchanged durable/visible frontiers, absence of replication continuation and absence of retries.

Replication chunks account for payload bytes plus a 32-byte allowance per event against half the transport frame limit. This is an accounting estimate with chosen headroom, not a measured codec-size or throughput guarantee. A single over-budget event still goes alone. The unused manager wrapper and its fallback helpers are removed.

Validation on implementation `2526e5d1a`: all six build/JBCT steps passed; **1,216 stream tests** and **62 focused batch/write-contract tests** passed with zero failures/errors/skips. These include the independently added stale-epoch and replica-floor pins from `7adbc45b6`. Incorrectly retrying `SEALING_BEHIND` makes the new regression fail; exact source restoration passes.

The final smoke Forge gate at `5ea42caaa` (documentation-only follow-up to the implementation) passed **16 cases, zero failures/errors/skips**, in 80 seconds. All 70 runtime dependencies were fresh; no build/install ran during Forge. New-head CI remains pending.

The one-JVM send-count test demonstrates reduced replication work for its same-partition fixture. It does not establish production throughput, WAN behavior or a serialized-frame-size guarantee. Independent review and merge approval remain outstanding.


siy:
Resolved the rc4 merge conflicts and reconciled the overlapping stream changes.

Local partition groups remain one storage/WAL/replication run, while the shared write router preserves current owner admission, committed-owner redirects, live min-sync limits, and durable visibility. Results retain rc4's per-input `PublishOutcome` API. A failed cumulative barrier marks every submitted event outcome-unknown; remote and oversized-run fallback chains retain a successful prefix and stop before later events. The indexed replication waiter implementation and extracted WAL test helper are preserved.

Added the batch entry point to the existing write-path contract scenarios and pinned oversized fallback outcomes. That regression also caught and fixed preflight size validation hiding `EventTooLarge` inside a composite failure.

Validation passed: full stream suite (1,212 tests), final focused suite (60 tests), full build/JBCT checks, and live Forge smoke (16 tests; all70 runtime dependencies fresh). No cloud credentials were available to builds/tests.

siy:
**Review of merge head `0c84bfc8d` (= `ad6992950` + one merge over rc4 tip `d444d22c6`) — APPROVE-WITH-NITS, one MEDIUM to address before merge.**

Every hunk classified: `DefaultReplicationManager`/`ReplicationManager`/chunking test/`Quic*` byte-identical to the pre-rebase head; `OffHeapRingBuffer` docstring-only; compositions in `StreamWriteRouter.publishBatch` (authority routing, STRONG refusal, min-sync read once, `publishLocalBatchAtFloor(minSync-1)`, per-event fallback for `RUN_DOES_NOT_FIT`/`NotOwnerAppend`/`EventTooLarge`, all-unknown otherwise, barrier), `DefaultStreamPublisher` delegating its group to the router, `StreamPartitionManager.publishBatchInSection` mirroring `appendToPartition`'s chain (fence → admission+floor → sizes → target → section), the batch entry in the 11-scenario write contract, and the stated contract change in `BatchPublishOutcomeTest` (failed cumulative barrier → every submitted event `OutcomeUnknown`, all appended — honest per #1236, stated in the fragment).

Verified on the build host: floor/admission refuse the whole group before any append (3× `Unknown(NOT_ENOUGH_REPLICAS)`, 0 appended; `NotOwner` → per-event redirect, 0 appended); cross-group partial pinned (mutations C2 → 4 pins red, C8 → 8, C1 → 1 + probe); 500-event batch under a concurrent single writer → contiguous `[1..500]`, ring content in order; stale epoch → 3× `Unknown(StaleEpochAppend)`, 0 appended. aether-stream 1213/0/0/0, node 1617/0/0/1 (pre-existing bench), gate 100 files 0/0; consensus gate residue identical to the tip.

**M1 (address before merge)** — the epoch fence (C5) and the floor *value* (C7) are unpinned on the batch path: 59/59 PR tests stay green when either is dropped; only the reviewer's probes (`staleEpoch_fencesTheWholeRunBeforeAppend`, `floorRefusal_…`) go red. Adopt those two as tests (or equivalents) — the composition's whole point is that the batch inherits these guards, and a refactor could drop either silently today.

NITs: `StreamPartitionManager.publishLocalBatch` now has zero callers (dead); the `SEALING_BEHIND` mid-run case on a no-WAL EVENTUAL ring is reachable and handled (all-unknown, replica gap → backfill) but unpinned; the fragment's "bounded per-event framing cost" is unverified against the byte-identical `chunkEnd`.

Merge order unchanged: last in the chain. Verdict file `oss/internal/sweep-state/s25-rev1287.md`, probes `oss/internal/probes/s25-rev1287/`.

siy:
Closed the remaining three nits in `2526e5d1a`, based on independent M1 pin commit `7adbc45b6`.

- Removed unused `StreamPartitionManager.publishLocalBatch` and its two private fallback helpers. The live owner-aware router continues to call `publishLocalBatchAtFloor`.
- Qualified the framing claim in source/changelog: payload bytes plus a 32-byte allowance is an accounting estimate; the half-frame reserve is chosen headroom, not a measured serialized-size/performance guarantee. A single over-budget event still goes alone. Chunking behavior is unchanged.
- Added a real no-WAL EVENTUAL batch regression for mid-run `SEALING_BEHIND`. The first two events append before the third needs a refused eviction. Every input is reported outcome-unknown; the prefix remains at its original offsets, durable/visible frontiers do not advance, no replication continuation runs and no retry occurs. The fixture uses the existing eviction-listener boundary, with no production seam added.

The new regression fails by assertion when `SEALING_BEHIND` is incorrectly added to the safe per-event retry fallback. After exact source restoration, all 62 focused batch/write-contract tests passed. Full stream suite: **1,216 passed**, zero failures/errors/skips. Complete six-step build/JBCT passed. All commands removed `HCLOUD_TOKEN` and used an isolated checkout/Maven repository; the independent review checkout was untouched.

**Live gate:** smoke at `5ea42caaa` passed **16 cases**, zero failures/errors/skips, in **80 seconds**, with all **70 runtime dependencies fresh**. No builds or installs ran during Forge. The documentation-only follow-up `5ea42caaa` also corrects the removed-method reference and qualifies oversized-run outcomes; executable code remains `2526e5d1a`. Published as a fast-forward of `7adbc45b6`, preserving the independently added M1 tests.

Issue #1276, opened 2026-09-19
Node tests read and write the machine-global /data/aether/storage (absolute StorageConfig default): non-hermetic, 35 false failures on any host where /data is writable


siy:
## Summary

`StorageConfig.storageConfig()` (`aether/aether-config/src/main/java/org/pragmatica/aether/config/StorageConfig.java:28-37`, at `ccba0dba5`) defaults `diskPath` to the **absolute** `"/data/aether/storage"` and the snapshot path to `"/data/aether/metadata-snapshots"`. Node tests that build the default `artifacts` storage instance therefore read and write a **machine-global** directory.

- **On the laptop and on CI** `/data` isn't writable, so these tests take their "unwritable" branch and pass.
- **On any machine where `/data` IS writable**, they share persistent state across runs, trees and branches. That covers a root-run build host, a developer box with `/data`, and container images.

## Observed (2026-09-19, the session's Hetzner build host)

A root `clean install` from a stream tree had **35 `aether/node` failures** unrelated to the change under test, all with the same message:

> `Failed to create storage 'artifacts': Storage instance 'artifacts' was encrypted under key id 'key-1' but no encryption keyring is configured for it now (#831)`

The failing tests were:
- `StorageFactoryEncryptionTest` ×9
- `StorageFactoryDhtMarkerRetryTest` ×7
- `AetherNodeDhtMarkerPostFormationBootTest` ×4
- `AetherNodeSecretResolutionRefusalBootTest` ×3
- `SwimHintInstallBootTest`
- other boot and wiring tests

An earlier run with a keyring had left an encrypted marker under `/data/aether/`, and every later run without a keyring refused it.

Running Maven as non-root on the host removed the **precondition** failure: `StorageFactoryEncryptionTest` requires `/data` to be unwritable. The residue kept 6 of its 29 cases red, which shows the state is read even by unprivileged runs, because it's world-readable. At least `WalAvailabilityGateTest`, `StorageMaintenanceWiringTest` and `StorageFactoryEncryptionTest` construct the default config.

## Why this is a defect, not only a host problem

- **Tests must be hermetic.** A test outcome must not depend on files that some other run left in a global path.
- **The same default reaches operators:** any node started without an explicit `diskPath` writes `/data/aether/storage`. That may be intended for production. It is wrong for tests, and it makes "two nodes on one machine" (for example, a same-host multi-node run) share storage.

## Fix

[...]

PR #1288, opened 2026-09-19
fix(node): make aether/node storage tests hermetic instead of sharing /data/aether (#1276)

siy:
Closes #1276

## Change
- **Test helper `HermeticStorage`** (aether/node test sources) roots node-test storage in the test's own `@TempDir`. The root sits **under a regular file**, so no directory can be created beneath it on any host, root included.
  - This keeps exactly the memory+DHT degradation (`handleDiskTierUnavailable`) these tests always had on CI and on laptops, where `/data` is not writable. Behaviour on CI is unchanged; the tests just no longer depend on `/data` to get it.
  - It provides `storageConfigAt(Path, boolean)`, `synthesisDefaultsIn(Path)` and `nodeStorageIn(Path, boolean)`.
- **Boot tests through `AetherNode`** now pass an explicit `artifacts` instance from `nodeStorageIn(tempDir, …)`: AntiEntropyCycle, ContentStorageWarn (the shared `minimalConfig`) and its callers (StorageShutdown, SecretResolutionRefusal, ContentStorageWiring, ResourceFacadeSeam), DeferredStartRedrive, ScheduledTaskDrain, SwimFaultyReArm, SwimHintInstall, WorkerJoinCtm, WorkerRosterPrune, and DhtMarkerPostFormation.
  - `content` and the stream data dir derive from the artifacts disk path as siblings, so they land under the same root.
  - `encrypted` is `true` exactly when the test configures `[storage.encryption]`. That is what the synthesized instance would have been (`encrypted = keyring present`).
- **One package-private production seam** in `StorageFactory`. `createAll` always synthesizes `artifacts`/`content` from `StorageConfig.storageConfig()` when the map lacks them, so even tests with their own `@TempDir` instance reached `/data/aether`. Several `StorageFactoryEncryptionTest` cases pin the synthesis itself, so an explicit config cannot stand in for them.
  - New overload: `createAll(configs, nodeId, dhtClient, keyring, StorageConfig defaults)`.
  - The existing 4-argument signature delegates with `StorageConfig.storageConfig()`, and the 5-argument `StreamSetupRequest` overload passes it too. `AetherNode` is untouched, so production behaviour is unchanged.
  - `StorageFactoryEncryptionTest`, `StorageFactoryDhtMarkerRetryTest` and `StorageFactorySnapshotEpochTest` pass `synthesisDefaultsIn(tempDir)`.
- **`requireDefaultDiskRootUnwritable`** used to assert `/data` is not writable. It now asserts the injected root sits under the blocker file.
- **New test** `createAll_synthesizedArtifacts_takesDiskPathFromInjectedDefaults` pins the seam on disk, using a writable injected root.
- **Not changed:**
  - The production default `/data/aether/storage`, which the ticket asks to decide separately.
  - `WalAvailabilityGateTest` and `StorageMaintenanceWiringTest`. They call `StorageConfig.storageConfig()` only for walPath or size values and never resolve its disk path.

## Follow-up: EmberCluster (added on review)
- `EmberCluster.perNodeStorageConfig` used to return `Map.of()` when no `withDataBaseDir` was set, so every Ember node resolved `/data/aether/...`. It now falls back to `unwritableStorageBase()`: a path under a regular file in a temp dir that belongs to that cluster, so nothing can be created there. The dir is created lazily, only when no data dir is set, and **deleted by `stop()`**. A restart gets a fresh one; it never holds state.
- **Behaviour change, stated plainly:** an Ember cluster without `withDataBaseDir` now ALWAYS runs its nodes without the disk tier and without the stream WAL. Storage is memory + DHT, and streaming is not crash-durable.
  - Before, that was true only where `/data` was not writable, which covers CI and laptops.
  - Where `/data` WAS writable, those nodes wrote to it, and every cluster on the machine shared one `/data/aether` across runs, trees and branches.
  - Clusters that call `withDataBaseDir`, as Forge does, are unchanged: they keep a writable disk tier and WAL.
- If the temp dir cannot be created, the cluster falls back to the old empty map.
- `EmberClusterHermeticStorageTest` does not depend on any host's `/data`. It reads the config the cluster hands its nodes and never starts one.
  - `perNodeStorageConfig_rootsEachNodeUnderTheClustersUncreatableTempDir_neverUnderData` checks that `artifacts` is present, sits under the blocker file, and is not under `/data`.
  - `stop_deletesTheClustersStorageTempDir` checks the cleanup on `stop()`.
- The try/catch in `createUnwritableStorageBaseDir` and `deleteUnwritableStorageBaseDir` is not marked with JBCT-EX-03, because #1294 is not merged. It matches the neighbouring `createParticipationMarkerBaseDir`.

## Evidence
All runs used the **build host** (remote-mvn). Its `/data` is writable and holds the foreign `key-1` encryption marker, which makes it exactly the acceptance host. Counts are `<testcase>` elements in surefire XML newer than a run marker; the attribute totals agree.

| Run | Tree / SHA | aether/node testcases | storage failures | classes whose output mentions `/data/aether` |
|---|---|---|---|---|
| RED | base `ccba0dba5` | 1493 | **32** (+1 unrelated error) | **15** |
| GREEN | `18546e64e` | 1493 | **0** (+1 unrelated error) | **0** |
| Control, re-run of base after GREEN | `ccba0dba5` | 1493 | **32** | **15** |
| Final, `clean test` | `63a0f1fc1` (head) | **1494**, 0 failures, 0 errors, 1 skipped | **0** | **0** |

- **Storage failures:** every storage failure is "encrypted under key id 'key-1' but no encryption keyring is configured", or a SecretResolutionRefusal assertion that got that storage refusal instead of the secret refusal.
- **Control:** the base re-run after GREEN shows the host residue was still present while GREEN ran.
- **Unrelated error:** `CoreSwimHealthDetectorTest$ObservationListenerPreRegistration` (ClassCastException `SwimHealthState$Stopped`) errored in 2 of 4 runs, at base and at the fix alike. It is a flake, not a storage failure.
- **Skip:** the 1 skip is the pre-existing class-level `@Disabled` `ObservabilityStep0BenchTest.step0_overhead_bench`, present at base too.
- **Mutation probe:** in the `defaults` overload I replaced `defaults` with `StorageConfig.storageConfig()` and confirmed by content. The 3 StorageFactory classes then went **15 of 38 RED**, including the new seam test. Restored and confirmed by content. The final `clean test` run above, at head after the restore, is green. The same 38 tests were also 38/38 green immediately before the probe.
- **Gate:** `jbct:check -Djbct.skip=false -pl aether/node` printed `Running JBCT check on 148 Java file(s)` and `JBCT check passed.`
- `scripts/changelog-check.sh`: ok (19 paths, 1 fragment).

### Evidence for the EmberCluster follow-up
All runs are on the build host, `-pl aether/ember`, and the head runs used `clean`.

| Run | SHA | testcases | errors | classes whose output mentions `/data/aether` |
|---|---|---|---|---|
| RED | base `ccba0dba5` | 27 (17 files) | **9** (all `key id 'key-1'`) | **7** |
| Control, base re-run after the first GREEN | `ccba0dba5` | 27 | **9** | **7** |
| GREEN, `clean test` | head `d9946d1b4` | **29 (18 files), 0 failures / 0 errors / 0 skipped** | 0 | 0 |

- The fresh XML contains both new test names.
- Mutation P1 removed `.orElse(this::unwritableStorageBase)`. `perNodeStorageConfig_…` went RED, 1 of 2 in the class.
- Mutation P2 removed `releaseUnwritableStorageBase()` from `clearClusterState`. `stop_deletesTheClustersStorageTempDir` went RED ("stop() must delete the cluster's storage temp dir").
  - Both probes ran with `clean`, were restored, and were confirmed by content. A clean run after the restore gave 29/29.
  - The P1 red does not depend on the host: the reverted code returns an empty map on any machine.
- Gate: `jbct:check -Djbct.skip=false -pl aether/ember` printed `Running JBCT check on 6 Java file(s)` and `JBCT check passed.` An earlier round had a formatting violation in my hunk (reviewer finding 1); `jbct format` fixed it.

### Summary of the review numbers
- aether/node on the contaminated build host went from **32 failures** to **1494 testcases, 0 failures, 0 errors**.
- aether/ember went from **9 errors** to **0**; head has 29 testcases.

[unverified: the "`/data` unwritable" acceptance leg. CI `build-and-test` on this head is that evidence, and no local root build was run.]
[unverified: how the seam test fails on a host with unwritable `/data`. On the build host the probe reddened it through the key-1 refusal. On a clean unwritable host the reasoning is that the synthesized disk degrades and no block file appears under the injected root.]
[unverified: forge-tests (`-Pwith-e2e` failsafe). This change covers them, because they build clusters through `EmberCluster`, but I did not run them on the build host. CI `forge-tests` on this head is the evidence.]


siy:
Follow-up pushed: `2239308cc` roots `EmberCluster` node storage at a temp dir that cannot be created when no data dir is set. The directory sits under a regular file, created once per cluster. Before this, those nodes got `Map.of()` and resolved `/data/aether/...`.

Evidence below comes from the build host, whose `/data` is writable and still carries the foreign `key-1` marker. Each run is `-pl aether/ember test`; counts are `<testcase>` elements in XML newer than a run marker.

| Run | SHA | testcases | errors | classes whose output mentions `/data/aether` |
|---|---|---|---|---|
| RED | base `ccba0dba5` | 27 (17 files) | **9** (all `key id 'key-1'`) | **7** |
| GREEN | `47fe6ccb9`, pre-format | 27 | 0 | 0 |
| Control, base re-run after GREEN | `ccba0dba5` | 27 | **9** | **7** |
| GREEN | `2239308cc` (head) | **27, 0 failures / 0 errors / 0 skipped** | 0 | 0 |

- Gate: `jbct:check -Djbct.skip=false -pl aether/ember` printed `Running JBCT check on 6 Java file(s)` and `JBCT check passed.`
  - The first check at `47fe6ccb9` flagged my hunk for formatting. `jbct format` fixed it, and the fix was amended into `2239308cc`; that change is whitespace only.
- Behaviour when there is no data dir is unchanged in kind: disk tiers and the WAL stay off, as they were on CI.
- If the temp dir cannot be created, the cluster falls back to the old `Map.of()`.

siy:
Review nit addressed in `1963f2ffd`: `clearClusterStateOnFailure` now releases the Ember storage temp dir as well. That clear is shared by `abortStart` and `handleStartResults`, so a failed start no longer leaks the dir, even if the caller never calls `stop()`.

- The new test `abortStart_deletesTheClustersStorageTempDir` is in `EmberClusterHermeticStorageTest`.
- Mutation probe P3: with the release removed from `clearClusterStateOnFailure`, that test went RED, 1 of 3 in the class ("a failed start must delete the cluster's storage temp dir"). The run used `clean`. The mutation was restored and confirmed by content.
- Build host `clean test -pl aether/ember` after the restore: **30 testcases in 18 XML files, 0 failures / 0 errors / 0 skipped**, and 0 classes whose output mentions `/data/aether`.
- Gate: `jbct:check -Djbct.skip=false -pl aether/ember` printed `Running JBCT check on 6 Java file(s)` and `JBCT check passed.`
- The changelog fragment now says the dir is deleted "on `stop()` or after a failed start".

Issue #1246, opened 2026-09-19
ReplicationBatcher scans every partition accumulator ever created every 1 ms and never evicts them; fix before #263 wires it


siy:
## Summary

`ReplicationBatcher` runs `flushAll` every 1 ms. That flush takes a `ReentrantLock` on **every accumulator ever created**, including empty and idle ones, and accumulators are never removed. The cost is O(partitions ever owned) lock acquisitions per millisecond, growing for the life of the node. This code is **dead in production today** because #263 has not wired the batcher. It becomes live the moment #263 lands, so it should be fixed first.

References are at `ccba0dba5`. `S/` = `aether/aether-stream/src/main/java/org/pragmatica/aether/stream/`.

## Evidence

- `S/replication/ReplicationBatcher.java:43`: `scheduleAtFixedRate(this::flushAll, 1ms)`. `:80`: `accumulators.forEach(this::flushPartition)`. `:146-151` locks each entry, even when it is empty.
- The only map operations are `computeIfAbsent` and `forEach`. There is no removal.
- It is not wired: `AetherNode.java:3657` uses the non-batching manager, and `batchingReplicationManager` has no main-source caller.

## Related measurement note (no defect)

Cursor checkpoints cost at most about 2 consensus writes/s per **active** (group, partition): the trigger is 1000 events or 500 ms for durable groups (`DurableGroupIdentity.java:28`), and `ClusterCursorStore.java:34` says one consensus round per checkpoint. Idle groups cost nothing. This is not yet measured at scale; include it in the benchmark acceptance for #263.

## Required fix

Either evict an accumulator when a drain finds it empty and on partition release, or replace the global scan with a per-partition one-shot flush that `add` schedules.

## Acceptance test

With 10k idle partitions, one flush tick performs zero lock acquisitions, and the accumulator count returns to 0 after the partitions are released.

## Provenance

P5 of an external source review (at `0143cdf79`). Confirmed at `ccba0dba5` by source reading, with the correction that the batcher is not currently wired.

Related: #263 (should depend on this).

siy:
Fixed by PR #1293, squash-merged into `release-1.0.0-rc4` as `582515465` (session 25, 2026-09-20). Closed by the CTO session; GitHub does not auto-close on a non-default base.

Issue #263, opened 2026-06-11
Wire ReplicationBatcher: per-event replication caps owner throughput ~25-50K events/s


siy:
## Finding (streaming deep assessment 2026-06-11, perf lever #1)

`AetherNode.java:2036-2038` constructs `ReplicationManager.replicationManager(self, registry, transport)` — the no-batcher path → `replicateImmediately` sends one `ReplicateEvents` message per event per replica (`DefaultReplicationManager.java:59-63, 84-114`). `ReplicationBatcher` (100 events / 1 ms window, `ReplicationBatcher.java:23-24`) exists and is tested but unwired.

Per-event replication bounds owner throughput at ~25–50K events/s (2× QUIC sends + ack handling per event at RF=2) versus multi-million events/s ring capability — a 20–100× gap for what is nearly a one-line wiring change.

## Acceptance
- Batcher wired in node assembly; batch path covered by the S1 offset-verification semantics (#260 — batches make offset fidelity MORE important, land together or after).
- A minimal data-plane throughput benchmark recording before/after (nothing load-tests streaming today — k6 targets app HTTP at 200 rps; suite 04 publishes a few hundred events via curl).

Part of #165.

siy:
Two constraints from #1293 (#1246, replication batcher) that this wiring must accept, from its adversarial review:
- **Latency:** the batcher no longer scans every 1 ms. The first event of a batch schedules a one-shot flush at `maxDelay`. The upper bound is unchanged (`maxDelay`), but a lone event now always waits the full `maxDelay` instead of about half of it on average. Size `maxDelay` for the mean, not only the bound.
- **Ordering:** successive batches for one partition are sent without ordering between them (the same as before #1293). The receiver rejects a gap and backfills, so correctness holds, but out-of-order arrival costs a backfill round trip. If this path becomes hot, send per-partition batches serially.

PR #1293, opened 2026-09-19
fix(streams): ReplicationBatcher per-batch one-shot flush, evict drained accumulators (#1246)

siy:
Closes #1246

## What was wrong
`ReplicationBatcher` ran `flushAll` on a 1 ms `scheduleAtFixedRate` timer. Each tick took the `ReentrantLock` of **every accumulator ever created**, idle ones included, and nothing ever removed an accumulator. The cost was O(partitions ever owned) lock acquisitions per millisecond, growing for the life of the node. The batcher is not wired in production yet (#263), so no running node paid this cost.

## Fix
- **No periodic scan.** The first event of a batch schedules a one-shot flush for that batch at `maxDelay` (`SharedScheduler.schedule`).
- **Single-use accumulators.** The first `drain` retires an accumulator, and the batcher evicts it (`accumulators.remove(key, accumulator)`). An accumulator therefore exists only while its partition has a pending batch. Idle or released partitions hold no map entry and cost no lock acquisition.
- **Race handling.** An `add` that finds the accumulator retired (`AddOutcome.RETIRED`) evicts it (idempotent) and retries on a fresh one, so eviction loses no event. A later drain of an already-retired accumulator returns `EMPTY`. This covers a batch's one-shot firing after a size flush already took the batch, and it never resends.
- **Latency.** The bound is unchanged: every event is flushed within `maxDelay` of its `add`. Mean latency for a sparse single event rises from about `maxDelay/2` to `maxDelay`. `[mechanism: one-shot scheduled at maxDelay on the batch's first event]`
- **No public API change.** There is no partition-release hook: the ticket's "count returns to 0 after release" holds because a released partition stops receiving events, and its last batch is drained and evicted within `maxDelay`.
- The decorative `@SuppressWarnings("JBCT-EX-01")` annotations on `BatchAccumulator.add`/`drain` (`try`/`finally`, no `throw`/`catch`) are removed here, not in #1247, because this PR rewrites those methods (#1247 names them).

## Evidence
**RED → GREEN.** The tests were committed first (`82a80aea2`). On unmodified code `ReplicationBatcherTest$AccumulatorEviction` fails 3/3: accumulator count `10000` / `10000` / `1`, expected `0`. After the fix, `ReplicationBatcherTest,ReplicationManagerTest` ran 32 testcases with 0 failures. The concurrent test also caught a double-send in my first fix version (43477 delivered vs 20000: a second drain returned the batch again) before that version was committed.

Tests (`ReplicationBatcherTest$AccumulatorEviction`):
- `flushAll_tenThousandPartitionsGoneIdle_accumulatorCountReturnsToZero`: 10k partitions go idle, and after one flush the map is empty, so the next flush makes zero lock acquisitions (flushAll locks exactly the map's entries).
- `scheduledFlush_tenThousandPartitionsGoneIdle_evictsEveryAccumulator`: the one-shot timer path evicts all 10k without `flushAll`.
- `add_concurrentWithSizeAndTimerFlushes_losesNoEvent`: 4 threads × 5,000 adds with a 1 ms delay and `maxEvents=7`; the delivery count is exact and the final count is 0.

**Mutation probes.** Each probe was applied after commit, with the revert verified by grepping the changed string:
| Probe | Result |
|---|---|
| P1: production file back to base (the seam kept) | 3/3 RED (10000/10000/1) |
| P2: drop the `alreadyDrained` guard | concurrent test RED: 41926 ≠ 20000 (resend) |
| P3: `RETIRED -> {}` (no retry) | concurrent test RED: 15952 ≠ 20000 (lost events). Detection is race-dependent; it went red on this run |
| P4: no eviction in `flushPartition` | 3/3 RED |

**Module / blast radius (remote-mvn, `-pl` only; root reactors are refused on the host, #1276).**
- `clean install -pl aether/aether-stream -am` at `7423f6a66`: 31 modules, BUILD SUCCESS, **4811 `<testcase>` elements, 0 failures, 0 skipped** (counted from the copied-back XML; this matches the surefire totals). aether-stream alone is 757/0.
- `compile -pl aether/aether-stream -amd`: 11 dependents (incl. aether/node, forge) compile. The public API is unchanged.
- An earlier remote ROOT `clean install` (`7423f6a66`) gave 145 reactor lines, 140 SUCCESS including Aether Stream. Aether Node had 35 failures, all **build-host contamination per #1276** (writable `/data` with a stale "encrypted under key id 'key-1'" marker; the same error appears in two other trees' logs), so the 4 modules after it were SKIPPED. **Reactor: CI build-and-test — pending.** A red in aether/node on CI would be real.

**Gate:** `mvn jbct:check -Djbct.skip=false -pl aether/aether-stream` printed `Running JBCT check on 96 Java file(s)` and `JBCT check passed.`

**Pre-existing flake, not this change (untouched file):** in one local full aether-stream run under box load, `OffHeapRingBufferCloseRaceTest.concurrentReaders_seeNoEscapingThrowable_whenArenaClosedUnderThem` failed with "reads refused because the arena closed UNDER an in-flight reader (0 would mean these 200 rounds never entered the race window) — Expecting actual: 0L to be greater than: 0L". It passed 3/3 in isolation and 757/757 in the remote run. It is load-sensitive, and the CTO is tracking it separately.

**Note for #263:** when it wires the batcher, it accepts the latency trade above: the bound stays ≤ `maxDelay`, but the mean for a sparse single event rises to `maxDelay`.

## Not verified
- `[unverified: behaviour under the production wiring — the batcher is still unwired (#263); all evidence is unit level]`
- `[unverified: scheduler cost of one SharedScheduler.schedule per batch per active partition at production rates — not benchmarked; belongs to #263's benchmark acceptance]`


siy:
Review nits N1/N2 (rev1293) applied in `8dcc0d649`:
- N1: the class javadoc now states the trade: a lone event waits the full `maxDelay` (mean up from about `maxDelay/2`); the bound is unchanged.
- N2: `ReplicationBatcherTest$TimeBasedFlush.scheduledFlush_loneEvent_sentWithinMaxDelayBound` pins the bound (`maxDelay`=200 ms; the send must land in [150, 600] ms). Green: remote `clean install -pl aether/aether-stream -am` gave 758 aether-stream testcases, 0 failing files; the new test took 0.201 s. **Mutation:** a one-shot at 5×`maxDelay` turns it RED (`Expecting actual: 1000L to be between: [150L, 600L]`, 14 run, 1 failure). Restored and verified by content count (0).
- Gate: `jbct:check -pl aether/aether-stream`: `Running JBCT check on 96 Java file(s)`, `JBCT check passed.`

siy:
Review N3–N6 (rev1293) applied, head `904349e19`. Tests committed first (`32b33c103`), then the fix (`904349e19`).

- **N6 (regression vs base):** `close()` now sets a closed flag, cancels every pending one-shot and drains what was accepted. It is idempotent. A later `add` is refused with the new `ReplicationError.Lifecycle.BATCHER_CLOSED` (`Cause.Terminal`). **`add` now returns `Result<Unit>`** instead of `void`, and its only caller (`DefaultReplicationManager.replicateEvent`, fire-and-forget) logs a refusal at WARN and drops it (FER, stated at the site). An add racing `close()` is never stranded: close drains it, or its own one-shot flushes it (class javadoc).
  - `AfterClose.addAfterClose_sendsNothing` was **RED at 8dcc0d649's code**: "Expecting empty but was: [SentMessage[… ReplicateEvents …]]".
  - `addAfterClose_isRefusedWithBatcherClosed` and `close_cancelsPendingOneShot_andDrainsAcceptedEvents` observe cancellation through a package-private `FlushScheduler` seam. They need the new API, so they could not be RED at the old head except as compile failures; their pinning is shown by probes N6a/N6b below.
- **N3:** `DeliveryIdentity.add_concurrentWithSizeAndTimerFlushes_deliversEveryIdExactlyOnce` runs 4 threads × 5,000 events with unique ids over 10 rounds, and asserts the delivered multiset equals the added set exactly.
- **N4:** the comment at the retry path states why the eviction is load-bearing (a drainer descheduled between `drain()` and `remove` → unbounded recursion). It is **pinned deterministically**: `RetryPath.add_onRetiredButNotEvictedAccumulator_evictsItAndAppendsToFreshOne` drains the live accumulator directly (retired, not evicted) through the package-private `accumulatorFor` seam, then adds.
- **N5:** no change (recorded on #263).

**Mutation probes** (remote `clean` runs, each verified by content count, restored to 0 dirty):
| Probe | RED |
|---|---|
| N6a: no closed check | `addAfterClose_isRefusedWithBatcherClosed`, `addAfterClose_sendsNothing` |
| N6b: no `cancelFlush` in close | `close_cancelsPendingOneShot_andDrainsAcceptedEvents` |
| N4: no eviction in retry | `RetryPath…` **StackOverflow**, plus the identity test |
| N3-resend: `alreadyDrained` guard removed | identity test + count test |
| N3-drop: RETIRED dropped | identity test + count test + `RetryPath` |

**Build:** remote `clean install -pl aether/aether-stream -am` gave **763 aether-stream testcases, 0 failing** (XML newer than the run marker; every new test name present). Gate: `Running JBCT check on 96 Java file(s)`, `JBCT check passed.`

PR #1294, opened 2026-09-19
fix(streams): ring out-of-bounds is RingIndexCorrupted, not BUFFER_CLOSED; JBCT-EX-03 catch rule (#1247)

siy:
Closes #1247

## Ring: an out-of-bounds access is no longer reported as a close
`OffHeapRingBuffer.guardedAccess`/`guardedRead` caught `IllegalStateException | IndexOutOfBoundsException` and handled both as the closed-arena race. `IllegalStateException` keeps that handling (`BUFFER_CLOSED` / `closedUnderReaderCount`). `IndexOutOfBoundsException` is an index or offset-arithmetic defect, and it now:
- logs at ERROR with stream, partition and the exception (`reportIndexCorruption`);
- increments the new `indexCorruptionCount()`, the sibling of `closedUnderReaderCount()`;
- on the `Result` paths, returns the new `StreamError.RingIndexCorrupted(streamName, partition, detail)`. The `long`/sweep paths have no failure channel, so they still return the refusal sentinel, but they are counted as corruption, not as a close race.

The two "single justified try/catch" comments (`buildFloorGuarded`, `allocateGuarded`) now name all four marked sites. The hunks stay inside `guardedAccess`, `guardedRead`, those two comments/annotations, plus one field, one accessor and one report method; the append path is untouched.

## Lint: JBCT-EX-03, a `catch` outside a marked JDK boundary
- `CstTryCatchRule` flags every `catch` clause unless the enclosing declaration carries `@SuppressWarnings("JBCT-EX-03")`, the explicit JDK-boundary mark. It reuses the existing rule-ID suppression mechanism, so there is no new annotation or API. A `JBCT-EX-01` suppression does **not** exempt a `catch`. `try`/`finally` and try-with-resources without a `catch` are not flagged, and test classes are exempt (as for JBCT-PAT-03). `@Contract` still suppresses every rule, as it does today.
- Registered in `CstLinter`, `LintConfig`, `RuleCategoryMapping` (EXCEPTION_HYGIENE), `RuleFixtures`.
- **Severity WARNING, not ERROR — reported to the CTO per the stream brief.** At introduction, over every `src/main/java` in the repo (131 dirs), the rule hits **330 `catch` clauses in 50 modules**, **131 in 26 `aether/` modules** (all jbct-gated) and 199 in 24 others. aether-stream is at 0. Raise to ERROR when the count is zero.
- The ring's four boundary methods move from the decorative `JBCT-EX-01` to `JBCT-EX-03`. The `BatchAccumulator` decorative suppressions are removed in #1246's PR, which rewrites those methods (this PR does not touch `ReplicationBatcher`, to avoid a conflict). `PartitionWal.writeSynced` keeps `JBCT-EX-01`, which is real there (`throws IOException`), and has no `catch`.

## Evidence
**RED → GREEN.** The tests were committed before the fix (`c4fdcbd4c` ring, `ed68684b4` lint; the ring commit carries a scaffold — the `RingIndexCorrupted` record and an `indexCorruptionCount()` that returned 0 — so the RED is behavioural, not a compile error).
- Ring (`OffHeapRingBufferIndexCorruptionTest`, which corrupts slot 0's data position to −1000 in a real ring): on unmodified code it fails 3/3. `read_…` and `readSlice_…` got `BUFFER_CLOSED`; `evictByAge_…` got `indexCorruptionCount` 0 (the fault was counted as a close race). GREEN after the fix.
- Lint: on the unregistered rule, 6 failures (`CstTryCatchRuleTest` ×4, `RuleFixtureCoverageTest` positive JBCT-EX-03, `RuleRegistryInvariantTest.everyFixture_targetsLiveRule` "targets rule IDs no rule registers: [JBCT-EX-03]"). GREEN after registration: jbct-lint 797, jbct-core 153.

**Mutation probes (ring).** Each was applied after commit, with the revert verified by content grep. The two probes redden disjoint test sets:
| Probe | RED |
|---|---|
| P1: `guardedAccess` back to `catch (IllegalStateException \| IndexOutOfBoundsException _) -> BUFFER_CLOSED` | `read_corruptedIndexEntry_…`, `readSlice_corruptedIndexEntry_…` (2) |
| P2: `guardedRead` back to the combined catch counting a close race | `evictByAge_corruptedIndexEntry_…` (1) |
The lint rule's control is the RED run itself (rule present but unregistered: all positives silent).

**Lint positive/negative controls on the real gate.** With the plugin rebuilt from this branch:
- `jbct:check -pl aether/aether-metrics`: `Running JBCT check on 48 Java file(s)`, **7 × `[JBCT-EX-03]` WARNING**, `JBCT check passed.` (the gate now sees the construct).
- The base-branch `OffHeapRingBuffer` (EX-01 marks) linted by the CLI gives exactly 4 hits at lines 247/521/538/677, the ticket's four sites. The marked version gives 0.
- `jbct:check -pl aether/aether-stream`: `Running JBCT check on 96 Java file(s)`, `JBCT check passed.`, and zero EX-03.

**Module / blast radius (remote-mvn, `-pl` only; root reactors are refused on the host, #1276).** `clean install -pl jbct/jbct-core,jbct/jbct-lint,jbct/jbct-format,jbct/jbct-cli,jbct/jbct-init,jbct/jbct-derive,jbct/jbct-maven-plugin,aether/aether-stream -am` at `e89d58e3b`: 40 modules, BUILD SUCCESS, **6127 `<testcase>` elements, 0 failures, 3 skipped** (the pre-existing JBCT Formatter skips; the XML count matches the surefire totals). This covers every Maven dependent of jbct-lint/jbct-core. **Reactor: CI build-and-test — pending.** It exercises the lifecycle `jbct:process` modules, where EX-03 is WARNING and every `jbct.toml` has `failOnWarning = false`.

## Not verified
- `[unverified: the 330/131 corpus counts come from the jbct CLI (built at this branch) run per src/main/java dir with the root jbct.toml; the Maven gate was run on aether-stream and aether-metrics only]`
- `[unverified: rule behaviour on catch clauses inside lambdas/initializers outside any method — reported with method '(unknown)'; no corpus site inspected]`


siy:
Review rulings (rev1294) applied, head `55c998ae5`. Tests committed first (`ac9992440`) and shown RED at the previous head, then fixed (`6d64948f2` M1, `87d9963b1` M2).

**M1 — the JBCT-EX-03 mark is method-scoped.** `CstTryCatchRule` opts out of the linter's generic line-range suppression (new `CstLintRule.usesScopedSuppression()`, default true) and accepts only annotations on the catch's nearest enclosing method or constructor: `@SuppressWarnings("JBCT-EX-03"|"all")` or `@Contract`. A mark on a class, field, local variable or parameter exempts nothing. A marked method does not cover anonymous- or nested-class methods inside it. A lambda belongs to its enclosing method.
- RED at the old head: 8 of `CstTryCatchRuleTest$MethodScopedMark` (the reviewer's r09/r11/r12/r13/r15, plus class-level `@Contract`, field mark and a local-class method). GREEN now: 21 cases.
- Probe Q1 (generic suppression restored) → those 8 RED. Probe Q2 (any member counts, not only methods) → `fieldMark_exemptsNothing` RED.
- **Re-measured corpus:** **352 catch clauses in 52 main-source modules, 151 in 28 `aether/` modules**, up from 330/131. The +22 are catches previously hidden by class-level marks (aether/cli +14 via class-level `@Contract`, pg-maven-plugin +5, jbct-maven-plugin +2, ember +1). aether-stream stays at 0. Controls: an unmarked fixture gives 1 hit, the base ring gives 4.
- r14 (`@Contract` on the method) stays exempt, since `@Contract` is documented to exempt every rule. r16 is out of scope (reviewer N3).

**M2 — the corrupted-ring cause reaches callers after a segment fallback.** `PartitionedStreamAccess.combineWithBufferEvents` now returns `Result`. `bufferReadFallback` propagates `RingIndexCorrupted`; other buffer failures keep the short-read degrade (FER, stated at the site).
- `SegmentFallbackTest$CorruptedRingAfterFallback.fetch_mixedRange_corruptedRing_failsWithRingIndexCorrupted_notTruncated` was RED at the old head ("expected RingIndexCorrupted, got 5 events") and is GREEN now. Probe Q5 (truncating again) → RED.

**Nits.**
- N1: `read_corruptedIndexEntry_logsOneErrorNamingStreamAndPartition` captures exactly one ERROR event naming `corrupt[3]` and carrying the `IndexOutOfBoundsException`. Probe Q4 (ERROR→WARN) → RED.
- N2: `defaultSeverity_isWarning`. Probe Q3 (LintConfig EX-03 → ERROR) → RED (`expected: <WARNING> but was: <ERROR>`).

Every probe ran as a remote `clean` run, was verified by content count, and was restored (0 dirty).

**Build.** Remote `clean install -pl jbct/{core,lint,format,cli,init,derive,maven-plugin},aether/aether-stream -am` at `55c998ae5`: 40 modules, **6142 testcases, 0 failures, 3 pre-existing skips** (XML newer than the run marker). Each new test name is present in the fresh XML. Gate: `jbct:check -pl aether/aether-stream` → `Running JBCT check on 96 Java file(s)`, `JBCT check passed.` Both changelog fragments are updated. Reactor: CI build-and-test on this head.

Issue #1249, opened 2026-09-19
DurableTopicDeliveryForgeTest is @Disabled ("never observed fully green") with no tracking issue; convert it to an enabled tripwire and make it pass


siy:
## Summary

The composed end-to-end test for durable-topic delivery is disabled. Its own annotation says it has never been observed fully green. No issue tracks enabling it, so the one test that exercises the composed durable pub-sub path gates nothing.

References are at `ccba0dba5`.

## Evidence

`aether/forge/forge-tests/src/test/java/org/pragmatica/aether/forge/DurableTopicDeliveryForgeTest.java:96-99`:

> `@Disabled("never observed fully green; enable on first green run — run 2 executed all five arms … three arms failed on test-side baseline carryover and the group-isolation arm was vacuous. Enabling it IS the acceptance criterion.")`

It is also `@Tag("Heavy")`, and unchanged since `0143cdf79`.

## Required fix

1. Fix the per-arm baseline carryover, which is a test-side defect, and make the group-isolation arm non-vacuous. Prove it can fail: mutate the isolation and watch it go red.
2. Enable the test. Until it can pass, replace `@Disabled` with an **enabled** tripwire that asserts the current known-failing behaviour and fails loudly, with instructions to delete it and enable the real arms, once that behaviour changes. A disabled test guarantees nothing.
3. Several defects filed from the same review sit on this path: #1238 (consumer push path) and #1235/#1236 (publish visibility and outcomes). The arms should be designed so that each of those reddens a named arm.

## Acceptance

Enabled, and green in 3 consecutive Heavy runs, **and** each arm shown to go red under a mutation of the behaviour it covers.

## Provenance

P7 of an external source review (at `0143cdf79`). Confirmed at `ccba0dba5` by source reading.

Related: #386, #739 (missing owner-SIGKILL arm), #751.

siy:
Fixed by #1341, squashed into release-1.0.0-rc4 as a5f649483: DurableTopicDeliveryForgeTest is enabled again with the pre-attach arm rebuilt as an honest tripwire (in-flight sampling, peer-ack predicate, owner-assignee predicate, move detection) plus repair-restart recovery; review s25-rev1341 (eight rounds) APPROVE. Merges into a release branch do not auto-close — closed by hand.

Issue #1109, opened 2026-09-13
Leader-issued node commands can be applied on a follower before the committed KV put the leader acted on has been applied there: LOAD/ACTIVATE (#1083 B1) and the ACTIVATE→stream-bindings read share the race


siy:
## What

The leader observes a committed KV put (e.g. `SliceTargetKey`), then issues a node command (LOAD / ACTIVATE) over the cluster-sync channel. That command is not itself ordered by Rabia behind the put, so a follower can APPLY the command a few milliseconds before its own KV apply has reached the put — the command's precondition read on the follower then misses. Measured in CI at PR #1083's first head (`27f744066`): leader observes the put at `.788`, follower applies LOAD at `.789`, follower's first observation of the put at `.793` — a 4–5 ms window on a loaded runner (a `SLOW-APPLY` of 10 s with queueDepth 33 preceded it).

Enumerated at `c275960b6` (every `ctx.kvStore()` read in `NodeDeploymentState`, 11 sites — `oss-internal verify-1083-r2-2026-09-14` §S6):

| command / path | key read on the node | race | recovery |
|---|---|---|---|
| LOAD, ACTIVATE, ACTIVE-claim redeploy | `SliceTargetKey`, `VersionRoutingKey` (`committedTargetPermits`) | **yes** (B1) | PR #1083 parks and re-drives on the put; the deferral is lost across a quorum cycle (S4, being fixed there) |
| ACTIVATE → stream binding (`lookupOwningBlueprintId` / `lookupBindingsFor` → `BlueprintStreamBindingsKey`, written in an EARLIER batch than the target) | bindings key | **yes** — same family | FAILED + RETRY → leader retry budget (3); bounded, but a put landing after the last retry **rolls the blueprint back** |
| ROUTING epoch acks (`collectTargetNodesForArtifact`, `seedEpochAckExpectationsFromKvStore`) | other nodes' keys | in principle | `ROUTING` 30 s timeout → remediated |
| UNLOAD, DEACTIVATE, DRAIN/LEAVING, `existingPausedFlag` | own prior value / none | no | — |

## Why it is one ticket

[...]

Issue #739, opened 2026-08-29
Durable-topic e2e is missing the owner-SIGKILL arm — the one that actually distinguishes durable from ephemeral


siy:
## The missing arm

`DurableTopicDeliveryForgeTest` (#386, landed `31428b961`, `@Disabled` until its first green run) covers delivery, ordering, bounded retries into the DLQ, and group attribution. It does **NOT** cover survival of an owner loss, and that is the arm which most directly separates the durable tier from the ephemeral one.

The fixture is already shaped for it: both topics are `partitions = 1, replicas = 2, min_sync_replicas = 2`. That scoping exists precisely so a SIGKILL of the partition owner is survivable by a second in-sync copy — **at RF=1 the arm would be vacuous by construction**, because the ack never depended on a second copy in the first place.

## Why it was not written

The arm needs the HRW owner of the topic's BACKING stream identified before the kill, then re-identified after it. That is the in-JVM replica-set introspection `AbstractStreamOwnerFailover` uses (`ownerView()` / `hasCaughtUpNonOwner` / `ownerChangedAndAuthoritative`), tied to the derived `topic:<address>` stream name (`DurableTopicNames.topicStream`, namespace auto-derived from the blueprint's Maven coordinates).

The alternative — kill an arbitrary node and assert events still arrive — was deliberately NOT written. On a 5-node cluster with one partition, an arbitrary kill usually hits a node that owns nothing relevant, so the assertion passes without ever exercising failover. **That is a test whose name claims more than it observes**, and this session has spent considerable effort deleting exactly that kind of test rather than shipping one.

Flagged as a ticket rather than a TODO comment so the e2e's incomplete acceptance lives somewhere it will be triaged.

## Shape of the work

1. Discover the backing stream (`topic:<address>`) — the existing suite already does this from the runtime's stream listing rather than reconstructing the address.
2. Wire the replica-set view for that stream; wait for owner + a CAUGHT_UP non-owner before publishing, so the min-sync barrier has a replica to await and the kill has somewhere to fail over to.
3. Publish, confirm delivery, `cluster.killNode(owner)`.
4. Assert the partition re-resolves to a NEW owner, and that events published before the kill are still delivered — and that publishing continues to work afterwards.

## Acceptance

- The arm identifies the real HRW owner before killing it (never an arbitrary node).
- It fails if the pre-kill replica set has no caught-up non-owner, rather than proceeding into a vacuous assertion.
[...]

Issue #1117, opened 2026-09-14
Stream-consumer slice instances stall in ACTIVATING/ROUTING ~90s until forceActivatingToActive while the leader already reports the blueprint fully deployed


siy:
**What.** In the #1066 Heavy runs (`DeclarativeStreamConsumerTest`, branch `fix/1066-body-publish-stream-bindings` @ `178814261`), two of five instances of the declarative-consumer slice stayed `ACTIVATING` / `ROUTING` for ~90 s after the slice was loaded and serving locally, until `NodeDeploymentState$Active.forceActivatingToActive` fired on the activation timeout ("remaining chain steps are publication and confirmation"). Meanwhile `ClusterDeploymentState$Active.trackBlueprintSliceActive` had logged `Blueprint forge.test:declarative-consumer:1.0.0 fully deployed — all 1 slices active` 90 s earlier.

**Why it matters.** "Fully deployed" on the leader and "ACTIVATING" on two nodes are contradictory operator surfaces for the same blueprint; anything gating on either (a test's readiness wait, an operator's rollout check, the stream-consumer assignment that reads `DeploymentMap.byArtifact` ACTIVE) sees a different cluster depending on which it reads. The stall is also the window in which #1119-style temporary assignees are taken.

**Where.** `NodeDeploymentState$Active.forceActivatingToActive`, `ClusterDeploymentState$Active.trackBlueprintSliceActive` at `178814261` (symbols, not lines — they drift).

**Evidence.** fix-1071-r2 round 3: `runs/r6-3-…/mvn.log` 03:22:05 vs 03:23:35; `probe/slices.log` in r5-1 shows `dsc-1 ACTIVATING, dsc-2 ROUTING` for the whole window. Written up in `oss/internal/fix-1066-r2-report-2026-09-14.md` (T1). Not reproduced deterministically; observed in 1 of 7 local runs and in CI run 34791307280.

**Not in scope of #1066/#1071** — deployment activation chain, not stream bindings. The #1071 fixture now waits for the owner-attached state and prints per-node assignment rows on failure, so a recurrence fails setUp with the nodes named rather than as a count mismatch.

siy:
[...]

PR #1341, opened 2026-09-19
test(forge): verify durable-topic delivery and repair restart recovery

siy:
Closes #1249. Enables all six DurableTopicDeliveryForgeTest arms and makes each verdict attributable to the events and consumer instances involved. Also fixes the durable-history setup and shutdown race exposed by the Heavy Forge slow-rejoin gate.

## Delivery contracts

- Each arm publishes its own event IDs and checks per-ID delivery/retry counts. Fixture instance IDs prevent forwarded HTTP requests from counting one instance twice.
- The fixture is pinned to five ACTIVE instances on distinct nodes. The group-isolation arm checks both the failing group's exact retry budget and the healthy group's handling of its own probe.
- Delayed acknowledgements make the ordering arm distinguish serial dispatch from overlapping delivery.
- The pre-attach arm identifies a specific readable, peer-acknowledged warm-up event before attachment, then requires its delivery without another append. It claims the subscribe-time kick only when the assigned consumer uses the owner's push path; unsuitable placement produces a named skip with readings.
- Exactly-once arms record assignee distribution before and after their observation. Consumer movement produces a named skip, reflecting the documented at-least-once replay contract (#1385), rather than an unsupported exactly-once claim. #739 owns the deterministic owner-crash successor.

Historical mutation checks established that removing the subscribe-time drain makes the backlog arm fail, and removing single-flight drain makes the ordering arm fail. Those are mechanism evidence from the reviewed earlier heads, not final-head validation.

## Restart correctness

Ember offers opt-in per-node consensus persistence. The slow-rejoin fixture retains real history and participation markers across restart, commits and observes an explicit history marker before stopping, and preserves its existing no-replacement hold, same-identity rejoin and 5→7 scaling assertions.

The node checkpoint adapter decodes the Git persistence phase envelope before Base64 decoding. Consensus shutdown rejects new submissions and drains accepted application work before saving the final checkpoint, preventing shutdown from recording a phase behind an in-flight application. A latch-controlled regression exercises that exact interleaving. No quorum, participation marker, assertion or timeout is relaxed.

## Validation and limits

`DurableTopicDeliveryForgeTest` is Heavy; ordinary `forge-tests` success does not execute these arms. The earlier local slow-rejoin run passed both cases in 153 seconds, but subsequent Heavy CI at c5a0bd3e0 remained red: 73 cases, 2 failures, 2 errors, 1 skip. It completed Maven normally; the 40-minute job duration was not a timeout cancellation. The observed stop/apply overlap motivated the new shutdown fix.

The same CI run found an ownership benchmark racing initial ownership establishment and a stream deployment failing on missing visible bindings before rollback. These remain separate findings; targeted restart validation cannot establish whole-suite success. See [validation details](https://github.com/pragmaticalabs/pragmatica/blob/fix/1249-durable-topic-forge-test/aether/docs/contributors/pr1341-heavy-forge-validation.md), #1398 for the repeated restart acceptance request, and #1109 for the unresolved deployment visibility failure and #1406 for the unresolved ownership benchmark fixture failure.

**Final validation at `2efc9e4e6` (merged rc4 `d444d22c6`):** all six build/JBCT steps passed; the unchanged consensus code passed 799 cases and the shutdown mutation failed as expected, followed by two restored focused passes; merged stream consumer tests passed 68 cases. Three freshness-checked live slow-rejoin runs each passed two cases with zero failures/errors/skips, **153 / 153 / 153 seconds** gate wall time. Each checked 70 fresh runtime dependencies, zero replacement creates, same five returning identities and exactly two positive-control creates. No builds or installs ran during Forge. A subsequent commit records this evidence only; new-head CI and independent review remain outstanding.


siy:
Fixed the heavy Forge failure in `c5a0bd3e0` without extending timeouts, resetting participation markers or weakening quorum assertions.

The slow-rejoin fixture retained its PARTICIPATED marker across restart but recreated in-memory consensus history. Once the fixture used real per-node persistence, a second defect became visible: the backup writer's `# Phase:` envelope was passed directly into the Base64 snapshot decoder, silently making the saved snapshot unavailable. The adapter now handles that envelope, with a round-trip/invalid-payload regression.

Validation:
- Affected 88-module reactor install passed.
- Node/Ember JBCT checks passed; two snapshot encoding regressions passed.
- Freshness-checked `PostRestartSlowRejoinDeficitFillProbeTest`: **2 passed**, zero failures/errors, 153 seconds. The unchanged 82.8-second hold provisioned no replacements; the same five identities rejoined; the 5→7 positive control provisioned exactly two nodes.

Fresh CI has been triggered. Original user/other-agent worktrees were not modified.

siy:
The latest Heavy CI failure was a real stop/application race, not a job-timeout cancellation. A consensus application was still inside a ten-second callback when shutdown checkpointed and reset the engine. One retained node then had phase-zero history while its two live peers restored phase one, so the recovery quorum correctly refused to count that node's history.

`36544f5c6` closes admission when stopping begins and queues final checkpoint/reset behind accepted application work. A latch-controlled regression proves stop cannot checkpoint or finish until the application returns; it also checks idempotent stop, late submission refusal and accepted-request completion. Replacing queued stop with the former synchronous behavior makes the regression fail; restoring it passes.

The fixture now commits and observes an explicit applied-history marker before restart. It retains real persistence, participation markers, the unchanged hold window, same-identity rejoin and positive scaling assertions. No timeout or quorum was weakened.

Merged current rc4 `d444d22c6` into validation head `2efc9e4e6`. Validation:

- Complete six-step build/JBCT passed.
- Consensus: 799 cases, zero failures/errors/skips before the rc4 merge (that merge changes only stream consumer code/tests and its changelog); focused restored shutdown cases: two passed.
- Merged StreamConsumerRuntimeTest: 68 passed, zero failures/errors/skips.
- Three freshness-checked live slow-rejoin repetitions: **2 passed each**, zero failures/errors/skips; gate wall times **153 / 153 / 153 seconds**. Each checked all 70 runtime dependencies fresh, observed zero creates through the unchanged hold/rejoin and exactly two positive-control creates. Same head/private Maven repository throughout; no builds or installs during live probes.

The earlier local green run remains valid historical evidence, but it did not resolve every Heavy CI failure. The missing-binding deployment failure remains tracked in [#1109](https://github.com/pragmaticalabs/pragmatica/issues/1109#issuecomment-5766344446), and the ownership benchmark's missing readiness gate/swallowed refusal diagnostics in [#1406](https://github.com/pragmaticalabs/pragmatica/issues/1406). Neither is claimed fixed by this restart patch. Full new-head Heavy CI and independent review remain necessary.

siy:
The latest Heavy run at `db7bd8c00` passed both restart cases and all six durable-topic cases. Its remaining failures are independent: the ownership measurement again reports 1,999/2,000 successes (#1406), and the projection loses LIVE model continuity before any rebuild request (#1400, with readiness evidence under #1117).

The [runner analysis and links](https://github.com/pragmaticalabs/pragmatica/issues/1398#issuecomment-5767513230) distinguish observed evidence from inferred causes. No timeout increase or speculative production change was made. The Heavy suite remains red; the passing restart tests do not make the entire run green.

siy:
The newer [Heavy run 35656226488](https://github.com/pragmaticalabs/pragmatica/actions/runs/35656226488), on independently updated head `d6dcc562c`, has a different remaining failure: **65 cases, zero failures, one error, one skip**.

`DeclarativeStreamConsumerTest.setUp:185` reaches its four-minute co-location readiness deadline. Node 18104 still reports `spread-events[2]` assignee/owner `dsc-3`/`dsc-1`, while the other four views report `dsc-1`/`dsc-1`. This is observed assignment-view divergence; the log does not establish why it persisted.

Ownership baseline (2/2), projection rebuild (1/1), and restart (2/2) pass in this run. Durable-topic coverage is five passed plus the named pre-attach-backlog skip. Therefore the current red run must not be attributed to the previous #1406/#1400 failures. Earlier evidence remains valid for its stated head/run. Tracking the new setup failure under #1398; no timeout relaxation or speculative production fix is justified by this trace.

Issue #1345, opened 2026-09-19
Restart after WAL compaction but before the metadata snapshot silently reassigns offsets: offset 0 carries a later event's payload


siy:
Found by vfy-seal-entity while reviewing #1297 (probe C, `Rev1297CrashProbeTest`). Pre-existing, reproduced unchanged across all four heads of that PR.

Sequence: the WAL is compacted, the node restarts before the metadata snapshot is written. Measured: pre-crash watermark 195; after restart **head = 3, and offset 0 carries event 196's payload**. So a reader asking for offset 0 gets a different event's data, and the log's offsets no longer mean what they meant before the restart.

This is silent corruption on the recovery path, not a refusal: nothing fails, nothing logs, and every consumer, replica and entity fold built on those offsets inherits it. Same family as #1340 (a ring read racing an eviction returning another offset's record), but on restart rather than under concurrency.

Next: lift the probe from the review into a committed test, decide whether recovery must refuse (no snapshot covering the compacted range → fail loudly) or reconstruct the offsets from the WAL, and pin it. Do NOT ship a recovery path that renumbers events.

siy:
Premise correction (s1345, session 25, verified at `StorageFactory.java:65-66`): the streams metadata store snapshots at **100 mutations / 30 s**, not 1000 / 60 s — those are other setups' defaults. The mechanism and the fix are unchanged. Fix is PR #1349 (converged with #1258's head-gap acceptance per know `5cbf54ef1`): WAL truncation is bounded by the sealed watermark rebuilt from the latest on-disk snapshot; replay refuses only mid-log holes/duplicates; a leading gap is accepted at stored offsets with a WARN and counter. The no-fsync snapshot write is #1353.

siy:
Fixed by #1349, squashed into `release-1.0.0-rc4` as `816dc16e1` (2026-09-20, session 25; adversarial review verdict in `oss/internal/sweep-state/s25-rev1349*.md`). Closed manually — merges into a non-default branch do not auto-close.

Issue #1311, opened 2026-09-19
core Promise: an Error thrown inside a callback is swallowed, leaving dependent state wedged with no signal


siy:
Seen twice in session 24 reviews:
- #1297: a StackOverflowError inside a SegmentSealer promise callback left `inFlight` set forever. Sealing stopped with no log and no failure count (s1264's repro, scratchpad `s1264/sealer-wedge-evidence.log`).
- #1258: a listener Error left the notifier flag set, so the ring stopped notifying for good (rev1258 probe H).

Each PR is being fixed locally, with finally-style release and error guards. The common root is the core promise library: an `Error`, as opposed to an exception, thrown from a continuation is not surfaced. The promise neither fails nor logs, so any flag set before the callback is never cleared.

Investigate `Promise`'s callback execution. Decide whether a `Throwable` escaping a continuation should (a) fail the dependent promise, (b) be logged at ERROR with the continuation's origin, and (c) rethrow `VirtualMachineError` after logging. Pin the chosen behaviour with a test in which a continuation throws `AssertionError` / `StackOverflowError` and the failure is observable. [unverified: the exact swallow site in Promise; trace it before fixing.]

siy:
Session 25 status. **Phase 1 (core)** is in PR from stream s1311 (branch `fix/1311-promise-throwable-containment`): a `Throwable` escaping any `Promise` continuation fails the dependent promise with a cause carrying the origin, is logged at ERROR, and `VirtualMachineError` is rethrown after logging; the swallow site was `AsyncExecutor.runAsync` discarding the `Future` (`Promise.java:3207` at `ccba0dba5`), reached from every handler/map/fold path and from the already-resolved synchronous paths. Ruled in addition: all 20 synchronous lift sites (`Result.lift` ×2, `Unit.lift`, `Tuple.lift` ×15, `Option.lift`, `VirtualThreadScheduler` ×2) rethrow `VirtualMachineError` through ONE shared helper (`Causes.rethrowIfFatal`), pinned by a parameterised test over the lift family; the one named exception is `VirtualThreadScheduler.timerLoop` (single platform timer thread — rethrowing there would silently stop every scheduled task in the JVM), which keeps WARN-and-continue. Evidence run is a full-reactor `clean install`.

[...]

Issue #1278, opened 2026-09-19
WAL recovery cannot distinguish reclaimed history from a lost WAL head: persist a retention reclaimed-through floor


siy:
## Summary

WAL recovery cannot tell **legitimately reclaimed history** from **a lost WAL head**. Both look like "the first WAL record starts above the sealed offset".

**Reproduced by the #1258 adversarial review** (probe `oss/internal/sweep-state/s24-rev1258-probe.java.txt`):
1. Retention deletes every sealed segment, so `SegmentIndex.lastSealedOffset` drops to -1.
2. WAL compaction has already removed the records those segments covered.
3. The WAL holds a perfectly contiguous 8001..8999, and #1258's strict recovery check refused it at boot (`expected 0 … holds 8001 (gap)`).

**CTO ruling for #1258:** a gap *before* the first record is accepted, with the records placed at their stored offsets and a loud WARN plus a counter. Gaps between records and duplicates still refuse.

That keeps normal operation from bricking a partition, but it also means **a genuinely lost WAL head is now accepted with a warning** instead of refused. The two cases need distinguishing evidence, and today none exists.

## Required

**A single retention authority.** This is #634 item 4: no joint `wal-floor ≥ seal-floor ≥ checkpoint-floor` invariant exists. It persists a durable **reclaimed-through** floor per partition *before* deleting segments or compacting the WAL. Recovery then does one of two things:
- **accepts** a head gap only when `firstRecordOffset ≤ reclaimedThrough + 1` (silently or at INFO);
- **refuses loudly** otherwise, because the head was lost, not reclaimed.

Related: #1234's contiguous sealed watermark (`sealedThrough`, rebuilt via `anchorAtLowestRef`) has the same blind spot when every ref is deleted.

## Acceptance

- Retention reclaims every segment, then restart: boots with no warning, and the recorded floor explains the start.
- A WAL whose head was truncated without a recorded floor: boot refuses loudly and names the missing range.
- The floor is written before the deletion it licenses. Crash between the two: the next boot still accepts.

Related: #634, #1231/#1232 (PR #1258), #1234, #1240.

PR #1362, opened 2026-09-20
fix(core): a Throwable escaping a Promise continuation fails the dependent promise and is logged (#1311)

siy:
Closes #1345

**Base `release-1.0.0-rc4`, merged up to `2185c8edc`** (#1297 squash, #1258, #1294, #1290). Merge commits: `24a8672ee` (#1297's head), `2cf46c38e` (rc4 `0db85d618`), `2d434d5a2` (rc4 `2185c8edc`, #1297's squash — squash side taken wherever it matched, #1345 kept where it differs). The diff against rc4 is now #1345's own 15 files.

## Defect
Segment refs reach disk only through the streams metadata snapshot (`StorageFactory`: 100 mutations / 30 s — the ticket's 1000 / 60 s are the other setups' defaults), while `truncateWalsToSealed` ran off the in-memory `SegmentIndex` and `PartitionWal` compacts past 8 MiB. Restart in that window: refs lost, watermark rebuilt lower than the compaction point, survivors appended at fresh offsets. Silent.

## Fix — (a) as ruled; (b) converged with #1258 (corrected ruling after rev1349 F1)
- **(a)** `StreamPartitionManager` takes a `DurableSealedOffsetSource` (new) and the truncation tick bounds by it: the sealed watermarks a restart would REBUILD. `AetherNode` supplies `fromLatestSnapshot(streams.snapshotManager())` — the latest snapshot file on disk, parsed the way boot parses it (`SegmentIndex.rebuildFromRefs(Map)`), read once per tick. No snapshot → nothing truncated. Truncation lags the snapshot cadence by ≤ 30 s. The production (9-arg) factory demands the durable source; test/standalone factories treat their single source as durable.
- **(b)** Recovery is #1258's: a leading gap is accepted as reclaimed history — survivors at their STORED offsets, WARN, `walRecoveryHeadGapsAccepted` — so nothing is ever renumbered; a hole or duplicate inside the tail is `WalReplayMismatch`. #1345 adds `requireContiguous` in `placeTail`: the whole tail is checked against the seed **before any record is appended**, so a refused recovery hands the sink nothing (#1258's `placeRecord` refuses at the bad record, after appending the prefix — the prefix can evict and seal). `WalRecoveryGap` from earlier rounds is gone; one cause type.
- **F3** Held-back reclamation is visible from the tick (on-disk bound below live and unchanged since the previous tick; WARN from the 2nd consecutive tick then every 10, with partitions + WAL bytes; `walReclamationHeldBackTicks()`), and on the operator surface: `GET /storage/retention` → `walReclamationHeldBackTicks` beside #1258's `walRecoveryHeadGapsAccepted` (`RetentionRoutes` reads it from the same manager whose WAL snapshot it renders; `management-api.md`, `cli.md`; pinned by `RetentionRoutesTest$WalTotals.assembleRetention_managerOverload_reportsWalReclamationHeldBackTicks`).
- **A1** (rev1349 delta): the node shape on a refused partition is documented from its producers — ERROR per ATTEMPT; attempts are boot hydration, once on the replica-registration edge (`ReplicaSetController.reconcilePartition` → `onBecameReplica`), and every publish; reads never materialize; nothing periodic. `StreamPartitionManager.hydrationFailed` now carries the cause instead of narrating an allocation failure. `durableAtPreviousTick` is cleared on stream deletion and close.

## #1278
A partition whose every ref retention reclaimed rebuilds base `-1` (#1278, open) and is indistinguishable from lost refs. It is ACCEPTED with the head-gap WARN on every restart, survivors at stored offsets (the ruled behaviour: never renumber; no refusal of a legitimate state). Pinned by the ENABLED `restartAfterRetentionReclaimedEveryRef_acceptedAsReclaimedHistory_warnedUntil1278`, whose message says a persisted reclaimed-through floor (#1278) is what makes the floor durable so the WARN stops — when #1278 lands, its WARN/counter assertions go red and are dropped. #1278's fix is not attempted here.

## Evidence (all counts from builds of this branch; head `2d434d5a2` = rc4 `2185c8edc` + #1345)
- **RED at base `0a244edc3`** (`e1cc848b1`): `[head=3 tail=0 count=4] expected: 199L but was: 3L` / `offset0 payloadIndex=196`. Also red at rc4 `ccba0dba5` (own scratch worktree).
- **GREEN** `aether/aether-stream` full module at `2d434d5a2`: **1,005 `<testcase>`** (marker-fresh XML), 0 failures, 0 errors, 0 skipped (973 at `07bee1da0` before rc4 `2185c8edc`). `aether/node` full module at `07bee1da0`: **1,521**, 0 failures, 0 errors, 1 skipped (`ObservabilityStep0BenchTest`, pre-existing); at `2d434d5a2`: see the report. `WalTruncationDurableBoundBootTest` (real `AetherNode` boot over a temp dir) 1/1.
- **Mutation table on the converged code** (`oss/internal/probes/s25-1345/mutate.sh`, `runs/*-converged.log`; restores by content, `git diff HEAD` empty):

  | | Mutation | Scope | Red (named) |
  |---|---|---|---|
  | M1 | truncation bound = live index | module, 973 | `…refsNotYetSnapshotted_recoversOriginalOffsets` (`[WAL still starts at offset 0: nothing was compacted] expected: 0L`), `…countsHeldBackTicks…` |
  | C1 | pre-append `requireContiguous` removed (per-record check remains) | module, 973 | `restart_walWithMidLogHole_refusesLoudly_andSealsNothing` — `[a refused recovery handed the sink segments [0-0, 1-1]]` |
  | C2 | head-gap acceptance removed (`seed = base`) | module, 973 | `…survivorsKeepStoredOffsets_headGapWarned`, `…acceptedAsReclaimedHistory_warnedUntil1278`, #1258's `rebuild_acceptsLeadingGap_*` ×2 |
  | M3 | `fromLatestSnapshot` reports nothing durable | class + boot | `fromLatestSnapshot_reportsOnlyRefsOnDisk`; boot test "snapshot on disk: the tick compacts" |
  | M4 | refs-map rebuild without lowest-ref anchoring | module | `SegmentIndexTest$ContiguousSealedWatermark.rebuildFromRefs_anchorsAtLowestSurvivingRef_andStopsAtHole` (pre-merge run) |
  | M5 / M6 | `AetherNode` / production factory hand the live index as the durable bound | boot | `expected: 14340800L but was: 1434080L` |
  | M8 / M9 | held-back never counted / never WARNed | class | `…countsHeldBackTicks_andWarnsAfterGrace` |

- **Gates** (direct `jbct:check -Djbct.skip=false`, plugin rebuilt from rc4 so #1247's rules apply) at `2d434d5a2`: `aether/aether-stream` **98 files** passed (427 warnings), `aether/node` **150 files** passed (1,618).

## Limits
- `[unverified: power loss]` — `DefaultSnapshotManager` writes without fsync, `LATEST` overwritten in place; "on disk" is process-crash-durable. Separate finding, CTO to ticket.
- Empty compacted WAL against a lower watermark is accepted vacuously `[unverified: unreachable via the ring today]` (#1278's floor closes it).


siy:
rev1362 fixes pushed as `6ecacd0f4` (one commit on top of `2b1b1e986`):

- **F1** — the already-resolved sync `fold`/`replaceResult` guards now fail + log + **rethrow** a `VirtualMachineError`, so `Promise.success(1).map(overflow)` throws exactly as the same mapper attached before `succeed()` does. Pins: P1a/P1b adopted.
- **F2** — a VME rethrown by one completion no longer abandons its CAS-claimed batch: `runAll` catches per completion, finishes actions then joins, and the first error is rethrown after the joins; `runEventHandlers` and `resolve()`'s drain loop do the same. Pins: P2/P2b/P7 adopted.
- **F3** — origin = first frame not under `java.`/`jdk.`/`sun.`/`org.pragmatica.lang.` (fallback `trace[0]`). Pin: P4 (`Integer.parseInt`). Consequence: `PromiseContinuationEscapeTest` moved to `org.pragmatica.usercode` — every core test lives under the skipped prefix, so a pin asserting "the origin is my frame" has to be caller-shaped.
- Class doc + changelog corrected accordingly; decision 3 (timeout-scheduler path) written as a decision, verified by P3/P3b.
- **Evidence at `6ecacd0f4`:** pin classes 23 + 3 + 20 = 46/46; full `core` 928 `<testcase>` (= 882 base + 46), 0 failures, 0 skipped; format diff-of-diffs clean on all touched files; lint 10 → 9 vs base (two pre-existing RET-06 on the replaced loops became one on `runAll`). Mutation rows M13 (F2 reverted) → P2/P2b/P7 red, M14/M14b (F1 reverted) → P1b/P1a red, M15 (F3 reverted) → P4 red; M12 (helper no-op) → 30 red incl. all 20 family rows.
- **F4** (`ConsumerRuntimeState.guardedCycle`) and **F5** (`QuicClusterNetwork.dialResolved` NPE surfaced by the guard, 7× in the reactor log) are recorded in the report for ticketing; not in this PR.

Full reactor re-runs once on the final head after the reviewer's re-check.

PR #1379, opened 2026-09-20
feat(cluster): add atomic leader-guarded mutations and canonical snapshots

siy:
Guarded KV transactions validate the committed leader and complete read/write set before applying any mutation. Every guarded operation, including a one-key update, reports acceptance or refusal using a caller-chosen transaction ID. Canonical KV snapshots explicitly use framework/generated Map/Set canonical encoding.

Current target remains rc4 under the owner's revised release decision; this batch may merge before rc5 when reviewed and ready. Current rc4 was merged without rewriting history. No merge or approval has been performed.

## Response to review

1. **Observable refusal:** removed `LeaderPut`; use `LeaderTransaction` for single-key writes too. Tag 123 is retired, not reused. `KVStoreLeaderTransactionTest:108` pins stale-leader rejection with the caller's ID. `TransactionResult` documents that a refusal identifies the transaction, not the conflicting key (`KVCommand.java:38`).
2. **Legacy unmarked mutations:** `LeaderAuthorized.java:3` explicitly documents that plain Put/Remove is silently dropped without notification and its previous-value return is not acceptance. `KVStoreNotificationIsolationTest:130` pins this behavior. New callers must use correlated transactions.
3. **Complete read set:** `KVStoreLeaderTransactionTest:91` changes a second read-only guard while written keys remain unchanged; refusal changes no written key and emits no notification.
4. **Remove/reinsert ownership bypass:** `KVStore.newerOwnerWitness` requires a strictly higher epoch (`KVStore.java:373`). `KVStoreOwnerFenceTest:88` proves the incumbent's equal-epoch witness cannot clear the fence and transfer authority to a contender.
5. **Coherent reads and costs:** `KVStoreNotificationIsolationTest:142` blocks a batch halfway through and proves a snapshot cannot see partial state. Reads synchronize with application; snapshot/forEach/find copy the full store under that monitor, with O(n) cost. Notifications release the storage monitor, but other writers still wait on the mutation lock while callbacks execute.
6. **Snapshot selection and reentrancy:** `KVStoreCanonicalSnapshotTest:21` exercises actual makeSnapshot with colliding keys and reversed insertion. `KVStoreNotificationIsolationTest:113` requires both nested applications to finish before queued callbacks execute.
7. **Claim precision:** changelog documents individual mutation/transaction ownership fencing, deletion rules, writer blocking and canonical scope. Canonical encoding covers Map/Set through framework/generated codecs; custom serializers and hand-written formats are not converted. `Serializer.canonical()` explicitly documents its default `this` behavior. The sequential claim test was renamed. The inherited formatter indentation nit is not represented as a correctness change.

Test filenames above are under `integrations/cluster/src/test/java/org/pragmatica/cluster/state/kvstore/`; production KV files are under the matching `src/main/java` package. All seven requested changes and applicable nits are implemented.

## Projection integration

rc4 `0002e2194` adds strict epoch minting for projection rewinds. The merged predicate rejects a lower epoch, or an equal epoch that either mints authority or changes owner. Same-owner non-minting refreshes remain valid. `KVStoreOwnerFenceTest.mintRequiresSuccessorEpoch_evenForTheSameOwner` pins both rules together, and the seventh mutation removes the mint condition to prove that pin fails.

## Reproducible validation

Current head `9f8a20fcd`, integrated with rc4 `0002e2194`, is ready for independent review; its [full repository build/unit, Forge and stub CI](https://github.com/pragmaticalabs/pragmatica/actions/runs/35553422578) passes. [Seven-mutation acceptance](https://github.com/pragmaticalabs/pragmatica/actions/runs/35553422589), fragment and stub checks pass.

The downloaded `foundation-mutation-evidence` artifact records tested merge `1015d2738c1be808da2c24a7f6d8a4015657c25a`; GitHub confirms parents rc4 `0002e2194f273cb4761483cf728471c7b6115bf7` and PR head `9f8a20fcd5f55ce802fc641e55eca05837dc9076`. All seven mutations were caught by assertion failures: ignored read guards, missing epoch-mint strictness, equal-epoch deletion, noncanonical snapshots, removed dispatch guard, unsynchronized snapshots, and ignored leader authority. Restored production code passed **93 actual KV cases with zero failures/errors/skips**. The artifact contains commands and JUnit reports.

Reproduce after installing the matching dependencies:

```sh
python3 tools/check-hierarchy-foundation-mutations.py
env -u HCLOUD_TOKEN mvn -T1 -pl integrations/cluster test -Dtest='KVStore*Test'
```

The mutation runner removes `HCLOUD_TOKEN` from child environments and restores source bytes after each probe. Seven named mechanism-removal probes are claimed here, not the reviewer's separate sixteen-probe campaign. Independent adversarial review remains required. Review/merge order: #1386 → #1379 → #1390. Nothing has been approved or merged.


siy:
## Review — request changes

**Target: rc5.** Keep the base on rc4 for CI until `release-1.0.0-rc5` exists, then retarget; this batch does not land on rc4. The primitives are the foundation every authority decision in #1390 rests on, so "opt-in and green" is the floor, not the bar: every guarantee the body claims must be pinned by a test that goes red when the mechanism is removed, and every refusal must be visible to the operation that submitted it.

Full verdict, 16 named mutations and the consistency-lens table: `oss/internal/sweep-state/s25-rev1379.md` (head `f6d85e7cc`; merged clean onto rc4 `c7c977e1a`, 76-module reactor green, node 1566/0; all runs local).

### What held under attack
Multi-key atomicity (a failing second guard changes no key), the committed-leader check evaluated inside the replicated apply against `LeaderKey` (every replica decides identically — the right place for the fence), `OwnerFenced` on a single `Put`/transaction, notification dispatch off the storage monitor (readers < 200 ms under a 2 s-blocked listener; ordering held under nested reentrant puts and 32 racing batches), canonical bytes under colliding-hash maps/sets and reversed insertion, plain-serializer bytes byte-identical across base/head/owner jars, tag baseline adds exactly 123–127 and the tripwire reddens when reverted. Counts: cluster 136→150, serialization 40→42, node 1493=1493, 0 failures.

### Required before approval

1. **`LeaderPut` has no observable refusal** (`KVStore.java:146–158`). `handleLeaderPut` computes `current`, runs the transaction, discards the `TransactionResult` and returns `current` — a stale-leader rejection with `expected == current` is indistinguishable from acceptance. And `RabiaEngine.commitChanges` (`RabiaEngine.java:2045–2058`) succeeds every correlated promise with the whole merged batch's result list, so a proposer cannot find its own entry — the reason `TransactionResult` carries a `transactionId`. The changelog's "rejected transactions report a correlated refusal" is true only of `LeaderTransaction`; the spec's H04 ("rejections must be visible to the submitting operation") is violated by this API. **Fix:** `LeaderPut` returns a `TransactionResult` with a caller-chosen id (or is removed in favour of a one-mutation `LeaderTransaction`). Pin: submit a stale-leader `LeaderPut`, assert the returned result is a refusal naming that id. Repro: probe `leaderPutResult_isPreStateWhetherAcceptedOrRejected`.
2. **A plain `Put` on a `LeaderAuthorized` key is silently ignored** (`processCommand`: returns the current value exactly as a successful put would). Same H04 violation from the other side. Either refuse visibly (a result the caller can distinguish) or document, in the `LeaderAuthorized` javadoc and the changelog, that unmarked puts against marked keys are dropped — and pin whichever you choose.
3. **The "complete read set" is exercised by no PR test.** Every `LeaderPut`/`LeaderTransaction` in the four new classes passes `guards = List.of()` (`KVStoreLeaderPutTest.java:34,35,42,52`, `KVStoreLeaderTransactionTest.java:36`, `KVStoreNotificationIsolationTest.java:74`); a mutation that ignores guards entirely leaves all 77 `KVStore*` tests green. The headline mechanism — refuse because a *read* key moved while no *written* key did — has no pin. Adopt probe `readSetFailsOnSecondKey_noKeyChangesAndNoNotification` (it also asserts zero notifications on refusal, which nothing else does).
4. **Same-epoch owner transfer is possible via a witnessed bare `Remove(incumbent)` followed by `Put(contender)`.** `staleRemove` (`KVStore.java:350–363`) accepts the incumbent's own value as a non-stale witness; after the remove, a same-epoch `Put` passes `staleEpochWrite` (nothing stored). The transaction path forbids removing an `OwnerFenced` value (`validMutation`, `:192`), so the two paths disagree on whether the fence can be dissolved. The `OwnerFenced` javadoc ("cannot transfer authority to a different identity") and the PR body are unqualified. Close it (`staleRemove` returns true for an `OwnerFenced` value unless the witness carries a *higher* epoch, mirroring `currentLeaderWitness`) and pin with probe `witnessedBareRemoveThenPut_transfersOwnerAtSameEpoch_transactionPathRefusesTheRemove` inverted.
5. **Read-side cost is unstated and unpinned.** `get`/`getTyped`/`snapshot` are now `synchronized` (`KVStore.java:508–533`) and `forEach`/`find` (`:543`) `Map.copyOf` the whole store under the monitor per call — 33 production call sites (`DashboardMetricsPublisher`, `TopologyCollector`, `OwnershipEpochHighWater`, `DynamicConfigManager`, …) that were lock-free weakly-consistent reads before. "Atomic reads" is the upside; state the cost in the body and changelog, and pin the property you are buying (a reader never observes a half-applied batch) so the synchronisation cannot be silently removed.
6. **Unpinned hunks, each needs a reddening test:** `makeSnapshot`'s `canonical()` selection (probe `kvStoreSnapshot_identicalAcrossInsertionOrder_collidingHashKeys` reddens at :82 when the plain serializer is substituted — adopt it); the `dispatching` reentrancy guard (`KVStore.java:95` — pin with `reentrantNotification_seesStoreAfterAllNestedApplies`, or remove it if the deque alone is the contract).
7. **Body/changelog precision (lens):** "same-epoch refresh cannot change owner" → *a single `Put` or transaction cannot* (see 4); "readers remain responsive when a listener blocks" → true, and **writers wait on `mutationLock` for the listener** — say so (harmless today because the single Rabia apply worker serialises them, `RabiaEngine.java:124`); "deterministic unordered collections" → for `Map`/`Set` written through `FrameworkCodecs` or generated record writers, not hand-rolled formats (`EntityFoldSnapshot`) or custom `Serializer`s (`canonical() == this`, `Serializer.java:32`).

### Nits
- `concurrentClaim_oneWinnerAndVisibleReadback` (`KVStoreLeaderPutTest.java:31–37`) is sequential; rename or make it race (probe `racingTransactionsOnOverlappingReadSet…` is the real one).
- `TransactionResult` names the transaction, not the conflicting key. Fine if intended — say so in its javadoc; a caller wanting the key has to re-read.
- `Serializer.canonical()` javadoc states an implementor requirement as a method property. "Returns `this`; a custom serializer's canonical form is whatever it already emits."
- `SystemTags.java` churn is `jbct:format` repairing a pre-existing base failure — fine — but the formatter mis-indents `rejectDuplicateTags` (`:94–101`); a `jbct-format` ticket, not yours.

Next round: push, and list per item what changed with file:line; the adversarial pass re-runs its 16 mutations plus the new pins against the new head.

siy:
## Review round 2 (`9f8a20fcd`) — APPROVE

All seven required items are done and pinned, re-verified by mutation, not read: `LeaderPut` removed (tag 123 gone from the baseline; the diff is exactly +124..127) and a stale-leader single mutation is refused with the caller's id; the plain-`Put`-on-`LeaderAuthorized` silent drop is documented and pinned with a zero-notification assert; the read set has a two-`ReadWitness` test (ignoring guards reddens it); the `Remove`+`Put` owner transfer is closed by `newerOwnerWitness` requiring a strictly higher epoch (both the `>=` and the arm-removed mutations redden); read-side cost stated in changelog and body with `snapshotCannotObserveHalfAppliedBatch` reddening under an unsynchronised `snapshot()`; canonical snapshot and the reentrancy guard pinned; wording per the lens. 23 mutations, every one reddens a PR test except three stated as unpinnable (`get` unsynchronised, replay without `mutationLock`, `forEach` copy). Counts: cluster 144 → 166 (+22 new methods), serialization 40 → 42, node 1617 = 1617, 0 failures; clean merge with the current rc4 tip `d444d22c6`, PR modules byte-identical.

Two nits, not blocking:
- `.github/workflows/hierarchy-review.yml` is gated on `head_ref == feat/hierarchy-primitives-rc4 / -runtime-rc4` — permanently skipped after merge (the node-axis review of #1390 found the same for the hierarchy Forge classes). Generalise the trigger to a path filter, or delete the job at merge; the 7-mutation check should keep running on the target branch.
- "tag 123 retired, do not reuse" spends one of the ten remaining one-byte slots on a type that never shipped; pre-GA (owner ruling 2026-09-16) reuse is safe — owner's call.

Merge target: rc5 (ruling on #1390); this PR lands there before #1390. Verdict: `oss/internal/sweep-state/s25-rev1379.md` round 2.

siy:
Addressed the remaining round-2 nits in `fff5d7fcd`:

- `hierarchy-review.yml` now triggers on relevant paths for all PR bases (including stacked PRs) and main/release pushes, with no branch-name condition on the foundation job. Mutation checks therefore continue after merge. Maven steps explicitly remove `HCLOUD_TOKEN`.
- Tag 123 is documented as available for reuse because `LeaderPut` never shipped before GA. Existing tags are unchanged; no unnecessary renumbering was introduced.

Both workflow variants parsed successfully with Ruby YAML, and their PR/push path filters and absence of branch-specific job gates were checked. Whitespace validation passed; fresh CI is running. The owner's later merge-timing instruction supersedes the earlier rc5-only request.

siy:
## Review round 3 (`fff5d7fcd`) — APPROVE-WITH-NITS (one nit to take before or with #1390)

NIT-A closed for the foundation: `foundation-mutations` is now keyed on paths (`pull_request` + `push` to `main`/`release-*`, filtered on `aether/**`, `integrations/{cluster,consensus,serialization}/**`, `core/**`, the script and the workflow) with no branch-name gate — it survives merge (measured on this head: 1m53s, success). NIT-B closed (tag 123 "available for reuse").

One new nit, same file: `runtime-acceptance` lost its `head_ref` gate WITHOUT a replacement condition and shares the broad path filter, so it now runs on every PR/push touching `aether/**`. On this head's own run (35638270186) it took 7m47s and reported `Tests run: 22` from the 4 of 16 named classes that exist on this branch — surefire silently dropped the other 12 — a green job that examined a subset; once #1390 lands it is a ~30-minute Forge job on every rc4/rc5 push. Narrow it with a job-level `paths:` on #1390's `Hierarchical*` test directory (or a label/`if:` on that job only), and make the selector fail on a missing class rather than drop it. One line; take it here or in #1390 before merge.

Verdict file: `oss/internal/sweep-state/s25-rev1379.md` round 3. Merge target rc5, after #1386.

siy:
Addressed the runtime-job nit in `9a40851aa`. The foundation workflow now runs only the guarded-KV mutation acceptance it actually contains; it cannot report success for a partial runtime selector.

The complete runtime job moves to a separate workflow in #1390. It is explicitly selected with the `run-hierarchy` PR label (including stacked PRs), manual dispatch, or a main/release push changing hierarchy acceptance tests/checker/workflow. Unrelated `aether/**` changes no longer start the expensive matrix. A pre-build check rejects any absent selected class, and a post-run check rejects missing, empty, failed, errored or skipped class results, including nested classes. Seven checker regressions cover the partial-green failure shape.

YAML parsing and whitespace checks passed for this foundation-only change. No Java behavior changed. The runtime checker and source changes receive their own build/live validation in #1390. The owner’s latest direction still allows the batch to merge into rc4 when reviewed and ready; there is no rc5 timing prerequisite.

siy:
**Adversarial delta review @ `9a40851aa` — APPROVE-WITH-NITS, no blocking findings.**

Reviewed the branch's own content (`0002e2194..9a40851aa`, 19 files) rather than the 437-file diff
that the five rc4 merges produce. All five findings from the previous round are addressed in code —
`LeaderPut` removed (tag 123 free, baseline delta is exactly +124/125/126/127 with zero removals),
the plain-`Put` drop documented and pinned, the read-set refusal pinned by a test that probes
behaviour rather than restating it, the same-epoch `Remove`+`Put` transfer closed by a
strictly-newer-witness rule, and the read-side cost stated with its residue disclosed.

**The merge commit `90822a3fc` is correct, and mechanically so.** Instead of reading the resolution I
computed the naive auto-merge and diffed against it: exactly one content conflict
(`incomingEpochIsStale`), and the resolution changes only that hunk plus one added test — nothing
that merged cleanly from either side was dropped. The resolution takes the **disjunction** of the two
predicates rather than choosing a side, so it is never more permissive than either parent and is
stricter than rc4 in exactly one cell (equal epoch, stored `OwnerFenced`, different owner,
non-minting). That cell is unreachable today — nothing in production implements `OwnerFenced` (one
grep hit repo-wide, a test; positive control on `EpochBearing` reaches `aether/`) — so the merged
rule is behaviourally identical to rc4's for every value a node can construct. rc4's `mintsEpoch`
rule survives and the merge brought its pin with it. Against the current rc4 tip `9801de082` the
merge is clean and none of the 28 intervening commits touch this PR's modules.

**Deleting `runtime-acceptance` is the right call and costs no coverage.** The job named 16 classes
of which 12 do not exist here (surefire dropped them silently — a green job examining a quarter of
what it named). All 4 real ones live in `aether/forge/forge-tests`: three are `@Tag("Smoke")` and run
in `ci.yml`'s `forge-tests` on every PR, and `EmberAddNodeRoleLabelTest` is `@Tag("Heavy")`, covered
by `heavy-forge.yml`.

**I validated the mutation script rather than trusting its score.** 7/7 KILLED reproduces at this
head (38 s). Adding an 8th row for a behaviour already known to be unpinned — `get`'s
`synchronized` — it correctly reported `NOT VERIFIED` and exited 1, so it is not a pass-everything
harness. Three things worth fixing, none blocking:

1. **The read set's acceptance arm is unpinned.** Inverting `matchesWitness`
   (`KVStore.java:167-170`) so that a *satisfied* guard refuses and a *stale* one passes leaves all
   93 `KVStore*Test` green. The existing test passes one matching and one stale witness and expects
   refusal, so inversion keeps it green by construction; the script's `read-set` row (`return true`)
   probes only the permissive direction. Repo-wide, `ReadWitness` is constructed in exactly one test
   method, and its expected outcome is refusal — no test asserts that a satisfied read set is
   *accepted*. One test plus one script row closes it.
2. **`LeaderAuthorized` has no downgrade guard, unlike `OwnerFenced`.** `staleWrite` carries
   `dropsOwnerFence` but no counterpart, so one leader-authorized transaction can replace a
   `LeaderAuthorized` value with a plain one — after which a bare `Put` with no leader witness lands
   and a *witnessless* `Remove` deletes the key. Demonstrated with an in-run control (the bare `Put`
   is dropped while the marker is in place) and the symmetric `OwnerFenced` case refusing. Possibly
   intended — the marker is a property of the value, not the key — but undocumented, and the javadoc
   reads otherwise to whoever writes the first caller.
3. **The script credits a kill to any failure in the target class.** It never checks the suite is
   green before mutating, so a pre-existing red is scored as a kill: the identical mutation flipped
   `NOT VERIFIED` → `KILLED` and the script's exit `1` → `0` when I added one unrelated failing test,
   with `KVStore.java` byte-identical to HEAD. The `Verify restored implementation` step still
   reddens the job on a persistent bad baseline, so the gate holds — but the uploaded evidence
   artifact can show an unearned 7/7, standalone runs are unprotected, and a red of this shape
   reports the wrong cause. Running `KVStore*Test` once before the loop fixes it.

Smaller notes: the script reads the `<testsuite failures>` attribute, which reports 0 for `@Nested`
classes — the five current targets are safe, but `KVStoreEpochFenceTest`, `KVStoreLeaderFenceTest`
and `KVStoreRemoveFenceTest` in the same module already show the defect (0 vs 15/9/20), and they are
the fence tests. Also, `ci.yml`'s gate passes no `-Djbct.skip=false`, so both of this PR's production
modules are skipped silently by the repo's format/lint gate (pre-existing, not this PR); measured by
hand, `integrations/cluster` is 34 files / 0 format / 0 lint, and the PR *repairs* the base's one
`serialization/api` format issue while leaving its 25 pre-existing lint errors unchanged. Adding
`jbct:check -Djbct.skip=false -pl integrations/cluster,integrations/serialization/api` to
`foundation-mutations` would gate these modules for the first time.

Counts at this head, all zero failures: `integrations/cluster` 166, `integrations/serialization/api`
42, `KVStore*Test` 93, wire tripwire + codec pinning 6/6 (and the tripwire reddens when the baseline
is reverted, so it is live). Not reached: the forge symptom (CI only), a flakiness bound (0 non-green
in 25 runs is not one), a test run of the merged tree, and the full `aether/node` suite.

Issue #1380, opened 2026-09-20
ReplicationBatcher.close() returns before an in-flight one-shot flush has sent (flaky add_concurrentWithSizeAndTimerFlushes_losesNoEvent on the rc4 tip)


siy:
Observed once in CI on PR #1364's merge ref (a docs-only PR, so the code under test is the `release-1.0.0-rc4` tip `866c05b3b` + docs), run at 13:35:54Z 2026-09-20:

```
[ERROR] org.pragmatica.aether.stream.replication.ReplicationBatcherTest$AccumulatorEviction.add_concurrentWithSizeAndTimerFlushes_losesNoEvent -- Time elapsed: 0.018 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: 20000 but was: 19995
at ReplicationBatcherTest.java:323
```

The test (added by #1293 for #1246, "evict drained accumulators") runs four virtual threads adding 5,000 events each into a batcher with batch size 7 and a 1 ms timer, then `close()`, then asserts every event was delivered. Five were not.

Two candidate mechanisms, to be discriminated by reading the producer and reproducing, not by arbitration:
1. **Product:** `add()` obtains an accumulator from the map, a concurrent size/timer flush drains AND evicts that accumulator, `add()` appends to the evicted instance → the event is never flushed. This is the race that eviction (#1246) creates by construction and that the test exists to pin.
2. **Test/close:** `close()` discards events still pending in an accumulator instead of flushing them, so the last partial batch of a thread that finished just before `close()` is dropped. That is still a product defect (a replica batcher that loses buffered events on close) but a different one.

The assertion failed at 0.018 s, i.e. the loss happened under real contention, not a timeout. Note the CI reactor is `-T 1C` (`.mvn/maven.config`), so module test phases run concurrently and the box is loaded.

Wanted: reproduce (loop the class on the build host, quote the sample size and failure rate; `-T 1C`-loaded and isolated arms interleaved), identify the mechanism from the code, fix it, and make the pin deterministic (inject the interleaving via the existing seams rather than relying on load). Until then every rc4 PR's CI carries this flake.

siy:
[...]

PR #1384, opened 2026-09-20
fix(stream): ReplicationBatcher.close() waits, bounded, for in-flight flushes (#1380)

siy:
Closes #1380

## What

`ReplicationBatcher.close()` returned while a one-shot flush that had already `drain()`ed its accumulator was still on its way to `transport.send` (the scheduler runs bodies on a separate virtual thread; `cancel(false)` cannot stop a started body; close's own drain of that accumulator yields `EMPTY`). No event was ever lost — it arrived microseconds later — but the pristine `add_concurrentWithSizeAndTimerFlushes_losesNoEvent` reads `delivered` the instant `close()` returns, so it saw one partial batch missing (CI: 5 of 20,000). Root cause: `inv1380` (`oss/internal/sweep-state/s25-inv1380.md`).

`close()` is now a barrier: `flushPartition` counts itself in flight from before its drain until its send returns, and `close()`, after its own cancel-and-drain pass, waits for that count to reach zero — **bounded** by `DEFAULT_CLOSE_BARRIER` (5 s). At the bound, or if the closing thread is interrupted (flag preserved, #914 shape), it logs a WARN naming the count still in flight and returns; the flushes complete on their own threads (late, never lost), and `close()` never hangs on a transport that does not return. Package-private factory overload takes the bound (test seam, like `FlushScheduler`). Hot-path cost: one `AtomicInteger` inc/dec per flush.

Tangential, same file, all one-liners: class docstring now says an `add` racing `close()` that finds its accumulator retired is *refused* by the retry's `closed` re-check (it said "may still append"); `flushAll()` cancels the one-shots it makes redundant (was leaving 10,000 no-op timers on the process-wide scheduler); `StreamPartitionManager.close()` comment said "fixed-rate flush" — one-shot per batch since #1246.

Blast radius at the base: **zero production callers** (`AetherNode` wires `batcher = none()`); if wired, the exposure is shutdown sending into a network being torn down.

## Evidence (build host `cto24-build-1`, `-pl aether/aether-stream`; base `866c05b3b`, rebased head `d986e46f0`)

- **RED** at `866c05b3b` with the new test only: `ReplicationBatcherTest$CloseBarrier.close_oneShotFlushBetweenDrainAndSend_waitsForItBeforeReturning` — `[delivered as of close() returning — the pristine test's assertion at :323] expected: 5 but was: 0` (`Tests run: 20, Failures: 1`). Same signature as inv1380's probe (afterClose=0, afterJoin=5).
- **GREEN** with the fix: `Tests run: 23, Failures: 0, Errors: 0, Skipped: 0` for `ReplicationBatcherTest*`; both WARN paths observed in the log (`interrupted with 1 flush(es) still in flight`, `gave up after 200 ms with 1 flush(es) still in flight`).
- **Mutations** (each applied by content, sentinel-counted, restored by `git checkout`, 23 tests run each):

  | Mutation | Red |
  |---|---|
  | remove the `awaitInFlightFlushes()` call | `close_oneShotFlushBetweenDrainAndSend…:530` (delivered 0), `close_inFlightFlushOutlivesTheBound…:549` (elapsed < bound), `close_closerInterruptedWhileWaiting…:570` (close returned at once) |
  | remove `incrementAndGet` | same three |
  | remove `decrementAndGet` | `close_oneShotFlushBetweenDrainAndSend…:527` (returned at the bound, not promptly), `close_inFlightFlushOutlivesTheBound…:558` (count never returns to 0) |
  | remove the interrupt exit (`if (false)`) | `close_closerInterruptedWhileWaiting…:575` (spun to the 5 s bound) |
  | `flushAll` back to `flushPartition` | `ManualFlush.flushAll_cancelsTheOneShotOfEveryBatchItDrains:133` |
- **Module run** at `d986e46f0`: `Tests run: 1134, Failures: 0, Errors: 0, Skipped: 0`; `grep -o '<testcase ' target/surefire-reports/*.xml | wc -l` = 1134, `<skipped` = 0 (at the pre-rebase head `85a354a00`: 1104/1104/0; the +30 are upstream's).
- **`AccumulatorEviction` ×10** in a loop, one Maven invocation each: 10/10 `Tests run: 3, Failures: 0`, XML 3 each. [unverified: CI's `-T 1C` reactor rate — inv1380 measured 1/5 cold-JVM failures on the shared host before the fix; 10 clean runs is a sample, the deterministic pin is the evidence]
- **Gate** `jbct:check -Djbct.skip=false -pl aether/aether-stream` at `d986e46f0`: `Running JBCT check on 97 Java file(s)` … `JBCT check passed.`
- Changelog fragment: `changelog.d/1380-batcher-close-barrier.md`.


siy:
rev1384 M1 adopted (head `e6ed2fa1c`): `CloseBarrier.close_oneShotStartedDuringTheDrainPass_isAwaited`, from `oss/internal/probes/s25-rev1384/Rev1384ProbeTest.java` (credited in the test doc). Pins the order drain pass → barrier wait: R5 (wait moved before the pass) reddens exactly this test 3/3 (`close() must not return while a flush that started during its drain pass is in send`), the other 23 stay green; at head 24/24. Gate 98 files passed (local, offline — build host down). N1 (WARN names the count, not the partition) left as recorded: structural, scalar counter.

PR #1386, opened 2026-09-20
docs(aether): specify hierarchical consensus, placement and worker contracts

siy:
Defines the proposed end-to-end hierarchy contract and implementation plan across consensus, community authority, placement, worker execution, metrics and metadata. This PR now contains only eight new specifications and their changelog fragment. Existing-product documentation changes belong to runtime #1390.

Review responses:

- **B1:** Removed all twelve changes to existing product documents from this PR. They are retained and corrected in #1390; this pure specification has no code prerequisite.
- **M2:** Replaced historical aggregate counts with a reproducible ledger in `hierarchical-cluster-implementation-plan.md`: source revisions, worktree, exact commands, named classes, and linked CI artifacts. Older private-worktree counts are withdrawn as merge evidence. Runtime acceptance is tracked in #1390’s committed reconciliation and CI.
- **M6/M7:** The immutable-role contract is explicit. Runtime #1390 owns the corresponding promote-route refusal, CLI/catalog corrections, and retained spokesman-path documentation.
- **M1:** Qualified rc4 observations versus behavior proposed by #1390 in the hierarchy, voter, and metrics specifications.
- **M3/M4:** Specified parent-directory fsync and crash boundaries; documented post-handoff-barrier write unavailability, no timeout rollback, and operator repair.
- **M5:** Defined a persisted monotonic producer epoch, fail-closed allocation, and separation from SWIM membership incarnation in metrics and placement specifications.
- **Limits and N1–N9:** Added greppable limits; standardized K for core count; stated f+1 of 2f+1; identified the configuration environment fallback; used `[spec:]` in the changelog; supplied numeric buffering limits. Runtime #1390 carries the actual API-field example and historical-plan/spokesman corrections.

Review order: #1386, guarded-KV foundation #1379, integrated runtime #1390. Per the owner’s revised release decision, the batch may merge into rc4 when ready; creation of rc5 is not a prerequisite. No merge or approval is performed by this work.

Validation: whitespace and relative specification links checked. The plan links passing foundation mutation CI and distinguishes compile, unit, live-cluster and unmeasured physical-scale evidence. No 10K-node or WAN throughput claim is made. Runtime #1390 now has passing current-head repository CI and a verified 50-case live acceptance artifact; its PR and committed reconciliation identify the implementation evidence. Breaking pre-GA schemas are intentional; no migration is proposed.

[...]

Issue #1038, opened 2026-09-12
Disabling auto-heal suppresses the replacement but not the deletion: a node whose JVM dies has its VM deleted within minutes, destroying all diagnostic evidence


siy:
Measured on a live 5-node Hetzner cluster, 2026-09-12, from the provider's own per-server action history.

## What was measured

A stream killed the JVM (by PID) on three nodes in turn, to observe whether a supervision unit restarted it. **Every killed node became completely unreachable — no ICMP, port 22 filtered — before `systemctl` could be read.** The provider action log explains why:

```
165595793   created 11:18:22 → delete_server 11:23:37 → 165596152 created 11:23:52  (+15s)
165595773   created 11:18:01 → delete_server 11:30:57 → 165596603 created 11:31:12  (+15s)
165595751   created 11:17:46 → delete_server 11:38:41 → no replacement
```

**All three were `delete_server`. None was stopped.** A stopped server still exists and can be inspected; a deleted one cannot.

## The part that is a defect rather than a design

Before the third kill, auto-heal was **disabled**, and the disabling was confirmed by consequence rather than by assertion — the cluster reported `deficit: 1` with `lastReason: AUTO_HEAL_DISABLED`, and **no replacement was minted**, which the action log independently confirms.

**The VM was deleted anyway.**

So the flag that gates *provisioning a replacement* does not gate *destroying the unhealthy node*. An operator who disables auto-heal — the documented way to stop the cluster acting while you investigate — still loses the machine they were trying to investigate.

## Operational consequence

**A node whose JVM dies takes its whole VM with it, within minutes, and there is currently no way to prevent that in order to debug it.** Every artifact needed to diagnose the original failure — `journalctl`, `systemctl status`, the JVM's own output, cloud-init logs, heap dumps — is destroyed along with the machine.

This is the runtime sibling of the bootstrap-time problem in the `--keep-on-failure` issue: in both cases the system removes the evidence before anyone can read it, and in both cases the operator is told to go and read something that no longer exists.

## What is NOT established

The action log records **what** happened and **when**, not **which component issued the call** — every action is attributed to the same API token. That deletion follows a JVM kill by minutes, and that two of three were followed 15 seconds later by a replacement, is strong circumstantial attribution to the auto-heal/topology-manager path, but the actor is **not** named by this evidence.

## Instrument note for whoever measures this next

[...]

Issue #660, opened 2026-08-26
Rabia sync adoption is off-by-one vs quorum: bare-majority cold start (3-of-5) deadlocks silently in Syncing — self counted for quorum establishment but not for sync responses


siy:
Found 2026-08-27 while gating the #642 fix batch (probe run 2), diagnosed by log+code investigation; recorded on #642's trail. Evidence: `aether/tests/integration/failure-logs/2026-08-26-509-probe-staggered-restart-selfdissolve/` — BOTH runs stall identically (run 1's stall was masked by the #642 ghost kills at +49s; run 2, ghost-fixed, let it run 2 hours).

## Defect

`RabiaEngine.syncQuorumSize()` = `clusterSize/2 + 1` (3 for a 5-node config), but sync RESPONSES only ever come from PEERS: `broadcastPayload` iterates `peers.values()` and self is never registered as a peer. So `handleSyncResponse` can never accumulate `syncQuorumSize()` responses unless `quorum + 1` nodes are live. **A bare-majority cold start deadlocks silently in `Syncing` forever**: 3-of-5 never activates (observed: `activePeerCount=2, clusterSize=5, quorumSize=3` then two hours of nothing); a 3-node cluster with one node down needs all 3. `activate()` is reachable only through the two quorum-gated `restoreState` call sites — there is no other exit from `Syncing`.

Consequence chain: consensus never ACTIVE → no `ClusterStateNotification` → `NodeDeploymentManager` never dispatches `QuorumEstablished` → leader-election FSM never leaves `Dormant` → no leader, no reconciler, cluster dead while every link and SWIM view is healthy.

**Observability hole, part of the defect:** the sync retry loop logs only at TRACE (`doSynchronize`), so the deadlock is invisible at INFO — 30 MB of log with zero indication. Consistent with the silence-killer doctrine: promote to a periodic WARN with the response-count arithmetic in the line.

## Origin

Introduced by `36712ba5a` (2026-08-17), which correctly removed the connectivity-derived sync threshold (the "sync quorum hit 1 at the partition edge" fix) but overshot by one: quorum ESTABLISHMENT counts self (3/5 live → ACTIVE routed to the engine), sync ADOPTION does not. The net requirement silently became quorum+1 live nodes.

## Production relevance

Any rolling restart / partial outage that leaves a bare majority to cold-start converges to a healthy-looking dead cluster (links up, SWIM green, no leader). Not a harness artifact.

[...]

PR #1390, opened 2026-09-20
feat(aether): implement durable core authority and hierarchical worker lifecycle

siy:
Aether’s hierarchical runtime now separates installed core voters from worker discovery, commits governor authority through guarded consensus, and preserves durable identities across recovery and placement. Workers cannot enter the core quorum or change roles. Missing metrics hold scaling without contaminating its historical baseline, and worker management endpoints forward cluster-wide reads to cores or explicitly refuse unsupported scope.

Review and merge **#1386 → #1379 → this PR**, then #1405 for the lifecycle reconciliation continuation. The owner permits merging the completed, approved batch into rc4; there is no rc5 timing prerequisite. Green checks alone do not constitute review approval.

| Review axis | Principal paths | Behavior to verify |
|---|---|---|
| Consensus | `integrations/consensus`, `integrations/cluster` | Ordered application, durable vote-before-send, checkpoint-only and WAL recovery, immutable voter configurations and certified handoff. Worker proposals and both vote rounds are excluded. Late and queued requests receive explicit shutdown refusals. Reconfiguration state and refusals are operator-visible. |
| Node assembly and authority | `aether/node`, `aether/ember` | Immutable roles, committed governor acquisition, guarded mutations, authenticated scoped health, assignment/replay/shutdown behavior. Explicit durable control storage and isolated test storage; process epochs survive restart and corruption refuses startup. |
| Deployment and placement | `aether/aether-deployment`, `aether/environment`, `integrations/cloud` | Source/account binding, shared capacity reservations, destination readiness before retirement, durable movement identities. Definitive no-create evidence survives conflicting ledger release, restart and lost completion; periodic reconciliation retries safe cleanup. |
| Metrics and metadata | `aether/aether-metrics`, `aether/aether-control`, `aether/aether-config`, `aether/aether-ttm`, worker metadata/health | Independent SWIM-incarnation and durable-process-epoch fences; producer identity, freshness, deduplication and coverage; bounded source distribution and metadata repair. Partial coverage never enters scaling history. Global metrics have explicit scope; node CPU/heap values use published keys and units. |

The [requirement reconciliation](https://github.com/pragmaticalabs/pragmatica/blob/feat/hierarchy-runtime-rc4/aether/docs/specs/hierarchical-cluster-reconciliation.md) maps H01–H10 and H-T01–H-T12 to implementation, named tests and remaining evidence. The [validation ledger](https://github.com/pragmaticalabs/pragmatica/blob/feat/hierarchy-runtime-rc4/aether/docs/specs/hierarchical-cluster-validation.md) separates historical checkpoints from final-head execution. The [durable-control recovery runbook](https://github.com/pragmaticalabs/pragmatica/blob/feat/hierarchy-runtime-rc4/aether/docs/operators/durable-control-recovery.md) explains fail-closed storage recovery and stalled handoffs.

This is a **pre-GA breaking change with no migration requirement**. New wire assignments are pinned. Node promotion cannot change roles; metrics latency fields are `intervalMeanLatencyP*`, and `errorRate` is a failure ratio rather than failures per second. On-disk restart correctness remains required independently of upgrade compatibility.

Material limits:

- Consensus assumes crash faults; identity-list certificates are not Byzantine proofs. Torn/corrupt durable evidence refuses startup. An agreed handoff has no timeout rollback.
- Core snapshot transfer remains bounded by the 32 MiB encoded frame envelope. Lane isolation provides neither scheduler priority nor a measured guarantee against connection-wide QUIC flow-control starvation.
- AWS/Hetzner classify definitive capacity refusal; unclassified GCP/Azure failures remain uncertain. Unknown creates retain identity/capacity; there is no force-clear API. #1405 supplies additional binding/readiness resumption, provider-confirmed stale-placement cleanup and the auto-heal deletion gate.
- Bounded fixtures and small live clusters do not establish physical 10K-node or WAN throughput. Core-hosted state, application dependency graphs and metric cardinality remain separate resource costs. TTM features remain positional and unversioned.

**Validation**

Implementation `aca8c8864` passed the complete six-step build/JBCT, the full consensus suite (**859 cases**), 43 node health/authority/assembly cases, and the targeted live hierarchy gate (**23 cases, zero failures/errors/skips**, 531 seconds). All 70 runtime dependencies were fresh. Membership passed 344 cases, the combined controller selection passed 13, and targeted mutations of the evidence-domain, terminal-side-effect, freshness and shutdown-refusal guards failed as intended before exact source restoration. The ledger records scope and ordering precisely. No runtime build/install ran during live Forge. New-head CI is pending.

The previous implementation passed the complete 50-case live hierarchy matrix; that historical checkpoint is recorded separately in the ledger. It is not substituted for validation of the latest changes.

Hierarchy runtime acceptance has its own workflow. The `run-hierarchy` label opts PRs into the full matrix, including this PR and its lifecycle continuation; manual dispatch is available, and main/release pushes changing hierarchy acceptance tests, the validator or its workflow run automatically. Preflight rejects absent selected classes; postflight rejects missing, empty, failed, errored or skipped results. Seven validator regressions passed.

Independent review and merge approval remain outstanding. Green checks alone do not constitute approval.


siy:
## Review — request changes (not yet reviewable)

**Target: rc5.** This runtime does not land on `release-1.0.0-rc4`. rc4 is being cut from the current queue; `release-1.0.0-rc5` will be created from it, and this PR retargets there. Until then keep the base for reference only — the ref is CONFLICTING against rc4 (base `ccba0dba5`, 46 squashes behind; `AetherNode.java` and `ManagementApiResponses.java` are the code collisions) and no CI has run on it, so nothing in the body is currently checkable by anyone but the author.

What has to be true before an adversarial review starts — these are preconditions, not findings:

1. **Rebase onto the rc5 branch tip** when it exists (merge the tip into the branch, no force-push), so CI runs on a merge ref. Every claim in the body ("final Forge matrix passed all 27 cases", module counts) is re-established on that head, in this repository's CI, not in a private worktree. Evidence that only exists in a worktree nobody else can read is not evidence for a merge decision.
2. **The reconciliation table from #1386 §11 is filled in and committed** (`aether/docs/specs/hierarchical-cluster-implementation-plan.md` or a sibling): every requirement H01–H10 and every H-T01…T12 scenario marked DONE / MISSING / STUB / SHORTCUT / OMISSION / SIMPLIFICATION, each DONE row citing the test class (and forge case) that pins it at the PR head SHA. This is the acceptance instrument the spec defines; the runtime is complete only when no row is in the last five states, and the review is a check of that table row by row.
3. **The review map in the body becomes a review PLAN by axis**, because 534 files / 27.7k lines cannot be reviewed as one unit: (a) consensus — `integrations/consensus` (70 files): distinct log slots, binary rounds, ordered application, voter configuration handoff, durable voting evidence, `DurableRabiaPersistence` (including the directory fsync after checkpoint rename that #1386 omits); (b) node assembly and authority — `aether/node` (217): immutable roles, committed governor acquisition, guarded multi-key mutations through #1379's primitives, scope-aware health, the fate of the `PromoteNodeRequest` route (`NodeLifecycleRoutes.java:387–405`) now that roles are immutable; (c) deployment and placement — `aether/aether-deployment` (60) + `aether/environment` (15) + `integrations/cloud` (7): source/account binding, capacity reservation, resumable community movement, retirement; (d) metrics and metadata — `aether/aether-metrics` (31) + `aether/aether-config` (14): origin-preserving metrics, freshness, no double counting, incarnation ordering (#1386 M5 — state where the incarnation is persisted). One adversarial reviewer per axis, each with the spec section as the claim list. State in the body which commits belong to which axis (`ac6c067a6..e4bdd3e7e` is the runtime-only range you cite — make the split explicit).
4. **Dependencies land first and the review order is enforced:** #1386 (with its B1 fix — the nine runtime-narrating doc files move HERE, where they become true) and #1379 (with its seven required items, notably a visible refusal for `LeaderPut` and the read-set pin, since this runtime is the first real caller of those primitives and inherits every unpinned hunk).
5. **Wire and persistence:** every new KV record / message type has a pinned tag in the wire baseline (`WireAssignmentTripwireTest`), and the durable consensus state has a stated format version with a crash-point test list (rename-then-fsync is one crash point). "No mixed-version migration contract before GA" is accepted (owner ruling 2026-09-16) — that is about upgrades, not about a node restarting on its own on-disk state, which must be pinned.
6. **Scale claims:** per #1386 §10, no throughput or 10K-node claim without hardware, workload cardinality, warm-up and percentiles; the 100×100 synthetic topology establishes algorithmic bounds only. The body currently makes none — keep it that way until measured.

The review itself starts when 1–2 are done; 3–6 shape it. Expect several rounds; every finding will carry file:line and a reproduction, and every "fixed" claim will be re-run, not read.

siy:
**Merge target ruling (CTO, 2026-09-21, owner-confirmed): rc5.** The body's "revised release decision … may merge into rc4" does not hold. Sequence: rc4 is cut when its `severity:blocking` set is empty and the tip is green; `release-1.0.0-rc5` branches at the cut; this batch (#1386 → #1379 → #1390) is rc5's first merge as soon as its reviews are complete — if the reviews finish first, the cut moves, not the merge. Until the branch exists the base stays rc4 for CI only; the three PRs keep taking tip merges. The by-axis adversarial review (consensus / node+authority / deployment+placement / metrics+metadata) is running against `e1eb4a344` with the reconciliation table as the claim list; nothing merges on CI green alone.

siy:
## Interim BLOCKING — consensus axis (`e1eb4a344`)

**B1 — a clean stop followed by a restart activates an EMPTY state machine and can re-decide a committed slot (H02).** `RabiaEngine.recoverLocalState` (`:1546-1549`) returns early when the journal is empty. A CLEAN stop (`performStop → saveAuthority`) always leaves `checkpoint.bin` plus an empty WAL once the open slot has closed, so on restart the node never loads its own checkpoint; `ownStateFloor` still refuses every peer response behind that checkpoint, and `activateWithoutAdoption` activates at phase 0 with nothing processed.

Measured (probes under `oss/internal/probes/s25-rev1390-consensus/`): p1 — disk says frontier=1 / snapshot=9 B / journal=0 → after restart `active=true phase=0 processed=[]`; p1b — the restarted node then applies `Decision(slot 0, "different")` and appends it to the WAL beside a checkpoint that says slot 0 = "committed"; control p1c — a CRASHED stop (WAL survives) restores correctly, which is the only arm the existing pin covers (`committedDecisionReplaysOnce…` sets `disk.crashed=true` before stopping). Reachable fault: A clean-stops ahead of C, B is down, A restarts → A and C can re-decide a slot B already holds.

Required: recovery loads the checkpoint whenever it exists, regardless of the journal (`if (journal.isEmpty() && saved.isEmpty())` is the shape the reviewer is confirming by mutation now), and a pin for the CLEAN-stop arm: clean stop → restart → `processed` equals the checkpoint's committed prefix and a conflicting decision for a committed slot is refused. The reconciliation row for H02 currently cites only the crash arm; it must cite both.

Full verdict with the 20-mutation table follows from `rev1390-consensus`; the other three axes are still running.

siy:
## Interim BLOCKING — deployment/placement axis (`e1eb4a344`)

**F1 — a definitive capacity refusal whose ledger release CAS conflicts is filed as `CREATE_UNCERTAIN` forever.** Any concurrent inventory `observeBound` (the CTM's 10 s pass is enough) makes the reservation-release CAS fail; `CapacityControlledLifecycle.java:109-118` rethrows the RELEASE's cause instead of the provider's, so the reconciler's `definitiveRefusal()` never sees the definitive refusal — the reservation is retained (a fleet slot leaked), no availability evidence is recorded, no fallback runs (H09 / H-T09 "deterministic fallback" does not happen). Probe p2 with control under `oss/internal/probes/s25-rev1390-deploy/`. No PR test can see it: none wires the reconciler to the ledger.

Required: the provider's cause is preserved across a failed release (release failure is logged/retried, the refusal still reaches `definitiveRefusal()`), and a pin that wires reconciler + ledger with a conflicting `observeBound` during the release.

MEDIUM so far (details in the verdict): the drain gate on `retirementSafe` is unpinned (0 of 1,363 tests red when removed); `BLOCKED`/`CREATE_UNCERTAIN` have no exit and make the community unremovable (`ClusterConfigRoutes.java:587`); GCP/Azure never produce `CapacityUnavailable`, so their refusals park as `CREATE_UNCERTAIN`; #1038 (auto-heal deletes even when disabled) is inherited, not fixed.

Full verdict with the mutation matrix follows from `rev1390-deploy`.

siy:
## Node assembly / authority axis (`e1eb4a344`) — BLOCKING on two mechanical items, otherwise APPROVE-WITH-NITS

Verdict `oss/internal/sweep-state/s25-rev1390-node.md`; probes `oss/internal/probes/s25-rev1390-node/`. Full `aether/node` on the build host: 1721 `<testcase>`, 1 error, 1 pre-existing skip.

**B1 — 109 consensus-storage residue files are COMMITTED** under `aether/node/data/aether/consensus/unnamed/` (`checkpoint.bin`, `owner.lock`, `voting.wal`) by `91eeff1ee` — test output from the CWD-relative fallback in `defaultConsensusDirectory`; not gitignored. Remove them, gitignore the path, and make the fallback refuse (or use a temp dir) under test.

**B2 — `WorkerRuntimeCommitWiringTest` is not hermetic**: no `HermeticStorage`, it resolves `/data/aether/storage` and trips the host's `.encryption-enabled` marker (the one error above); CI is green only because the GitHub runner has no writable `/data`. It also writes a uuid-named control dir into `aether/node/data/` per run. The changelog cites it `[verified:]` twice.

**MEDIUM (cross-axis, the most consequential): 17 of 18 hierarchy Forge classes are Heavy** — excluded from `ci.yml`, and `hierarchy-review.yml` is gated on `head_ref == feat/hierarchy-runtime-rc4` — so **after merge no CI runs any of them** on rc4/rc5 (heavy-forge nightly is main-only). H06 reassignment/replay has no unit pin; its only pin is Heavy Forge. Required: either the acceptance classes run on the target branch's CI (a hierarchy job keyed on paths, not on the PR branch name), or the reconciliation rows that rest on them say `[unverified after merge]`.

MEDIUM: `isActiveLeader` guard (M5) and the H07 stale-nonce replay defence in `CommunityHealthIndex.matches` (M9) are unpinned by the PR's tests — the replay tests are refused by the missing pending entry, never by the nonce; probes supplied, red under mutation. Immutable-role docs the PR owns are still stale: `management-api.md:4389` (write-through promotion), `ManagementApiResponses.java:1022-1046` docstrings, `AetherCli.java:955`. H-T06: the `CommunityReachability` leadership-age grace is unreachable for assigned workers (`AetherNode.java:4116-4118` folds them to `!isReachable`, zero grace); the measured bound is the 1 s challenge cadence + 20 s `communityAbsence`; the row cites the wrong test.

Held: the authority mechanism — 18 mutations, 12 red in PR tests (acceptance check, read witnesses, term continuity, expected term, client sender, worker-evidence exclusion, recovery, immutable role, peer scope); competing nominations and restart continuity red under the acceptance mutation; #1343's SWIM chain byte-identical to rc4; every new component reached; no KV write bypasses `staleWrite`.

siy:
## Metrics / metadata axis (`e1eb4a344`) — BLOCKING (two items), plus MEDIUMs

Verdict `oss/internal/sweep-state/s25-rev1390-metrics.md`; probes `oss/internal/probes/s25-rev1390-metrics/`. Green at head on the build host: metrics 268, control-attr 11, node set 59, config 434, ttm 41, Forge `HierarchyAuthorityAcceptanceTest` + `HierarchicalWorkerReconnectTest` 2/0 (on ports 41100–41899 patched in the reviewer's clone only — the branch's hard-coded 35100–36800 are unchanged). 32 mutations: 21 red with a named test, 9 unpinned.

**B1 — H08 violated on the control loop: a partial-coverage sample is ingested before the coverage check.** `ControlLoopContext.computeArtifactLoad` (aether-control `:480-486`) records the partial sum into the load-factor window BEFORE `METRICS_INCOMPLETE` is evaluated. RED probe at head: 2 of 3 producers' typed reports expire for 5 cycles (HELD / METRICS_INCOMPLETE), then return with UNCHANGED load → `SCALED_UP 3→4`, loadFactor 1.686 (predicted 1.686). "Missing, partial or old measurements are not zero load or fresh evidence" is the invariant. Guarding `recordSample` with `covered` makes it HELD 1.0 and keeps the 11 attribution tests green (proven in the clone, restored). Required: that guard + the probe as a pin.

**B2 — a worker answers cluster-wide metrics routes from a {self, leader} view.** `/api/v1/metrics`, `/metrics/history`, `/nodes/metrics` and the dashboard are LOCAL routes; a worker now holds only {self, leader} and answers them as if cluster-wide. `management-api.md:1212` still says "every node it knows, fetch ONCE"; the PR's own worker-metadata spec says forward-or-reject; the code does neither — no marker, no forward, no test. `ClusterEvent.java:195`'s comment is now false. Required: forward to a core (or refuse with a typed cause naming the scope) on a worker, pinned; docs updated.

MEDIUM: (M1) `/api/v1/nodes/metrics` reads `cpuUsage`/`heapUsedMb`, keys that never exist in `allMetrics()` (`cpu.usage`, `heap.used`) → 0/0/0 for every node — pre-existing, untested; (M2) changelog omits the wire renames `latencyP*` → `intervalMeanLatencyP*` and `errorRate` per-second → ratio; (M3) durable producer-epoch wiring/ordering unpinned (a wall-clock epoch stays green over 59 tests; `ProducerIncarnationTest` never touches `AetherNode`); (M4) pong/health incarnation is still wall-clock (`bootIncarnation = currentTimeMillis`, `AetherNode:3549/3891`) and `CommunityHealthIndex.validMembers` rejects the WHOLE community report when one member's incarnation regressed, pruned only on assignment change — the check is unpinned; (M6) H10 "priority" is lane isolation only — say so; (M7) `bytesPerSecond`, `scopesPerWorker`, `chunkBytes` bounds unpinned.

Instrument note for the other axes and for the author's own mutation script: `forge-core` is a SHADED jar — a per-module install never reaches `forge-tests` until `forge-core` is rebuilt; three live mutation runs were vacuous until then (once rebuilt, MM1 reddens H-T08 live at :158). `[unverified:]` H-T12 envelope not run; QUIC connection-window starvation; M4 by reading only.

siy:
## Deployment / placement axis (`e1eb4a344`) — final: BLOCKING (F1), full verdict

Verdict `oss/internal/sweep-state/s25-rev1390-deploy.md`; probes, harness and per-run summaries `oss/internal/probes/s25-rev1390-deploy/`. Baselines on the build host: aether-deployment 1,363 testcases 0/0/0; environment + cloud 434 0/0/0; the head's CI artifact: 50 Forge cases 0/0/0 (takeover, fallback, movement, loaded movement, drain). 18 whole-module mutations (1,369 cases each): 15 pinned by named tests; UNPINNED — M1 the drain gate on `retirementSafe` (0 reds), M7 `capacityRefused` release-proof, M8 `existingAttempt` identity-reuse refusal.

**F1 (BLOCKING, from the interim, unchanged):** definitive refusal + conflicting ledger-release CAS → `CREATE_UNCERTAIN` forever, reservation retained, no evidence, no fallback (`CapacityControlledLifecycle.java:109-118` rethrows the release's cause; reconciler `:404-415` never sees the provider's). Clear-the-block shape: decide `definitiveRefusal` from the provider cause BEFORE the release; refusal evidence + release in one guarded transaction; the three pins (F1, M7, M8); provider coverage stated in the spec.

MEDIUM: (F3) `BLOCKED` / `CREATE_UNCERTAIN` / `DRAIN_UNCERTAIN` have no exit and block community removal (`ClusterConfigRoutes.java:587`); (F4) only Hetzner and AWS map `CapacityUnavailable` — GCP/Azure refusals park as `CREATE_UNCERTAIN` (pre-existing mapping, new consequence); (F5) ready target + unsafe source parks `AWAITING_READY` silently, no deadline or escalation; (F6) a dead worker's `NodePlacementKey` is never removed and `hasCapacity` counts it forever (probe p6); (F7) #1038 inherited, not fixed (no auto-heal gate on the departed reap).

Movement-interruption matrix: rows 1, 2, 4, 5, 6, 7 tested; row 3 half (drain-before-proof untested); row 8 by reading; row 9 partly refuted ("bounded retry" for `BLOCKED` is no retry). Shared reservations "exactly one wins": confirmed by the reviewer's own concurrent probe — the PR has no concurrent-reservation test; add one. Cross-axis: `CommunityDrainCoordinator` dedup and `DeploymentMap` replay rebuild underpin `retirementSafe` (node); `CoreVoterReconciler` not reviewed here (consensus).

siy:
## Consensus axis (`e1eb4a344`) — final: BLOCKING (B1), full verdict

Verdict `oss/internal/sweep-state/s25-rev1390-consensus.md`; probes, mutation script and JSON `oss/internal/probes/s25-rev1390-consensus/`. Baseline `integrations/consensus` 853/0/0/0; 20 whole-module mutation runs (861 cases each incl. 8 probes): 16 pinned by the named tests; M6 (directory fsync) unpinnable without fs injection — the row already says so; M8/M12 equivalent by construction.

**B1 (from the interim, confirmed with the fix shape):** `recoverLocalState:1546-1549` skips checkpoint restore when the journal is empty — the state every CLEAN stop leaves. `if (journal.isEmpty() && saved.isEmpty())` verified as mutation M19: probes p1/p1b green, 851/853 shipped tests green; the two reds (`RabiaSyncAdoptionQuorumTest.ownMoreAdvancedState…`, `…futureHistory…`) assert `lastRestored() == null` on the #660 premise this PR retires and never assert phase or content — fixtures to update, not a defect re-buried. Add the clean-stop pin (restart → `processed` = the checkpoint's committed prefix; conflicting decision for a committed slot refused) and cite both arms in the H02 row.

MEDIUM: (M1) the H01 row/contract is untagged for certificate-chain trust — two workers with a forged chain activate the engine with roster {A, W, W2} (probe p2c); in-model per the spec's "not Byzantine proofs", but a tag sweep sees an unbounded claim → tag it; (M2) no operator surface for reconfiguration state although the spec promises one — reality is `LOG.warn` at 1 s in `AetherNode:2866`, zero references under `aether/api` (cross-axis with node); (M3) a torn tail from a crash mid-append refuses boot even though the torn frame was never acknowledged — spec-conformant, but say it is a design choice; (M4) worker `Propose`/`VoteRound1`/`VoteRound2` counting toward quorum is UNPINNED — the mutation reddens only the Decision test; H01 needs the vote-round pins.

Held: the f+1 boundary (`VoterConfiguration:38-40` + `PhaseData:367/377`) satisfies the `weak_mvc.ivy` axiom (`:27-28, :153-156`) for all n; no `Retry`/recursion per slot — a 5,000-slot replay took 104 ms at stack depth 20 (the #1392 shape is absent); durable vote-before-send, v1 framing corruption/version cases, rename and directory-force failures all pinned.

---
**Batch status after the four-axis review:** #1386 APPROVE-WITH-NITS (r2), #1379 APPROVE (r2), #1390 BLOCKING on five items across three axes — consensus B1, node B1/B2, deploy F1, metrics B1/B2 — plus the cross-axis MEDIUM that no CI runs the hierarchy Forge classes after merge. Next round: address the blocking items and the workflow gating, then each axis re-takes its own findings on the new head.

siy:
Addressed the four-axis review in this update. The changes preserve the crash-fault trust model and fail closed where external effects remain uncertain; they do not weaken quorum, erase durable identities, or reset recovery evidence.

**Consensus**
- Clean-stop recovery now restores a checkpoint even when the voting journal is empty. `RabiaDurableRestartTest` covers checkpoint-only and crash/WAL recovery, preserved state/frontier through another restart, and refusal of a conflicting decision for an already committed slot. The two obsolete snapshot-adoption expectations were updated to assert restored content/frontier.
- Worker proposals and both vote rounds have independent exclusion/progress controls in `RabiaHierarchySafetyTest`.
- `/api/v1/nodes/status` exposes `voterReconfiguration`: installed/target electorates, epoch/barrier, handoff stage, certified witness counts and refusal details. The core assembly/status test exercises the public response.
- Handoff checkpoint reuse and persistence-before-local-self-vote ordering were corrected and pinned. The contract now explicitly limits certificates to the crash-fault model and distinguishes lane isolation from scheduling priority. The durable-control recovery runbook documents intentional refusal on a torn WAL, including an unacknowledged tail; there is no unsafe truncation or post-barrier timeout rollback.
- The restart investigation also found that the legacy Git backup envelope was passed directly to the Base64 decoder. The adapter now strips its exact phase header; `ConsensusSnapshotEncodingTest` pins valid restoration and malformed-data refusal.

**Node assembly and authority**
- Removed the committed control-storage residue and ignored its output path. Startup requires explicitly configured durable control storage; it no longer falls back to the working directory. `WorkerRuntimeCommitWiringTest` and the affected boot fixtures use isolated storage.
- Assembled-node tests verify durable metric epochs across restart, corrupt-epoch refusal before publication, and wrong-role activation through the real subscription path. Both pong/governor health evidence and metrics now use the durable process epoch; SWIM's boot/refutation incarnation stays separate. A regressed member is excluded without discarding valid evidence for its whole community.
- Added the live-pending-nonce replay and inactive-leader grant pins. Non-leader governor requests receive an immediate negative response; announcer failures are observable, and obsolete SPOT-directive tolerance is removed. Health ingress documents that authenticated peer binding is enforced by `InboundMessageAuthority`; the index no longer pretends a payload-derived sender is independent transport evidence.
- Promotion API/CLI documentation and H-T06 attribution were corrected. Assigned-worker exclusion and community-absence recovery are distinguished from the grace for unassigned nodes. Announcer local flags are not described as authority.

**Deployment and placement**
- Definitive no-create evidence is committed independently of the contended ledger release. A `RELEASED` reservation records that evidence; the capacity counter stays conservatively charged until guarded cleanup succeeds. Active movement consumes the evidence before deletion. Periodic refusal reconciliation works in standalone #1390, without relying on #1405's inventory loop or leader turnover.
- `CapacityRefusalRecoveryTest` composes the real lifecycle/reconciler through a release conflict, restart and lost completion, and races independent controllers for the last fleet slot. Added independent retirement-proof, fallback-release-proof and unresolved-identity pins.
- The remaining operational improvements from F3/F5/F6/F7 are explicitly implemented in #1405: binding restoration preserves the interrupted phase; late readiness/evacuation proof resumes safely with an escalation deadline; provider-confirmed absence removes stale placement together with capacity accounting; disabled auto-heal holds departed deletion. #1390's reconciliation identifies this dependency rather than claiming those behaviors already exist standalone.
- The local-admission diagnostic nits remain explicit limits in the reconciliation: `NodeCapExceeded` currently produces location-scoped backoff even though it means fleet-wide pressure, and `CAPACITY_UNAVAILABLE` groups slot contention, ledger conflicts and fleet admission. These are conservative pre-dispatch deferrals, not proof of provider exhaustion. This update does not claim a new failure taxonomy or global-backoff policy.
- Provider coverage is explicit: AWS/Hetzner map definitive capacity refusals; unclassified GCP/Azure errors remain uncertain. No force-clear API is introduced for an unresolved create or missing drain acknowledgment. Reading new evidence is not permission to repeat an unknown create under another identity. Loss of authority before refusal evidence commits must remain uncertain.

**Metrics and metadata**
- Coverage is checked before any sample enters the scaling baseline. `ControlLoopCoverageRecoveryTest` is the three-producer regression with an injected clock: expired sources, five held cycles, then unchanged returning load must retain factor 1.0 and produce no scale-up.
- The three cluster metrics routes target a core. Worker dashboard connections receive an explicit `INCOMPLETE_CLUSTER_VIEW` error and close before registration or any cluster-state payload; worker publishers do not publish a scoped view as global. Route and websocket scope tests pin the boundary. The websocket refusal occurs immediately after upgrade because the existing endpoint API has no pre-upgrade predicate.
- Node metrics use the published CPU/heap keys and convert heap bytes to MiB. Raw-view expiry and expired/future typed-relay exclusion are pinned. Snapshot factories require explicit producer epoch/sequence.
- Metadata tests now pin bandwidth, server/client scope-count limits, and client chunk-size rejection using valid content. The changelog records the error-rate ratio and interval-mean latency field renames. Forecast analysis no longer manufactures unused multi-minute ratio/quantile aggregates, and resolved configuration bindings remain opaque literals, including `${`.

The hierarchy workflow is path-triggered for PRs and main/release pushes, including stacked branches. It no longer stops covering the hierarchy when the runtime PR merges. Dependency order remains #1386 → #1379 → #1390, followed by #1405's lifecycle continuation. The reconciliation and runbooks separate required safety behavior from provider, transport, resource-envelope and positional-TTM limits; no WAN or physical 10K-node performance claim is made.

Final implementation `4ae759b0a`, followed only by validation-ledger commit `7836e6802`, passed all six build/JBCT steps and the full **50-case live hierarchy matrix**: zero failures/errors/skips, 1,597 seconds, all 70 runtime dependencies fresh. Complete downstream suites passed: metrics 270, deployment 1,370, control 120, TTM 43, node 1,745 (zero failures/errors, one existing skip); configuration passed 435 cases. The initial 858-case consensus run found one new test counting initial recovery as retry; after correcting that fixture, all 44 focused consensus cases passed. The ledger explicitly distinguishes that rerun from the 857 other full-suite cases already passed.

Checkpoint-only recovery and metrics coverage both failed by assertion under targeted mutations, then passed with byte-for-byte restored sources. No build/install ran during Forge. New-head CI is pending; historical green checks do not validate this update. Independent review and merge approval remain outstanding.

siy:
## Node assembly / authority axis — round 2 (`7836e6802`): APPROVE-WITH-NITS

B1 fixed: no files under `aether/node/data` at HEAD, `.gitignore:49` covers the path, and the CWD-relative fallback is gone — boot refuses without `cluster.consensus_path` or an artifacts path, and refuses non-absolute paths. B2 fixed: `WorkerRuntimeCommitWiringTest` and the two pre-existing boot tests use `HermeticStorage`; full `aether/node` on the build host = 1745 `<testcase>`, 0 failures/errors, 1 bench skip, and afterwards `aether/node/data` does not exist and `/data/aether` is unchanged. Workflow: both branch-name gates removed, `hierarchy-review.yml` is path-keyed (pull_request any base; push to main/release-*); runtime-acceptance ran on this head (32m50s). M5/M9 now pinned by the adopted `GovernorAuthorityRaceTest` / `CommunityObservationAuthorityTest` (re-run reddens exactly those). Immutable-role docs corrected (`management-api.md`, `ManagementApiResponses` docstrings, `AetherCli:955`); H-T06 row and contract §8 now state the measured mechanism (immediate exclusion for assigned workers, 1 s challenge cadence, 20 s absence, leadership grace only for unassigned nodes). Also fixed: non-leader replies with a refusal, SPOT dead tolerance removed, index accept/acceptAuthenticated split, directive-path refusal pinned (`CoreRuntimeActivationAndStatusTest`, red under M15). 13 delta-touched mutations re-run; M6/M7/M10 inert/equivalent; M16/M17 Forge-only and now CI-executed.

Still `[unverified: forge-hierarchy-live]` by this reviewer. Verdict file `oss/internal/sweep-state/s25-rev1390-node.md` round 2. Remaining on #1390: consensus and metrics round-2 verdicts.

siy:
## Metrics / metadata axis — round 2 (`7836e6802`): APPROVE-WITH-NITS (one new MEDIUM, one doc contradiction)

Green on the build host: metrics 270, control 12 (the reviewer's probe adopted as `ControlLoopCoverageRecoveryTest`), node set 79, config 435, ttm 43; Forge `HierarchyAuthorityAcceptanceTest` 1/0 (60 s) + `HierarchicalWorkerReconnectTest` 1/0 (34 s) on re-run with shaded `forge-core` rebuilt. B1 fixed exactly as proposed and pinned (reverting the `covered` guard → red). B2 fixed: METRICS / METRICS_HISTORY / NODE_METRICS are now `AnyCoreNode` (forwarded), the dashboard refuses `INCOMPLETE_CLUSTER_VIEW`; pinned by `ManagementCompleteViewTest` + `DashboardWorkerScopeTest`; `ClusterEvent` comment corrected. M1/M2/M3/M5/M6/M7 fixed and pinned — the wall-clock-epoch, forwarded-freshness, overlapping-summary and three bound mutations that were green in round 1 each now redden a named test.

**New MEDIUM M8 (by reading):** pong/health incarnation is now the durable epoch, but `MembershipFsm.onGovernorHealthy` / `onWorkerAdmissionHealthy` (`:566/:592`) compare it against `MemberTracking.incarnation`, which is ALSO fed by SWIM's wall-clock incarnation (`onSwimHealthy`, `AetherNode:6452`; boot seed `:3565`). Governors sit in core SWIM (contract `:346`) and report themselves, so their durable-epoch evidence compares "older" than the wall-clock value forever — liveness survives via SWIM, the field's meaning does not. Fix: one incarnation source per comparison (compare durable epoch with durable epoch), pinned. `[unverified: live]`

NITs: `management-api.md:1212-1215` still says "cluster-wide despite LOCAL / no forwarding" and now contradicts the new forwarding paragraph in the same file; `collectSliceSources` freshness filter unpinned (MC4).

Instrument note (not a finding yet, log kept): the first round-2 Forge attempt failed BOTH classes at first worker admission — worker→leader QUIC dial hung 15 s per attempt, the SWIM announce never joined; each class passed alone afterwards. 1 fail in 2 runs per class on the host, cause not attributed — worth a look alongside #1398's runner failures. Verdict `s25-rev1390-metrics.md` round 2; probes `probe-r2.py`, `FORGE_r2_*`.

siy:
**Consensus axis, round 2 at `7836e6802` — APPROVE-WITH-NITS.** (Node and metrics axes already at APPROVE-WITH-NITS; deployment axis on #1405 pending.)

- Module run: 866 cases (858 shipped + 8 reviewer probes), 0 failures / 0 errors / 0 skipped.
- **B1 closed** (clean stop → restart activated an empty state machine): the fix has the M19 shape — `p1` → phase 1 `[committed]`, `p1b` drops the conflicting slot-0 Decision; the new pin `cleanStopRestoresCheckpointBeforeStalePeersCanActivateOrReplayOldSlots` (two restarts, snapshot-carrying SM) goes red when the guard is reverted (mutation M20), together with the two `RabiaSyncAdoptionQuorumTest` fixtures, which were updated to `SELF_SNAPSHOT` + phase 42 — the right side of the fixture was corrected, nothing re-buried.
- M1: `[limit: crash-fault-certificates]` now stated in the contract and the reconciliation table. M2: `VoterReconfigurationStatus` exposed on `StatusRoutes:231` (HTTP not exercised in this review). M4: three worker-ballot pins — mutation M3 now reds all four. All round-1 NITs closed, including a persist-once pin for the voter handoff. M3 (torn tail) unchanged — design call, not a blocker.
- Only M1/M3/M20 were re-run; the other 17 mutation targets are byte-identical to round 1.

**New MEDIUM R2-1 — no `stopping` refusal on `submitCommands`/`reconfigure`.** After `stop()`, `apply()` fails only via the 30 s `applyTimeout` and `reconfigure()` never resolves. rc4 PR #1341 carries exactly this fix (commit `36544f5c6`: `performStop` enqueued on the apply executor + admission closed, plus the two refusals). **When this branch takes the rc4/rc5 tip after #1341 lands, `stop()`/`safeExecute` will conflict textually: keep this branch's `ensureRecovered`/`participationStarted`/`performStop`, and take #1341's two refusals.** Resolve it in the catch-up merge rather than re-implementing here.

Verdict file `oss/internal/sweep-state/s25-rev1390-consensus.md`; probes `oss/internal/probes/s25-rev1390-consensus/`. `[unverified: forge shape — not run in this review]`.

siy:
Addressed the second-round findings in `b538479e4`, then integrated foundation `9a40851aa` through merge `8ddda6250`. Final validation is recorded separately below.

**M8 — independent membership evidence domains.** Membership now keeps separate monotonic high-water marks for SWIM incarnation and durable process epoch. Governor reports and direct worker admission share the durable fence; neither compares against SWIM's wall-clock-scale incarnation. Death captures both fences, and recovery through either domain preserves the other. Existing membership incarnation diagnostics/deltas remain explicitly SWIM-specific.

Equal or older durable evidence is rejected before healthy side effects in DEAD/DEPARTING, under the existing member transition lock. Otherwise a rejected state transition could still clear death confirmation or cancel eviction. Other states accept the current process epoch and reject older evidence.

`Membership*Test`: **344 passed**, zero failures/errors/skips. The 12 new parameterized cases exercise real FSM ingress, transitions, membership projection and JOINED/REMOVED deltas for governor and admission evidence. They cover both evidence orders, death/rejoin, departing recovery, retained fences and equal-epoch side effects. They do not start transport or prove assembled node routing. Mutating durable comparisons back to the SWIM fence made all 10 initial interleaving cases fail; weakening terminal acceptance to `>=` made both additional cases fail. Exact source restoration passed all 12.

**MC4 and forwarding documentation.** The new control regression retains an expired producer's cached metrics after its placement leaves ACTIVE, while the remaining ACTIVE producer reports fresh load. Coverage stays complete. It checks the actual view delivered to the controller, independently of the coverage guard: removing `collectSliceSources`' freshness filter changes the expected active-load component from `100/180` to `1.0` and fails the assertion. The combined control selection passed **13 cases**; the restored focused case passed. The management API documentation now states `AnyCoreNode`, worker forwarding and incomplete-scope refusal, with freshness/coverage limits.

**Acceptance workflow.** Runtime acceptance is separated into `hierarchy-runtime.yml`: PR execution requires `run-hierarchy`; manual dispatch remains available; automatic main/release pushes are restricted to hierarchy acceptance tests, validator and workflow paths. The validator checks every explicitly selected source before building and actual testcase records afterward, refusing missing, empty, partial, failed, errored or skipped coverage. Its seven regressions include nested classes and unrelated-package rejection. Foundation-only branches cannot obtain a misleading partial runtime pass.

**Instrument correction.** The saved combined run contains two different failures: reconnect's worker repeatedly stalled dialing its leader; authority's held-back core failed to bind UDP 41102 with `Address already in use`, then timed out awaiting its SyncResponse. Authority did not fail at first worker admission. Separate reruns passed in 59.99s/34.47s, but establish neither socket ownership nor the stalled-dial cause. No speculative QUIC change or timeout increase was made.

**Consensus R2-1.** Addressed R2-1 now in `aca8c8864` so this PR is ready independently of #1341 merge timing. Both admission methods synchronize with stop and refuse late requests with `NodeInactive`. The latch pin also exposed queued requests whose registration/reconfiguration was canceled before their promise was tracked. A per-request cancellation callback now completes those promises; ordinary timers/messages retain the existing inner stopping guard. This intentionally differs from the legacy #1341 drain policy and preserves this branch’s `ensureRecovered`, `participationStarted`, passive-client and final checkpoint/stop machinery.

`RabiaStopAdmissionTest` pins queued and late apply/reconfiguration, plus idempotent stop. Independently removing either queued-request refusal callback makes it fail with `Timeout` instead of `NodeInactive`; restored source passes. Full consensus: **859 tests, 0 failures/errors/skips**. Final six-step build/JBCT passed. Live runtime gate results follow in the overall validation record.

**Final merged-head validation:** Implementation `aca8c8864` passed all six build/JBCT steps with a clean tree, all seven validator regressions, 43 node health/authority/assembly tests, and the targeted live gate: **23 cases, zero failures/errors/skips**, 531 seconds, all 70 runtime dependencies fresh. The strict checker verified all ten selected classes and their actual testcase records. No install/build ran during Forge. The ledger records this separately from the earlier 50-case matrix; new-head generated-merge CI remains pending. The owner permits merging reviewed, ready changes into rc4; there is no rc5 timing prerequisite.

siy:
**Delta review at `74830210b`** (s26, adversarial, by axis). Verdict: **APPROVE-WITH-NITS**.

Scope note: three of the four s25 axes already re-took at `7836e6802`, so the genuinely unreviewed
delta is 17 files. The real gap was the **deploy axis**, which never had a round 2 and still
carried an open BLOCKING. I closed that here.

**Closed.** F1 (definitive refusal lost to a conflicting ledger release) is fixed with the exact
shape asked for: refusal evidence is persisted independently of the contended counter, and
`releaseRefusal(...).fold(_ -> cause.promise())` preserves the provider's cause through a lost CAS
(`CapacityControlledLifecycle.java:109-123`). `CapacityRefusalRecoveryTest` is a real probe — it
injects a ledger version bump and ships a no-conflict control beside it. F2 and F8's three unpinned
guards now have named tests. On the metrics axis, M8 is genuinely resolved by carrying two fields:
SWIM incarnation and durable process epoch are disjoint at every FSM state
(`MembershipContext.java:38/41`, `MembershipState.java:71-73,102-104,135-137,183-185,206-208`).
Consensus R2-1 is fixed and pinned by a test that **asserts the cause** under a 1 s timeout and runs
in 0.011 s — the "passes by 30 s timeout" defect cannot recur. I verified 859/0/0/0 on the consensus
module myself, and mutating either shutdown callback reddens exactly that pin.

**Please look at these four before the rc5 cut:**

1. **`tools/check-hierarchy-selection.py` cannot detect a partial selection** — the one thing its
   docstring, the workflow step name ("Verify complete selection before building") and the
   reconciliation doc ("partial selections refuse") all say it does. Feeding it **1 of 24 classes
   exits 0**. Its negative controls do fire, so it is not broken, just narrower than its name — but
   the final-head validation record in this very PR documents a **10-class** gate that the validator
   passed. Fix is ~4 lines: derive the expected set from `TEST_DIRECTORY.glob("Hierarch*.java")` and
   require a superset.
2. **Post-merge acceptance coverage narrowed.** `hierarchy-runtime.yml`'s release-push trigger is
   keyed only on the acceptance tests and tooling, not on `aether/**` or `integrations/consensus/**`.
   Since 17 of 18 `Hierarch*` classes are Heavy (excluded from `ci.yml`) and `heavy-forge.yml` has no
   push trigger, a production change on a release branch would get no hierarchy acceptance at all.
   This regresses the node-axis M3 finding that was closed at `7836e6802`. It is honestly disclosed
   in the reconciliation doc, so this is a decision to take knowingly — adding the production paths
   back to the `push` trigger keeps the label for PRs and costs nothing.
3. **Merge order.** A trial merge of current rc4 (`29262fbb2`) into this head conflicts in
   `RabiaEngine.java`, `QuicClusterClient.java` and `QuicClusterServer.java`. rc4 has taken #1341,
   which resolves shutdown admission the **opposite** way — it drains an admitted request where this
   head refuses it. That difference is real and is caught: removing this head's inner `stopping`
   re-check lets a reconfiguration complete after `stop()` and reddens `RabiaStopAdmissionTest:68`.
   So keep that test intact through the resolution, and gate the merge **result**, not the two PRs'
   separate greens. (Also: #1405 is based on this branch and will auto-close if it is merged and the
   branch deleted — retarget it first.)
4. **Still open from s25-deploy, unchanged:** F3 (BLOCKED/CREATE_UNCERTAIN have no operator exit and
   the spec's "operator intervention is required" is untagged prose), F4 (`definitiveRefusal` maps
   Hetzner and AWS only — GCP and Azure have **0** `CapacityUnavailable` mappings, and there is no
   `[limit: provider-mapping]`), F5 (`beginDrain`'s unsafe-retirement arm still returns a bare
   `unitPromise()` — no phase, no escalation, no deadline), F6 (`handleNodeRemoval` still does not
   remove `NodePlacementKey`, so a dead worker permanently consumes a fleet slot in `hasCapacity`).
   Each is one test or one tag; none blocks on its own.

Credit where it is due: the "no scale claims" precondition is not merely met — every scale-shaped
sentence the branch adds carries a `[limit:]`/`[unverified:]` tag or an explicit disclaimer (80
verified / 18 unverified / 31 limit tags across `aether/docs`), so a tag sweep returns the bounds
next to the claims. All 20 `[verified:]` citations resolve. The full 24-class acceptance matrix
genuinely ran at this head — I downloaded the artifact and counted 50 `<testcase>` elements, 24
classes, 0 skips, with the element count agreeing with the attribute sum.

One caveat on the validation record: `hierarchical-cluster-validation.md:340` still cites the
24-class selector as living in `hierarchy-review.yml`; a later commit on this branch moved it to
`hierarchy-runtime.yml`, leaving that file with zero hierarchy classes. And the `/private/tmp/…`
log paths cited throughout the specs resolve only on the author's machine, in a public repo.

siy:
**One unpinned refusal, found while predicting this PR's effect on rc4's existing tests.** Worth a test before merge; everything else here is a note for the merge, not a request.

`GovernorAnnouncementValue implements OwnerFenced<Epoch, NodeId>, LeaderAuthorized` makes this PR the **first production implementor of either interface** — rc4 has zero (positive control: `EpochBearing` → 14 files). That key is live: 27 rc4 production files reference it, including `GovernorAnnouncer`.

**The gap:** implementing `OwnerFenced` makes `incomingEpochIsStale` **also refuse an equal-epoch write from a different `fenceOwner`** (`KVStore.java:332` — its docstring says *"re-writes require the same governor identity"*). All five `GovernorAnnouncementFence` tests reannounce as the **same** governor, so none of them reaches that branch. A strictly new refusal rule ships unpinned.

That is where a test is worth spending. `[unverified: only those two test files were searched.]`

**Two things that are NOT problems, recorded so they don't get re-raised:**

1. **The six rc4 tests this PR touches are predicted to PASS at merge, not redden.** Both files are modified by this PR and untouched on rc4 since the merge-base, so the merge takes this PR's version verbatim. The shared `apply()` helper moves from a plain `Put` to `LeaderTransaction` + `Mutation`; **the assertions are unchanged**, and the production side converted in the same PR (`GovernorAuthority.java` is a production caller). Measured on a composed tree: exit 0, 32 tests, 0 failures. Recorded as a written prediction before the post-merge suite run, so the result gets scored rather than rationalised.

2. **`staleGovernorOlderEpoch_rejected` is not vacuous**, though a rejection test that starts passing for a new reason is exactly the thing to check. `KVStore.validMutation` routes replacements through `!staleWrite()`, which still contains `staleEpochWrite`, so the epoch fence is genuinely exercised on the `LeaderTransaction` path.

**A note for whoever scores the post-merge run:** both parent classes report `Tests run: 0` because every test is `@Nested`. Confirm the six actually ran by walking `<testcase>` elements in the surefire XML — the attribute will say zero whether they ran or not.

Issue #1388, opened 2026-09-20
Flaky consumer-runtime tests assert async state without awaiting it (DlqStreamSinkTest cursor, StreamConsumerRuntimeTest unsettled-commit count)


siy:
Two `aether/aether-stream` tests reddened once each in CI on 2026-09-20 (rc4 tip runs / PR merge refs), each a single isolated failure with a timing shape. Neither is a product regression on its face; both assert asynchronous state without awaiting it. Together with #1380 (fixed by #1384) and #1359 that is four timing-shaped flakes in the module's consumer/replication tests in one day, under CI's `-T 1C` module-parallel reactor.

1. `DlqStreamSinkTest.malformedTopicEnvelope_isQuarantinedRaw_andPartitionContinues` (added by #1306) — tip `5a5f0099c`, run 35516247198, 0.072 s:
   `[the cursor passes the garbage] expected: 2L but was: 1L` at `DlqStreamSinkTest.java:271`.
   The test polls until `delivered` is non-empty, then asserts `cursorPosition == 2` at once. The handler records into `delivered` BEFORE the runtime advances the cursor, so the assertion can run between delivery and the advance (`1` = past the garbage, not yet past the good event). Fix: await the cursor position (poll `cursorPosition == 2` with the same deadline), and keep the delivery assertion as is.

2. `StreamConsumerRuntimeTest$CursorCommitObservability.close_countsBothUnsettledCommits_whenPeriodicAndFinalCommitShareOneConsumer` — PR #1382's merge ref (`8d0edccd3` + tip `edcf8f4c0`), run 35515972207, 5.064 s:
   `[two distinct unsettled commits for one consumer at the bound must count as two incidents, not one] expected: 2L but was: 1L` at `StreamConsumerRuntimeTest.java:1738`.
   Elapsed ≈ the 5 s bound: one of the two commits was not yet registered as unsettled when the bound fired, or had settled. Read the producer (`ConsumerRuntimeState` periodic/final commit, the #1355/#1370 slot change) to name which, and pin the interleaving through the existing seams (a held store) rather than the wall clock.

Rule for the fix (not a timeout widening): each test must await the condition it asserts, or force the interleaving deterministically through a seam; state for each which one and why. Run each class ×20 locally after the fix and quote the count.

siy:
[...]

Issue #1355, opened 2026-09-20
ConsumerRuntimeState.issueCheckpoint assigns the periodic-commit predecessor after the async dispatch, so a detach flush can overlap the in-flight commit (#1239 race)


siy:
Found by s1335fix in the #1335 fix round (session 25); written up in `oss/internal/sweep-state/s25-1335fix-report.md` §Findings F1. Pre-existing since `834fb1e2f` (#1239); the CTO ruled it its own ticket.

`aether/aether-stream/src/main/java/org/pragmatica/aether/stream/ConsumerRuntimeState.java` (at `4c0154de8`):
- `:618-627` `issueCheckpoint`: `state.periodicCommit(observedCommit(key, state, NO_PREDECESSOR).timeout(...).onResult(...))` — the predecessor ref is assigned AFTER `observedCommit` returns;
- `:711` `issueTrackedCommit`: `predecessor.fold(_ -> lifted(() -> commitCursor(...)))` — `lifted` is `Result.lift(...).async().flatMap(...)`, so the store call is already dispatched on another thread before `observedCommit` returns;
- `:69` `NO_PREDECESSOR = Promise.success(persisted())` and `:1287` `periodicCommitRef = NO_PREDECESSOR` — the initial value is already settled;
- `:553-558` `flushCursorForKey`: `observedCommit(key, state, state.periodicCommit())` — the detach flush waits on whatever the ref holds at that instant.

**Race:** periodic commit A reaches the store before the issuing thread has stored A's promise; a detach flush B arriving in that gap reads a settled predecessor and is issued at once, overlapping A — exactly the overlap #1239's serialisation exists to prevent. With #1239's per-commit outcomes the outcomes are not misattributed; the cost is the overlap itself (a flush racing a periodic write of an equal-or-older cursor).

**Evidence:** `StreamConsumerRuntimeClusterCursorTest.detachFlush_waitsForTheInFlightPeriodicCommit_andEachCommitReportsOnlyItsOwnOutcome` (`aether/node/.../StreamConsumerRuntimeClusterCursorTest.java:368`): "expected: 1, but was: 2" — 3/3 green isolated, 1 red in a full `aether/node` run under host contention (`oss/internal/probes/s25-1335/full-node.log`).

[...]

PR #1393, opened 2026-09-20
fix(stream): register a cursor commit before its store call; await the cursor in the DLQ partition test (#1388)

siy:
Closes #1388

Rule applied: each test awaits the condition it asserts or forces the interleaving through a seam — **never a wider timeout**. Both mechanisms were read from the producer (`ConsumerRuntimeState`) and each is pinned deterministically.

## 1. `DlqStreamSinkTest.malformedTopicEnvelope_isQuarantinedRaw_andPartitionContinues` — await

Producer: `deliverSingleEvent` runs the handler (which fills `delivered`) in `invokeHandler` and advances the cursor in `deliverySucceeded` only after the handler's promise settles. The dead-letter hold orders the quarantine entry and the advance past the garbage *before* the good event's delivery; only the advance to `2` is unordered relative to the delivery poll. Fix: poll `cursorPosition == 2` under the **same** 3 s deadline; the delivery/quarantine assertions are unchanged.

Evidence: with a 100 ms stall placed before `advanceCursor`, the unmodified test fails `[the cursor passes the garbage] expected: 2L but was: 1L` (CI's signature); the awaiting test passes under the same stall.

## 2. `StreamConsumerRuntimeTest$CursorCommitObservability.close_countsBothUnsettledCommits_…` — a product window, pinned through the store seam

Producer: `issueTrackedCommit` called `store.commit(...)` inline (the test's `commitsIssued` latch fires *inside* it) and only then minted the `TrackedCommit` and did `inFlightCommits.add`; `issueCheckpoint` arms the periodic 5 s `.timeout()` after that returns. One stall of the issuing thread in that window gives both halves of CI's failure: `close()` (woken by the latch) snapshots `inFlightCommits` without the periodic commit — so it is neither bound-awaited nor counted for this shutdown — and the periodic's own timeout is armed later than the shutdown bound, so the bound counts only the final commit → `1` at ≈5 s. No existing seam observes registration (the #1355 probe points are both before the store call), so a test-only await did not exist.

Fix: the `TrackedCommit` carries a handle minted and registered **before** the store call, settled from the chain via `withResult` (an immediate completion). The outcome handlers stay attached to the chain itself: attached to a chain a synchronously-failing store has already settled they run inline, which `close_countsFailure_whenFinalCommitFailsSynchronously_andDoesNotWaitOutTheBound` pins — the first attempt put them on the pending handle and that test went red (`Promise.processActions` dispatches handlers on a pending promise to `AsyncExecutor`).

New pin `close_countsAPeriodicCommitWhoseStoreCallIsStillInProgress_asUnsettled`: the store parks *inside* `commit()` (a held store) until the test releases it after `close()` has returned, so the count can be 2 only if registration preceded the call. The existing wall-clock sibling is unchanged apart from a comment: its latch fires inside the store call, which the registration now precedes, so it is a sufficient condition.

Evidence (base `c7c977e1a`, local, offline; build host down):
- RED: new pin at the base — `Tests run: 13, Failures: 1`, `… :1783 […] expected: 2L but was: 1L`. Mechanism probe: the base with a 200 ms stall between the store call and the registration (periodic path only) fails the *existing* test 3/3 with `but was: 1L` at `:1738`; the fix under the same stall placed after the call passes 3/3.
- GREEN: `CursorCommitObservability` 13/13, `DlqStreamSinkTest` 8/8.
- ×20 loop of `StreamConsumerRuntimeTest*` + `DlqStreamSinkTest`: 20/20, `Tests run: 68, Failures: 0` each, XML 60+8 per run, 0 failures/errors.
- Module `-pl aether/aether-stream test`: `Tests run: 1150, Failures: 0, Errors: 0, Skipped: 0`; surefire dir wiped first, XML `<testcase>` 1150, `<skipped>` 0.
- Gate `jbct:check -Djbct.skip=false -pl aether/aether-stream`: `Running JBCT check on 98 Java file(s)` … `JBCT check passed.` (98: the tip added `DurableSealedOffsetSource.java` since `866c05b3b`.)
- [unverified: CI's post-fix rate under `-T 1C` — the deterministic pins are the evidence.]

Changelog fragment: `changelog.d/1388-await-async-state.md`.


siy:
Ruling follow-up (head `db4e7063c`): the held-store pin now also reads the new package-private `ConsumerRuntimeState.inFlightCommitCount()` from INSIDE `commit()` and asserts it is already 1 — the commit is registered before its own store call. Reversed order (registration moved back after the call) reddens it 3/3 with `but was: 0`; green with the fix. Both classes 68/68; gate 98 files passed. Changelog fragment states the product change: a cursor commit is counted in flight from before its store call.

siy:
rev1393 M1 adopted (head `06938e8ca`): `CursorCommitObservability.pendingCommitsSettledAfterCloseBegan_closeReturnsWhenTheyDo_countsNothing`, from `oss/internal/probes/s25-rev1393/Rev1393ProbeTest.java` (credited in the test doc). Pins that the tracked handle settles with its chain: K2 (the `withResult` forwarding deleted) reddens exactly this test 3/3 (`close() returned once the store settled, well inside the 5 s bound`), the other 13 stay green; at head 14/14. Gate 98 files passed (local, offline).

Issue #1401, opened 2026-09-21
Flaky: CursorCommitObservability.unsubscribe_thenResubscribe_resumesFromCommittedCursor_noRedelivery on the rc4 tip — committed offset not yet one past the last delivered


siy:
rc4 tip `93a5e3089`'s combination-gate run 35545733848 (2026-09-21 ~00:00Z) failed `build-and-test` on one test:

```
StreamConsumerRuntimeTest$CursorCommitObservability.unsubscribe_thenResubscribe_resumesFromCommittedCursor_noRedelivery
  assertion: "committed offset is one past the last delivered offset"
```
(aether-stream: 1 failure / 1182 run.) The previous tip `8d04025e6` — which already contained #1335 (fenced consumer-group assignment, `28a28043a`… no: #1393 `28a28043a` commit registration) and #1393 — passed the same gate; `93a5e3089` adds only #1343 (SWIM start failure, `aether/node` + `aether/ember`). So this is timing-shaped, not a deterministic merge result: the seventh timing-dependent signature in the consumer-runtime test family this cycle (#1380, #1388 ×2, this).

Shape: after `unsubscribe` then `resubscribe`, the committed offset read is not yet one past the last delivered offset — either the final commit on unsubscribe (the detach flush, #1355/#1370 territory) had not settled when the assertion ran, or the resubscribe read a cursor committed under the previous fenced assignment epoch (#1335 changed `fetchCursor` to require the assignment epoch) and the assertion observed the older record. Discriminate from the producer; do not arbitrate.

Rule for the fix (as for #1388): the test awaits the condition it asserts or forces the interleaving through a seam (the held store / `inFlightCommitCount()` seams from #1393 exist) — never a wider timeout; if the mechanism is a product window (a resubscribe that can read a stale cursor before the detach flush lands), fix the product and pin it. Run the class ×20 locally after the fix and quote the count.

siy:
Fixed by #1402 (with #1393 for #1388's first two), squashed into `release-1.0.0-rc4` as `d444d22c6` (rev1402 APPROVE, `oss/internal/sweep-state/s25-rev1402.md`; the unsettled-commit incident count is now recorded via with*() in attachment order before the handle settles). Product residual: #1403. Closed manually — non-default branch.

PR #1402, opened 2026-09-21
test(stream): await the cursor advance before detaching in the resubscribe test (#1401)

siy:
Closes #1401
Part of #1388

Rule applied: the test awaits the condition it asserts — **never a wider timeout**. No product change.

## Discrimination, from the producer

CI run 35545733848 (attempt 1): `[committed offset is one past the last delivered offset] expected: 1L but was: 0L`, elapsed 0.011 s. The store stub sets `committed` synchronously inside `commit()` and resolves inline, so the detach flush **had** settled — it committed 0. And the fenced `fetchCursor` (#1335) is not on the path of the assertion that failed (it reads the stub's own `commit` argument). So neither ticket candidate.

The mechanism: the handler (which counts `firstLatch` down) runs in `invokeHandler`; the cursor advance runs in `deliverySucceeded`, after the handler's promise settles; `unsubscribe` → `cleanupConsumer` → `flushCursorForKey` commits the cursor **as it stands**. The test thread, woken by the latch, could detach before the advance and the flush committed 0. Same producer shape as #1388's `DlqStreamSinkTest` case.

## Fix

The property under test is the flush of an **accounted** delivery, so the test awaits the runtime's own accounting — `cursorPosition("orders", 0, "group-1") == 1` — before `unsubscribe`, under the same 5 s deadline the delivery latch uses. Assertions unchanged.

Recorded, not changed: an `unsubscribe` issued while a delivery's handler has returned but its advance has not run commits the pre-delivery cursor; the late advance's checkpoint bails on the cancelled consumer, so the event is redelivered on reattach. That is the runtime's at-least-once contract at detach, not a stale-read window — flagged for the CTO in the report; no product change here.

## Evidence (build host `cto24-build-1`, `-pl aether/aether-stream`, base `93a5e3089`)

- Mechanism probe: a 100 ms stall placed before `advanceCursor` in `deliverySucceeded` — the unmodified test fails `expected: 1L but was: 0L` (CI's signature); the awaiting test passes under the same stall.
- GREEN: `CursorCommitObservability` 14/14.
- ×20 loop of `StreamConsumerRuntimeTest*`, one Maven invocation each: 20/20 `Tests run: 67, Failures: 0, Errors: 0, Skipped: 0`, XML `<testcase>` 67 per run, 0 failures/errors.
- Module: `Tests run: 1182, Failures: 0, Errors: 0, Skipped: 0`; XML 1182, `<skipped>` 0 (the report dir holds only this run's files).
- Gate `jbct:check -Djbct.skip=false -pl aether/aether-stream`: `Running JBCT check on 99 Java file(s)` … `JBCT check passed.`
- [unverified: CI's post-fix rate under `-T 1C` — the stall probe is the evidence.]

## Scope +1 (head `10c9bbaf0`): `close_countsBothUnsettledCommits_whenPeriodicAndFinalCommitShareOneConsumer` rewritten on the #1393 seams

This PR's own CI (run 35548028994, head `278a4dfe9`, which contains #1393's registration fix) reddened it once more: `expected: 2L but was: 1L`, 5.063 s. The log names the interleaving: `Cursor commit unsettled at the 5,000ms shutdown bound` (the **final**) at 00:46:06.1993, then `Cursor commit (local) failed … Promise timed out after 5000ms` (the **periodic**) at 06.1996. The store stub returned a pending promise, so the periodic commit's own 5 s `.timeout()` was armed at issue — before `close()` — and raced the shutdown bound. It fired first: the bound rightly skipped the already-settled periodic handle and reported the final alone, and the periodic's increment arrived on an **async Promise event handler** (`onFailure` on a pending promise dispatches to `AsyncExecutor`) 0.3 ms later — after `close()` returned and after the test had read 1. The missing commit was the periodic; the product counted it, asynchronously.

Rewrite: the periodic's store call is **held** inside `commit()` until after `close()` returns, so its timeout does not exist before the bound — the arming race is impossible, not unlikely. The stub reads `inFlightCommitCount()` on entry (== 1: registered before the store call); while `close()` waits, the test awaits `inFlightCommitCount() == 2` (the final registered beside the held periodic) and confirms `close()` has not returned; `close()` returns at the bound with the count read on the closing thread == 2; then the store is released, the periodic failed, and the count must stay 2 (per-commit token). The #1393 held-store pin is unchanged. `committingSequence` (only used by the old shape) is removed. Mutation: the bound reporting only the first in-flight commit reddens this test and the #1393 pin (2/14). Class ×20 → 20/20 (67 tests each), module 1182/0/0/0, gate 99 files — all at `10c9bbaf0`.

## Product change (head `44f76010f`, owner's hint): the incident is counted in the frame that settles the handle

`issueTrackedCommit` recorded the outcome with `onSuccess`/`onFailure`, which `Promise.processActions` dispatches to the executor (`runEventHandlers(events)` runs before `runAll(joins, runAll(actions))`), while the handle settled inline via `withResult`. That is the CI interleaving above from the product's side: a handle settled by its own timeout δ before the bound, an increment on an async event after the caller's read. Now `withSuccess`/`withFailure` record the outcome, attached BEFORE the unregister and the handle's `withResult` — `with*()` actions run inline in the resolver's frame in attachment order, so nothing that sees the handle settled (the bound's `allOf` join above all) can see the count short. Stated in a comment at the attachment site.

Pin `commitFails_incidentIsCountedBeforeItsHandleSettles_observedFromTheSameResolveFrame`: `close()` is waiting on a pending periodic and its chained final; the test fails the periodic's store promise on its own thread; the runtime's `with*()` actions run inline in that frame, then (timeout dependency → periodic slot → the final's `fold`) the FINAL commit's store call — which the stub uses as a same-frame observer of `cursorCommitFailureCount()`: 1, no wait, and `inFlightCommitCount()` 1. The final fails synchronously, so `close()` returns inside the bound with both counted.
- Mutation MA (increment back on `onFailure`): the observer reads `0` — 10/10 red with the pin alone, 12/13 overall (the one green ran inside the full class with hot executor threads: the async handler is dispatched before the inline actions run, so an inline observer races it and cannot exclude it — no test can make the async variant lose deterministically).
- Mutation MB (handle's `withResult` attached before the count, both inline): green — the only observer past the handle's settle is `close()`'s join waking a thread, a nanosecond window no test can close. The order is documented at the site; `[unverified]`.
- Class ×20 at `44f76010f`: 20/20 (68 tests each, XML-counted); module 1183/0/0/0, XML 1183, 0 skipped; gate 99 files passed (one blank-line reformat first).

Changelog fragment: `changelog.d/1401-resubscribe-cursor-await.md`.


siy:
Head `7914b1448`: added the CI-shape test `periodicCommitTimedOutByItsBoundWhileCloseWaits_bothIncidentsAreCountedAsOfCloseReturning` (periodic failed by its own bound while `close()` waits; final fails synchronously; count read on the closing thread with no wait == 2). Green deterministically with the with*() change. It does NOT redden with the increment back on `onFailure` (0/10 alone): the async handlers are dispatched before the inline frame that unparks the closing thread, so the executor wins — `commitFails_incidentIsCountedBeforeItsHandleSettles_observedFromTheSameResolveFrame` (observer inside the resolve frame, 10/10 red under that mutation) is the discriminating pin. Class ×20 → 20/20 (69 each), module 1184/0/0/0, gate 99 files.

Issue #1404, opened 2026-09-21
Cluster lifecycle: reconcile durable provisioning, retirement and capacity state without relying on lost edges or leader churn


siy:
Found during the membership/provisioning design review of rc4 and the hierarchy runtime PR #1390 (`e1eb4a3`). This is a code/spec review, not a new cloud reproduction. Owner requested a separate implementation PR starting from rc4. The hierarchy runtime and guarded KV foundations (#1390, #1379; specs #1386) are prerequisites for the complete change.

## Problem

Membership and provisioning do not yet form a consistently recoverable control system. Every unmet intent or unfinished operation must remain discoverable from durable state: events accelerate reconciliation, but losing an event must not prevent eventual progress.

### 1. Core replacement progress disagrees with durable capacity accounting

`LeaderReconciler.seedInFlightFromRetainedDispatched` reconstructs local progress from metrics gossip, not committed reservations. `evictInFlightPastCeiling` forgets an unresolved attempt and reopens the deficit, while `CapacityControlledLifecycle` correctly retains its allocation after ambiguous create failure. A lost dispatch observation or repeated uncertain attempts can exhaust the cap without restoring target capacity. Retaining uncertainty is correct; losing the operation explaining it is not.

### 2. Core cleanup depends on another edge

`ClusterTopologyManagerRecord.scheduleActivationReplay` performs a one-shot two-observation sweep. Listing failures wait for the next leadership activation. `terminateDeparted` logs failed termination without durable retry state; deferred reaps eventually wait for another SWIM failure. A stable leader and transient provider failure can leave an orphan indefinitely.

### 3. Capacity accounting lacks continuing reconciliation

`CapacityControlledLifecycle.ensureInventory` initializes a ledger once. Fleet listings adopt observed instances, but missing allocations are released only through node-specific queries. Lost workflow ownership or out-of-band deletion can leave allocations charged indefinitely. Inventory completion must be scoped to source configuration, and continuing inventory must distinguish unconfirmed creates from previously observed resources whose absence can be confirmed.

### 4. Legacy workers use provider headcount

`ClusterTopologyManagerRecord.applyWorkerDelta` computes desired minus inventory size, omitting pending reservations and readiness. A delayed listing can cause repeated creates; a VM which exists but never joins can indefinitely satisfy the target. Prefer normalizing implicit worker topology into the community operation model over preserving two provisioning algorithms. Breaking changes are allowed; no migration is required before GA.

[...]

Issue #1328, opened 2026-09-19
CloudProviderSupport eagerly fans out entire batches and rebuilds shared ProvisionSpec per node


siy:
## Summary

CloudProviderSupport rebuilds the same group-level ProvisionSpec for each node and eagerly launches all asynchronous operations in a batch. Termination/status paths have analogous fan-out. Large batches can create an avoidable burst of provider requests; provider throttling and memory impact have not been measured in this review.

## Evidence

Source: [aether/environment-integration/src/main/java/org/pragmatica/aether/environment/CloudProviderSupport.java:22](https://github.com/pragmaticalabs/pragmatica/blob/039830ef6/aether/environment-integration/src/main/java/org/pragmatica/aether/environment/CloudProviderSupport.java#L22). Found in the third external review (outside streaming, durable pub/sub and durable entities) at `9d782a186`; the reviewed file is unchanged at `039830ef6`. Source inspection, not a runtime reproduction; no project tests or benchmarks ran.

## Fix direction

Validate/build group-level provisioning inputs once, while retaining any genuinely per-node identity/context. Bound outstanding provision/terminate/status operations using explicit Promise composition. Define how successful partial results remain available for reconciliation when another operation fails; do not hide already-created resources behind a single opaque batch failure.

## Acceptance

- An instrumented provider shows no more than the configured number of in-flight calls for a large batch.
- Group-level invalid configuration causes zero provider calls.
- Every provision retains its intended node identity and placement; do not regress #1027.
- Mixed success/failure preserves enough successful resource identity for reconciliation and reports typed failures.
- Record before/after spec-build count and peak in-flight requests.

## Related

#1027 concerns dropped node identity. This optimization must preserve its fix and is not evidence that all partially successful batches currently orphan resources.

Issue #1251, opened 2026-09-19
Deterministic cluster supervision and predictive capacity planning — executable runbooks, continuous reactive scaling, calendar-aware preparation (rc5)


siy:
## What

A single supervision mechanism covering three things:
- **Executable operational runbooks:** bounded, verified recovery, with escalation when a remedy is exhausted or unsafe.
- **Continuously active reactive scaling.**
- **Calendar-aware predictive capacity preparation:** intraday to year-plus horizons, moving events such as Black Friday, regional calendars, and declared campaigns.

Prediction supplies anticipated requirements. Existing reconcilers execute them, through one fenced admission authority and the existing budget and reporting channels.

**Spec:** `aether/docs/specs/cluster-supervision-spec.md`. It defines 11 invariants, typed contracts, a deterministic forecast v1, a durable action protocol, the management API/CLI/dashboard, phases P0–P7 and a 32-test acceptance matrix. Spec PR: #1252.

## Owner rulings (2026-09-19)

- The design decisions in spec §1 were made with the owner and are **confirmed**. They are requirements, not open alternatives.
- **Target: 1.0.0-rc5.** The feature is UX- and operator-facing.
- **Start condition:** the spec is good to go once the prerequisites below are met.

## Prerequisites (gating; the spec defers to these)

| Needed by | Prerequisite |
|---|---|
| §11.1 incident channel on `system:cluster-events` | **#1230**: that stream is appended on non-owner nodes today (multi-writer) |
| §11.1 durable outbox, stream-backed events | the stream correctness set **#1231–#1239** (ordering, WAL recovery, drop-as-success, sealing, visibility, publish outcomes, consumer delivery) |
| §6.1 conditional admission and reservations | **#1250**: submitter-visible rejection for fenced KV writes, plus an atomic multi-key command |
| §7.4 / SUP-R03/R04 replacement adoption; INV-02 one effect owner | **#1038**: auto-heal deletes the VM even when disabled. The supervisor cannot adopt or hold replacements if destruction ignores the gate |
| §2.1 10K+ nodes, multi-region in one logical cluster | scale validation **#365 / #366 / #367**. 10K workers is the architecture's own shape, ~100 nodes per community × ~100 communities (`architecture/08-scaling.md`). **Both** factors are design targets pending validation: the single-community cap (#365/#366) and how many communities one core coordinates (#367). The spec requires a topology capability report and refuses unsupported modes |

## Related, and how they relate

[...]

Issue #1250, opened 2026-09-19
Consensus KV: fenced writes need a submitter-visible rejection and an atomic multi-key command (prerequisite for fenced admission)


siy:
## Summary

The consensus KV applier already fences writes, but two capabilities that fenced admission needs are missing:

1. **A rejected fenced write is not reported to its submitter as a rejection.**
2. **There is no atomic multi-key command.**

The first is needed by the cluster-supervision admission protocol (spec §6.1). The second is needed for a reservation set spanning several capacity pools. Both are prerequisites for that rc5 work (see the epic linked below). They are also the missing pieces behind two rulings made this session. One is the #1243 projection-claims backing (a consensus-KV compare-and-set). The other is the #1239 cursor finding that the runtime could not express a conditional cursor write.

All references are at `release-1.0.0-rc4` tip `ccba0dba5`. `KV/` = `integrations/cluster/src/main/java/org/pragmatica/cluster/state/kvstore/`.

## What already exists (read, not run)

`KV/KVStore.java` `staleWrite` (~`:92-103`) applies four deterministic fences inside the applier, all pure functions of committed state plus the incoming value:

| Fence | Rejects | Origin |
|---|---|---|
| leader view | a stale `LeaderKey` write | H4 |
| `EpochBearing` | a stale-epoch write | #345 piece 1a |
| `VersionFenced` | a non-successor write, i.e. single-record compare-and-set by version | RFC-0018, #570 |
| `MonotonicFenced` | a regressive write | #700 |

`KV/KVCommand.java` has `Put`, `Get`, `Noop` (a linearizable barrier) and `Remove(key, witness)`, which is witness-fenced (#379).

**So single-record conditional writes exist, as a property of the value type rather than of the command.** This issue is NOT "add compare-and-set".

## Gap 1: rejection is silent to the submitter

`handlePut` (`KV/KVStore.java:80-90`):
- **Stale write:** it returns `Option.option(storage.get(key))`, the *current* value, mutates nothing and emits no notification.
- **Applied write:** it returns the *previous* value.

Both outcomes have the same type. **[unverified]** I have not traced how, or whether, the submitting node's `Promise` distinguishes them. Establishing that is step one.

The admission protocol requires **explicit rejection**: a compare-and-set whose loser cannot tell it lost is not usable as a fence.

## Gap 2: no atomic multi-key command

Supervision admission must atomically commit one or more of:
- a pool target together with its reservations;
- a cross-pool reservation set;
[...]

PR #1405, opened 2026-09-21
fix(cluster): reconcile durable lifecycle work without lost-edge stalls (#1404)

siy:
## Problem and behavior

Cluster lifecycle work could remain stuck after a missed event, provider failure or leadership change. This change makes provisioning, inventory and retirement reconcile from committed intent and evidence on periodic passes.

Core replacement counts durable pending allocations; deletion records retirement before provider effects and retries only after current safety checks. Worker scaling uses the community operation protocol. Independent communities and inventory sources progress in bounded lanes. Confirmed provider absence releases accounting and stale placement atomically; uncertain creates remain reserved.

The review corrections preserve previous-worker allocation exclusion during delayed readiness, escalate delayed evacuation, resume safely after restoration of a recorded source binding, and retain definitive no-create evidence across accounting conflicts and lost callbacks. Regression tests pin retirement proof, disabled auto-heal, existing identities, last-slot admission and recovery across controller restart.

Closes #1404. Related: #1038, #1250, #1251, #1328, #241, #234. This does not close the physical-scale acceptance gates #367/#368/#599.

## Stack and implementation contract

**Base: `feat/hierarchy-runtime-rc4` (#1390).** This is the lifecycle follow-up to the hierarchy specification/foundation/runtime stack (#1386, #1379, #1390); merge it after its runtime prerequisite. Runtime review head and validation ledger `74830210b` are incorporated. The lifecycle contract is `aether/docs/specs/cluster-lifecycle-reconciliation.md`.

Pre-GA changes require no migration. The existing `RETIRING` and `READINESS_DELAYED` wire additions require matching cluster artifacts. Timeouts bound observation, not external provider effects; genuinely ambiguous creates and missing drain acknowledgements still require matching evidence. Definitive cloud-capacity mapping currently covers AWS/Hetzner; no 10K-node or physical-WAN performance claim is made.

## Validation

- Final six-step isolated build passed after integrating runtime membership-domain and shutdown-admission fixes.
- Integrated focused regressions: 57 passed (consensus 1, deployment 48, control 2, node 6), zero failures/errors/skips. Earlier broad integration: 679 passed before the consensus-only follow-up and lifecycle test corrections.
- Review mutations prove delayed-readiness exclusion, disabled auto-heal deletion, existing-identity refusal and stale-placement admission checks. Latest F6 mutation: one intended assertion failure; restored affected suites: 36 passed.
- **Executable head `be02ea807`: 30 live Forge cases passed**, zero failures/errors/skips, 619 seconds, all 70 runtime dependencies fresh. The strict checker verified all ten selected classes: formation, slice invocation, declarative consumers, core restart/resize, worker formation/drain, capacity fallback, community movement and leader takeover.
- The first freshness preflight refused an artifact older than byte-restored mutation source; no cluster started. Reinstalling the unchanged deployment artifact resolved the preflight. No builds/installs ran during live tests.

The subsequent runtime `74830210b` merge and validation ledger update change documentation only. The committed ledger records exact selectors and tested revisions; generated-merge CI remains independently required. Historical results are not attributed to the revised executable head.


siy:
## Review (`52798dad3`, lifecycle delta over #1390's `e1eb4a344`) — BLOCKING

Verdict `oss/internal/sweep-state/s25-rev1405.md`; probes, harness and per-run summaries `oss/internal/probes/s25-rev1405/`. Reviewed by the deployment-axis reviewer of #1390 against its own findings, then against this delta's claims. Counts on the build host: aether-deployment 1,385 / 0 failures (the body says 1,378); node 1,722 with the same `WorkerRuntimeCommitWiringTest` non-hermetic error as #1390 (shared `/data/aether/storage`; the delta touches neither).

**G1 — new regression: `READINESS_DELAYED` drops the previous worker's retirement exclusion.** `CommunityRetirementIndex`'s prepared set does not include the new phase (file untouched by the delta), so entering `READINESS_DELAYED` removes the exclusion; on resume, `beginDrain` waits on `retirementSafe` for a node that is no longer being evacuated → a silent permanent park. Probe p7. The author's resume test cannot see it: its fixture holds `retirementSafe` constant true and has no index. Fix: add the phase to the prepared set, plus a pin that ENUMERATES every phase against the index (so the next phase cannot fall out either).

**Against #1390's deployment findings:** F1 (definitive refusal vs release CAS → `CREATE_UNCERTAIN` forever) — UNTOUCHED, probe p2 still reproduces, and `cluster-lifecycle-reconciliation.md:33` now overclaims it; F3 (no exit from BLOCKED/…) — PARTIAL: only readiness-expiry `BLOCKED` became resumable, a restored binding stays `BLOCKED` (p3); F5 (silent `AWAITING_READY`) — UNTOUCHED (p4); F6 (dead worker's `NodePlacementKey`) — UNTOUCHED: the ledger side is now released by the continuing inventory, the placement key is still counted (p6); F7 (#1038) — closed by reading, UNPINNED: removing the auto-heal gate reddens 0 of 1,385 — do not close #1038 on this evidence; M1/M7/M8 still unpinned.

**The delta's own claims hold:** wire baseline diff exactly two enum lines with `UNKNOWN` last; bounded lanes (p8, pinned — a stalled item still delays its lane by the full 30 s deadline, state that); periodic progress pinned; `RETIRING`-before-delete pinned; durable-capacity union pinned.

**CI:** the hierarchy acceptance job is SKIPPED on this head — `hierarchy-review.yml` gates on `head_ref == feat/hierarchy-runtime-rc4`, so a stacked branch gets no live acceptance and the 30-test Forge claim is local-only (same finding as the node axis on #1390: key the job on paths, not the branch name).

Next round: G1 + F1 + F6 + the F7/M1/M7/M8 pins here (this is the deployment axis's closure PR), then a re-take on the new head.

siy:
Addressed the lifecycle review in `b5f9b527c`. The related runtime repairs and validation ledger from #1390 are incorporated; the distinction below is deliberate so local evidence is not confused with final merged-head acceptance.

### Correctness and recovery

- **G1 — retirement exclusion:** `READINESS_DELAYED` is now in the prepared/protected set in `CommunityRetirementIndex`. The switch and its parameterized regression enumerate every phase explicitly. A worker being evacuated cannot become an allocation target merely because replacement readiness missed its deadline.
- **F5 — ready target but unsafe evacuation:** retirement proof still gates drain. Once the drain deadline passes, the operation enters `READINESS_DELAYED` and emits an escalation identifying the missing evacuation proof. It retains the previous-worker exclusion and resumes only after the safety/readiness/authority checks pass. `readyDestinationCannotDrainBeforeEvacuationAndDelayedProofEscalates` covers the refusal, escalation and subsequent recovery.
- **F1/G2 — definitive refusal lost on accounting conflict:** the provider's definitive no-create outcome is persisted as `RELEASED` evidence using a reservation-only, leader-guarded transaction. This is distinct from decrementing the shared fleet counter. An active movement retains that evidence until its availability/`COMPLETE` transaction consumes it. `CREATE_REQUESTED` and `CREATE_UNCERTAIN` can recover from that matching durable evidence. Periodic accounting cleanup then removes the reservation and decrements the counter; a conflicting release leaves both the evidence and conservative charge for another pass. A failed release cannot replace the provider's definitive refusal with an ambiguous-create diagnosis. #1390 has the standalone periodic refusal-cleanup mechanism; this branch includes it in continuing inventory.
- **Lost callback:** `successorConsumesDurableRefusalEvenWhenCreateCompletionWasLost` persists the request, drives the real capacity lifecycle through refusal, omits the reconciler completion callback, creates a fresh lifecycle/reconciler over the same store, and verifies completion without another create. `definitiveRefusalSurvivesConflictingReleaseAndControllerRestart` combines the real lifecycle and reconciler, forces ledger conflict during cleanup, then verifies cleanup after restart. Loss of authority *before* any refusal evidence commits remains conservatively uncertain; an uncommitted external response cannot safely authorize a new allocation.
- **F6 — stale worker placement:** source-bound confirmation that a previously observed allocation is absent now removes its `NodePlacement` in the same guarded transaction as reservation release and ledger decrement. With a ledger present, fleet admission counts committed allocations rather than orphaned placement records. Membership disappearance alone does not establish cloud absence. `confirmedProviderAbsenceRemovesStalePlacementAlongsideReservation` pins the cleanup.
- **F3 — safe resumption:** temporary source-binding unavailability retains the exact interrupted phase while escalating the reason. Restoring the recorded binding resumes that phase; it does not redirect an operation to another account. `restartCannotRebindRecordedCreateToChangedProviderIdentity` now tests both refusal while changed and successful resumption after restoration, preserving the operation identity. Existing `BLOCKED` operations may resume through matching target observations; readiness, assignment, evacuation and acknowledgement guards still apply.

### Requested regression pins

- **M1:** a ready target cannot start drain without retirement proof.
- **M7:** `refusalCannotFallbackWhileReservationStillClaimsPossibleCreate` refuses availability/fallback while the reservation still records `DISPATCHED`.
- **M8:** `unresolvedIdentityCannotBeReusedEvenWithSpareFleetCapacity` proves the existing-identity refusal with spare capacity, rather than accidentally relying on a full fleet.
- **Concurrent allocation:** `competingReservationsForLastSlot_exactlyOneWinsViaLedgerCas` interleaves two independent lifecycle instances at the reservation commit and observes exactly one provider create.
- **F7/G1b:** `disabledAutoHealHoldsDepartedDeletion` drives the real departed-node reap with auto-heal disabled and proves no provider deletion.

I also ran the three targeted sensitivity checks requested for the new guards:

| Mutation | Observed assertion failure |
|---|---|
| Move `READINESS_DELAYED` from protected to unprotected retirement phases | `CommunityRetirementIndexTest.everyPhaseExplicitlyPinsRetirementProtection[11]`: expected protection, got false |
| Remove the auto-heal check from departed deletion | `ClusterTopologyManagerActuatorTest.DepartedReapLivenessRecheck.disabledAutoHealHoldsDepartedDeletion`: unexpected deletion of `node-d` |
| Make `existingAttempt` unconditionally succeed | `CapacityControlledLifecycleTest.unresolvedIdentityCannotBeReusedEvenWithSpareFleetCapacity`: generic admission refusal instead of the explicit unresolved-identity refusal |

Each mutation compiled and failed through an assertion, not a build error or missing-test selection. Sources were restored byte-for-byte in `finally`; all 15 selected cases then passed. The identity mutation still encounters the downstream reservation CAS, so that particular pin establishes the explicit identity-refusal contract, not that it is the sole duplicate-create barrier.

### Scope and documentation dispositions

- **G3:** the upstream hierarchy workflow correction removes the branch-name gate and includes relevant release-branch pushes, so a stacked lifecycle branch and the merged release receive acceptance coverage. Final-head CI remains required; the local mutation runs are not Forge evidence.
- **G5:** the specification and fragment now say explicitly that a stalled item delays later work in the same lane by up to the thirty-second observation deadline. Other lanes remain independent; timeout does not cancel an already-issued external effect.
- The legacy `requestRetirement` compatibility seam is marked for isolated actuator fixtures; normalized community intent owns production retirement.
- The specification explicitly states that disabling auto-heal also holds retries of already-`RETIRING` core allocations, including an earlier planned retirement whose deletion failed.
- The initial-inventory pause after restart/configuration change is documented: creates defer while the complete inventory is rebuilt, bounded by the ten-minute observation deadline.
- **F3 remaining limit:** there is no force-clear/cancel API for genuinely ambiguous creates or missing drain acknowledgements. Restore matching inventory/drain evidence; neither elapsed time nor a policy edit authorizes destructive resolution. This is now an explicit operator-resolution limit rather than an implied API.
- **F4 remaining limit:** definitive cloud-capacity mapping currently covers AWS and Hetzner. Generic GCP/Azure failures remain uncertain; particularly for Azure, an allocation failure can leave a resource, so broad error-string classification would not establish no-create evidence. The provider coverage limit is explicit.
- Local admission backpressure still shares one typed pre-dispatch failure family for source-slot contention, reservation-version conflict and fleet capacity. It means no provider create was dispatched, not proof of cloud location exhaustion. The distinct `NodeCapExceeded` location-evidence nit remains a documented limit: a local cap can still produce per-location backoff evidence. Neither that diagnostic classification nor the shared admission-failure taxonomy is claimed fixed; the upstream review response records both explicitly.

### Validation performed on this lifecycle change

- Initial focused deployment suite: **155 tests, 0 failures/errors/skips**, including the auto-heal gate.
- Final focused deployment suite after the lost-callback test: **72 tests, 0 failures/errors/skips**.
- JBCT module process: **0 errors**; existing non-fatal warnings remain.
- Three assertion-sensitive mutations, then **15 restored tests passing**.
- `git diff --check` passed.

These are focused local results for the lifecycle fixes. Final integrated results are recorded below; submitted-head CI remains independently required. No 10K-node or physical-WAN validation is claimed.

Integration update: merged reviewed runtime head `4ae759b0a` as `4040d0e95`. The merged lifecycle uses one inventory tick (which includes refusal cleanup), retains the broader lifecycle phase handling, and exposes the shared refusal API without duplicate implementations. The full six-step isolated build passed, followed by **333 merged-head focused tests** (292 deployment, 41 node), with zero failures/errors/skips. This includes both assembled-core activation/status tests, all four worker assembly tests and the wire-assignment tripwire. The worktree has no tracked drift. After the docs-only upstream validation merge at `b26b54c0e`, lifecycle Forge acceptance passed **30 live cases, zero failures/errors/skips, in 620 seconds**. All 70 runtime dependencies were fresh. The selector covers cluster formation, invocation, declarative consumption, core restart/resize, worker formation/drain, fallback, movement and takeover. No builds or installs ran during Forge. The only subsequent change records this evidence; submitted-head CI remains independently required.

Final lifecycle head: `dbc60bd4f`; its last commit only records the passing validation above.

siy:
**Deployment/placement axis, round 2 at `dbc60bd4f` — APPROVE-WITH-NITS.**

- **G1 closed**: `READINESS_DELAYED` is in the prepared set; removing it is a compile error (exhaustive switch); mutation R5b (`-> false`) reds `everyPhaseExplicitlyPinsRetirementProtection[11]` plus the reviewer's g1/f5 probes.
- **F1 closed**: `RELEASED` is written on the reservation key BEFORE the counter; the provider cause is rethrown regardless of the decrement; `refusedTarget()` resumes fallback from `CREATE_REQUESTED`/`UNCERTAIN`; `reconcileRefusals` reclaims on a later pass. Reviewer probe f1 (conflict injected on the first decrement → refused/retained → reclaimed next pass) green; R6 reds four named tests + f1. Spec line 33 is now true.
- **F3 closed** for the binding change (phase kept, resumes on restore; R8 pinned) and for changed-assignment / unresolved-refusal BLOCKED (exits on the observed target); `DRAIN_UNCERTAIN` late-ack only, by spec. **F5 closed**: drainTimeout → `READINESS_DELAYED` + one escalation, exclusion kept (R7; M1 pinned by the same test). **F6 closed**: ledger-based `hasCapacity`, stale placement removed on absence. **F7 closed and pinned** (R4 `disabledAutoHealHoldsDepartedDeletion`). M7/M8 pinned (R2/R3). Concurrent-reservation test present.
- Counts: deployment 1,406 (0 F/E/S); node 1,746 (0 F/E). The hierarchy CI job is now path-triggered and RAN on this head: 50 Forge cases, 0 F/E/S, including movement/fallback/takeover/loaded/core-restart (G3 closed).

**NITs — the first one must be fixed before this lands, the others may ride the catch-up merge:**
1. `ReconciliationBatchTest.lostCompletionIsBoundedAndCannotStopHealthyLane` is flaky (2 failures in 11 unrelated runs): the `failed` list is written from the timeout thread without synchronisation. A flaky pin in the merge unit will red the batch's own combination gate; use a concurrent collection or synchronise the write.
2. F6's "not counted" half is unpinned (R9: 0 reds) — add the pin.
3. F4 (GCP/Azure refusal mapping) untouched — out of scope, tracked separately.

Verdict file `oss/internal/sweep-state/s25-rev1405.md`; artifacts `oss/internal/probes/s25-rev1405/round2/`. With this, all four review axes of the batch (#1386 spec, #1379 primitives, #1390 node/metrics/consensus, #1405 deployment) are at APPROVE or APPROVE-WITH-NITS; landing order and target (rc5) as ruled on #1390.

siy:
Addressed the round-2 nits in `be02ea807`, with runtime membership and stop-admission corrections integrated through `0ad24bb5f`.

1. **Timeout fixture:** verified that reviewed `dbc60bd4f` already used `CopyOnWriteArrayList` for both `failed` and `visited`. The remaining ordering gap was observing the failure list after recovered-lane completion without waiting for the failure observer itself. `lostCompletionIsBoundedAndCannotStopHealthyLane` now explicitly awaits a promise completed after the failure callback records its item. There is no sleep or increased timeout.
2. **F6 admission pin:** `stalePlacementsDoNotConsumeFleetCapacityWhenLedgerExists` retains two unrelated stale placement records, initializes the ledger, admits two nodes with a fleet limit of three, and drives the real reconciler through reservation and create. It verifies the stale records remain, so passing cannot be explained by their cleanup. Removing the ledger-presence filter in `hasCapacity` fails the named assertion (1 case, 1 assertion failure, 0 errors). Exact production source restored afterward; both affected classes passed all 36 cases.
3. **F4:** GCP/Azure definitive-refusal classification remains the separately tracked scope; this follow-up makes no new provider guarantee.

Validation: isolated six-step build passed after the final consensus merge. Integrated focused tests passed 57 cases (consensus 1, deployment 48, control 2, node 6), zero failures/errors/skips. Earlier membership/control/lifecycle/node gate passed 679 cases before the consensus-only follow-up. Logs: `/private/tmp/lifecycle-round2-final-build.log`, `/private/tmp/lifecycle-round2-final-focused.log`, `/private/tmp/lifecycle-round2-f6-mutation.log`, `/private/tmp/lifecycle-round2-restored.log`.

Final executable head `be02ea807` passed the ten-class lifecycle/cluster smoke selector: **30 live cases, zero failures/errors/skips, 619 seconds**, all 70 runtime dependencies fresh. The strict checker verified all ten selected classes executed. Log: `/private/tmp/lifecycle-round2-forge.log`. The first preflight refused the deployment artifact timestamp after byte-restored mutation; no cluster ran, the unchanged module was reinstalled, and the gate then passed. No installs ran during live tests. Subsequent integration of runtime `74830210b` and the validation ledger is documentation-only. CI remains independent evidence for the published revision.

siy:
Reviewed at `976a388f886e46908f0c1b13197638f6afe4ccef` (pinned via `git ls-remote`). This is the
delta review over the session-25 round-2 head `dbc60bd4f`; #1405's own content is
`74830210b..976a388f8` (the base tip is an ancestor of the head).

**The code in this delta is good.** `be02ea807` is test-only (+30/−1) and closes both items the
previous round left open. I mutated each and both are genuinely pinned:

| Mutation | Test that goes RED |
|---|---|
| delete the ledger-existence filter in `hasCapacity` | `stalePlacementsDoNotConsumeFleetCapacityWhenLedgerExists` |
| delete `.onFailure(cause -> failure.accept(item, cause))` in `ReconciliationBatch.attempt` | `lostCompletionIsBoundedAndCannotStopHealthyLane` |

The new capacity test is a real probe, not a restatement of the behaviour: it seeds KV commands and
drives the actual reconciler, and it asserts the two stale placement keys *remain present*, so it
cannot pass by deleting them. Baseline at this head: **1,419 `<testcase>`, 0 failures / 0 errors /
0 skips** (element count and `tests="N"` sum agree). The live acceptance run's artefact
`pr-head-sha.txt` reads this exact head with **50 cases, 0 F/E/S**. Of six merge commits, five are
clean; `4040d0e95`'s resolution (promoting `reconcileRefusals` to `@Override public` for the new
interface default) is correct and drops no hunk from either parent.

**What blocks is the evidence chain, not the diff.**

Exactly one `CI` run has ever executed on this branch: run `35570264594` at `52798dad3`, created
`2026-09-21T06:51:24Z` — **34 minutes before the base was retargeted** at `07:25:00Z`. It is 22
commits behind this head and predates `b5f9b527c`, the commit that closed the previous round's
blocking findings. `ci.yml` and `changelog.yml` both trigger on
`pull_request: branches: [main, release-*]`, and this PR's base is `feat/hierarchy-runtime-rc4`.
Consequences on the current head:

- no full-reactor build, and **no module unit tests** — the `runtime-acceptance` job does run
  `./build.sh`, but its install step is `-DskipTests`, so this PR's own suite is never executed by CI;
- no `Changelog fragment` check;
- the only other check, CodeRabbit, reports `success` while its description reads
  *"Review skipped: reviews are disabled for this base branch"* (visible via
  `gh api repos/.../commits/<sha>/status`, not via the PR rollup).

The spec added in this PR says *"CI must validate the submitted merge revision independently."* On
this base that cannot happen. Suggested fix: retarget to `release-1.0.0-rc4` (which also avoids the
stacked-PR auto-close below), or `workflow_dispatch` `CI` on this head and link the run.

**Merge order.** #1405's base *is* #1390's branch, so it cannot reach rc4 except through #1390, and
#1390 is currently `CONFLICTING` against rc4. #1405 is the only PR based on
`feat/hierarchy-runtime-rc4`, so **if that branch is deleted when #1390 merges, GitHub will close
#1405 automatically.** Retarget first (`gh pr edit 1405 --base release-1.0.0-rc4`), then merge.

**Non-blocking, worth a follow-up:**

1. `hasCapacity` suppresses placement-key counting on ledger **existence**, while
   `CapacityControlledLifecycle` gates on the same ledger's **`inventoryComplete`**. During the
   post-config-commit inventory window the fleet cap stops counting placements while the ledger
   itself says it does not yet know what is allocated. `inventoryComplete` has 6 production uses and
   **0 test references** anywhere in the tree.
2. Most of `hasCapacity` is unpinned. I ran six mutations; four reddened nothing, and three of those
   are in this one method — deleting the `CapacityReservationKey` counting, deleting the
   active-operation-target counting, and ignoring placement keys unconditionally each leave all
   **1,419** tests green (all compiled; these are real results, not no-ops). The only fleet-capacity
   behaviour any test pins is the one this PR added: *"do not count placement keys when a ledger
   exists."* `CommunityPlacementReconcilerTest` is the only fixture that can reach the private
   `hasCapacity`, and it never seeds a `CapacityReservationKey` at all.
3. A fleet-cap refusal is completely silent (`if (!hasCapacity(...)) return Promise.unitPromise();`)
   — no log, no escalation. All five `ReconciliationBatch` failure observers are likewise
   `log.warn` and nothing else, so an item failing every pass forever produces only repeated WARNs.
4. The changelog fragment does not mention that placement keys no longer count toward `maxNodes` —
   an operator-visible change, and the one this PR added a test for.
5. Inherited from #1390: `hierarchical-cluster-validation.md:340` cites *"the 24-class selector in
   `.github/workflows/hierarchy-review.yml`"*, but that file now holds no selector — the 24 classes
   moved to `hierarchy-runtime.yml`. Also, the acceptance job is now **label-gated** on
   `run-hierarchy` (this PR has the label, so it ran), and the release-branch `push` trigger no
   longer lists `aether/**`.

siy:
**Update from the CTO, after this review:** I have retargeted this PR onto `release-1.0.0-rc4` (`gh pr edit 1405 --base release-1.0.0-rc4`), which addresses B1 directly — the workflows gate on `branches: [main, release-*]`, so CI now runs here for the first time since `52798dad3`. It also removes the auto-close hazard: this was the only PR based on `feat/hierarchy-runtime-rc4`, so deleting that branch on #1390's merge would have closed it.

Note what the retarget revealed: this PR reported `MERGEABLE` before and `CONFLICTING` after. That was not a change in its content — it was measured against #1390's branch, which itself conflicts with rc4, so the stacking hid the conflicts this PR inherits. Among them is the `WorkerDeploymentManager` modify/delete, now settled by an owner ruling: **the class is deleted**, but only once the #1184 fix it carries (`9324ace76`, merged into rc4 today) is ported into the replacement path with a test that reddens if the failure cause or its `PERMANENT` flag is dropped. Deleting it as-is would silently reopen a closed blocking ticket, and the test that would catch it dies in the same change.

siy:
## HOLD — do not merge this PR. Merging it merges #1390.

**Measured, not inferred:** `git merge-base --is-ancestor <#1390 head> 976a388f8` returns **true**.
This PR's head **contains all of #1390**.

Against the current rc4 tip (`be6639e7c`) this PR delivers **494 files, 96 commits,
+35,366/-6,457** — `changedFiles` reports 494, and #1390 itself is 488. This PR is not "#1405"; it is
"#1390 plus #1405".

### Why that is not visible on this page

Two things conceal it, and they compound:

1. **The base was retargeted to `release-1.0.0-rc4`.** That was correct and necessary — this PR was
   stacked on `feat/hierarchy-runtime-rc4`, and deleting or merging that base would have
   auto-closed this PR, a failure already on record here. But retargeting does not rebase: the
   commits from the old base came with it, so the retarget converted a 100-file PR stacked on #1390
   into a 494-file PR that *is* #1390.
2. **`gh pr view --json files` returns at most 100 entries.** Read as a file count it says "100",
   understating the real figure by a factor of five. **`changedFiles` is the count; `files` is a
   page.** I made exactly that misread before checking, which is why it is written down here.

### Why it matters

The owner's cut plan is: finish rc4 → run the full integration suite as a **baseline** → merge #1390
→ run the suite again, with the **merge result** as the gate. Merging this PR would put #1390 into
rc4 **before the baseline run**, and would apply #1390's conflict resolutions implicitly rather than
deliberately. Those resolutions are pre-decided and recorded as `know:` commits on
`docs/know-s26-rulings` — the `WorkerDeploymentManager` delete, the `RabiaEngine` union, and the QUIC
`OutboundMessageLimit` alias. A merge that resolves them by accident discards those rulings silently.

The 17 conflicts against the current tip are almost all #1390's, not this PR's: `LeaderAuthorized`,
`KVStoreAuthorizedMutationTest`, the `WorkerDeploymentManager` modify/delete, `RabiaEngine`,
`QuicClusterServer`, `SystemTags`.

### Also still true: this PR has never had a full-reactor CI run

Per the adversarial review at `976a388f8`: exactly **one** `CI` run has ever executed on this branch
(`35570264594`, head `52798dad3`, 2026-09-21T06:51Z), it is **22 commits behind** the head, and it
**predates** the round-2 fix commit that closed the previous round's BLOCKING finding. `ci.yml` and
`changelog.yml` trigger on `pull_request: branches: [main, release-*]`, which the old base did not
match. So `build-and-test` and `fragment` have never run on any head carrying the current fixes.

The base now *does* match, so a fresh push would trigger them — but `gh pr update-branch` currently
fails on the 17 conflicts above, and resolving those means resolving #1390's merge, which is
deliberately sequenced after the cut.

**Correct order, unchanged:** cut rc4 → baseline suite → merge #1390 with its recorded resolutions →
suite again → then bring this PR's own ~100 files forward against the result and let CI run on them.

Issue #1164, opened 2026-09-14
SWIM gossip 'daily' derived-key rotation happens at BOOT only: a node up ≥2 days and a node booted ≥2 days later cannot decrypt each other (accept window is prev/current/next day)


siy:
**What.** `SwimGossipEncryptors.fromCertificateProvider` derives the gossip key for the boot day once, in `AetherNode`/`Main`; nothing re-derives at UTC midnight. The accept window is previous/current/next day only. So a node that has been up for ≥2 days still encrypts under its boot-day key, and a node booted ≥2 days later derives a key the first node's window does not contain — **they cannot decrypt each other's gossip**. SECURITY.md:61 "daily key rotation with overlap" therefore overclaims: rotation is per-boot, not daily.

**Why it matters.** Any cluster older than two days that adds or replaces a node (auto-heal, scale-up, rolling restart spread over days) partitions its SWIM gossip between the old and the new derivation day. The symptom would look like a join that never becomes MEMBER, or a replacement that is never seen — a membership defect with a cryptographic cause, hard to attribute in the field.

**Found by** b-silent while premise-checking #683 (`oss/internal/fix-683-report-2026-09-14.md`, draft finding). Not #683's scope (that ticket is the KV-driven rotation producer); a KV rotation (#683's wire) would incidentally re-sync every node onto one key, but the derived scheme must be correct on its own.

**Fix shape.** Re-derive at UTC midnight (a scheduled re-derivation in the encryptor with the same prev/current/next overlap), pinned with a controllable clock: boot on day D, advance to D+2, a peer booted at D+2 must decrypt; mutation (no re-derivation) reds. Correct SECURITY.md to say what is earned and by what mechanism.

siy:
Fixed by #1414, squash-merged into `release-1.0.0-rc4` as `fabb4ce2e` (CTO session 26, 2026-09-22).

The merged PR is the reviewed fix **plus** a fix round closing two review findings:
- the **rebuilt** accept window is now pinned — mutating the rebuild path to drop the next-day key, then the previous-day key, previously left all 7 tests green; both mutations now redden `rolledOverNode_stillAcceptsPreviousAndNextDay` on distinct assertions with distinct key ids.
- the "clocks agree to within one day" guarantee is requalified at all **four** sites (a fourth — the test class's own javadoc — was found by grepping the phrase rather than working the review's three-site list) and cites #1415.

[...]

Issue #1200, opened 2026-09-14
OWNER DECISION: a joining node has no path to the current gossip key after a rotation — a rotated cluster cannot grow


siy:
**OWNER DECISION REQUIRED — every available option moves a guarantee.**

After one `GossipKeyRotationKey` rotation, the cluster's SWIM accept set excludes the `cluster_secret`-derived key. A later-booting node speaks only the derived key, so SWIM fails **in both directions**; the QUIC dial set is SWIM-fed and self-only, so quorum never forms; so the consensus replay that would deliver the key never runs. **The cycle closes.**

**Auto-heal replacements and scale-up are included.** So an emergency rotation — performed precisely because something was compromised — leaves the cluster **unable to self-heal**.

**Where (symbols at `561ffdda1`):** `GossipKeyRotationHandler.applyRotation`, `SwimGossipEncryptors.fromCertificateProvider`, `AetherNode.createGossipEncryptor`, `TopologyObserver` dial set.

**Why this is not fixable inside #683:** #683's scope is the rotation *producer*. Closing this needs a key-delivery path for a **pre-consensus** joiner, and the options each move a different guarantee:

- **permanently accept the derived key** — guts #683's purpose; a leaked `cluster_secret` decrypts gossip forever;
- **serve the key over the cert-authenticated QUIC/TLS channel** — creates the second trust root the §5.8 design explicitly declined;
- **require out-of-band re-provisioning of new nodes post-rotation** — changes the operational contract and needs a gossip-key config path that does not exist.

**Evidence:** `GossipKeyRotationBootDivergenceTest` (both directions, with a positive control); `TopologyObserverTest.SwimOnlyDialSet` (the dial set is pinned self-only, not merely documented); the deadlock stated outright at `AetherNode:1694-5`; `NettySwimTransport` drop-on-decrypt-failure with no plaintext fallback.

**Mitigation already shipped in #1172** (not a fix): `GossipKeyDivergenceGuard` refuses the boot **where the divergence is detectable** — i.e. for restarted members, whose peers still hold them in a configured seed set and therefore probe them. **It is blind for the auto-heal case**, because nobody probes an address they do not yet know, so the stranded node receives nothing at all.

[...]

Issue #683, opened 2026-08-27
GossipKeyRotationKey has no production writer or CLI/admin trigger — emergency gossip-key rotation cannot be invoked


siy:
## Two distinct gossip-key mechanisms exist — only one is reachable

**1. Automatic daily rotation (real, working, unaffected by this issue).** SECURITY.md:61
claims "AES-256-GCM, daily key rotation with overlap" — verified true. Every node
independently derives its current/previous/next gossip key via HKDF from `cluster_secret`
plus the UTC epoch day (`SelfSignedCertificateProvider.deriveGossipKeyWithLabel`,
`integrations/net/tcp/src/main/java/org/pragmatica/net/tcp/security/SelfSignedCertificateProvider.java:140-166`;
equivalent cloud providers delegate the same shape). No consensus/KV propagation is needed
because every node computes the identical key from the same secret + calendar day. Wired at
`AetherNode.createGossipEncryptor` (`AetherNode.java:5006`) →
`SwimGossipEncryptors.fromCertificateProvider` (`aether/node/.../SwimGossipEncryptors.java:33-36`).
[mechanism: `SelfSignedCertificateProvider.java:140-166`]

**2. A second, KV-delivered rotation path — declared, consumer-wired, producer-less.**
`AetherKey.GossipKeyRotationKey` / `AetherValue.GossipKeyRotationValue` carry an arbitrary
Base64-encoded key pair (NOT derived from `cluster_secret` — genuinely independent key
material). `AetherNode.java:2654` subscribes
`.onPut(AetherKey.GossipKeyRotationKey.class, gossipKeyRotationHandler::onGossipKeyRotationPut)`,
and the surrounding design comment calls this **"the sole delivery path"** for gossip-key
rotation, with idempotent replay-on-late-join. `GossipKeyRotationHandler.applyRotation`
(`aether/node/.../GossipKeyRotationHandler.java`) decodes the value and calls
`RotatingGossipEncryptor.rotate(...)` on the live encoder — purely reactive to whatever `Put`
arrives.

**The gap:** nothing in production ever writes that `Put`. Searched the full non-test
codebase (`aether/`, `integrations/`) for any writer, scheduler, or trigger of
`GossipKeyRotationKey` / `GossipKeyRotationValue` — the only construction site is test code
(`KVStoreSerializer.java:1072`, itself on a dead serialization path with zero production
callers). There is also no CLI command (checked `aether/cli`) and no Management API route
[...]

Issue #256, opened 2026-06-10
Gossip-key day-rollover locks out next-day joiners (UTC-midnight boundary)


siy:
Found 2026-06-10 during the topology-overhaul Wave-2 gate investigation.

`SelfSignedCertificateProvider.createProvider` derives the gossip key as `HKDF(clusterSecret, "aether-gossip-key-v" + LocalDate.now().toEpochDay())` with a 2-key window (today + yesterday). Consequence: a node booted on UTC day N+1 joining a cluster whose nodes booted on day N encrypts SWIM datagrams with the day-(N+1) key, which day-N nodes do not hold — every ANNOUNCE/probe from the joiner is rejected ("Failed to decrypt gossip ... Unknown gossip key ID") and it can never join SWIM until the older nodes restart (or a KV gossip-key rotation lands). The reverse direction decrypts fine (joiner holds day-N as its previous key), making the failure asymmetric and confusing to diagnose.

The KV `GossipKeyRotationKey` mechanism would converge keys cluster-wide, but no producer ran in any observed cluster (zero "Gossip key rotation received" across all nodes), and until commit (pending) late joiners did not replay the rotation value from synced KV either.

Mitigations landed/pending in the overhaul branch: late-joiner replay of `GossipKeyRotationKey` on Rabia state restore (RabiaEngine.onStateRestored seam). Remaining design work: either widen the static-key window / make the day-label rotation producer actually run, or decouple the SWIM key epoch from wall-clock days entirely (key agreement at join, or KV-distributed key as the sole source after formation).

Related observation, same session: a node that boots, syncs, and then loses quorum visibility while never having been a member self-drains with reason=QUORUM_LOSS and exits — a never-member probably should not self-drain (minor, separate decision).

siy:
Already fixed in rc2 by a71343c2a — gossip-key accept window widened to a 3-key (prev/current/next) set across the UTC-midnight boundary. Closing: stale tracker.

PR #1414, opened 2026-09-21
fix: re-derive the SWIM gossip day key on day change instead of at boot only (#1164)

siy:
Closes #1164

## What was wrong
`SelfSignedCertificateProvider` derived the previous/current/next-day gossip keys once, in its constructor, and `SwimGossipEncryptors.fromCertificateProvider` snapshotted them into a fixed `AesGcmGossipEncryptor`. Nothing re-derived at midnight. With a one-day accept window either side, a node up ≥2 days (still on its boot-day key) and a node booted ≥2 days later hold no key in common, so SWIM silently partitions them — the auto-heal / scale-up / rolling-restart case from the ticket.

## Fix (minimum, two hunks)
- **Producer** — `SelfSignedCertificateProvider.keysForToday`: the three keys are cached per calendar day and re-derived when `LocalDate.now(clock)` differs from the cached day. A `java.time.Clock` seam is added (`selfSignedCertificateProvider(secret, clock)`); the one-argument factory passes `Clock.systemDefaultZone()`, so the day label is byte-identical to the former `LocalDate.now()`. Derivation and labels are unchanged.
- **Consumer** — `SwimGossipEncryptors.ProviderKeyedGossipEncryptor`: every encrypt/decrypt reads the provider's current key id; when it differs from the delegate's the delegate is rebuilt from current/previous/next and CAS-swapped. It is the delegate *inside* `RotatingGossipEncryptor`, so a KV rotation (#683) still replaces it wholesale — #1200's question (what guarantee the gossip key provides) is untouched.

**Guarantee:** two nodes whose clocks agree to within one day can always decrypt each other's gossip regardless of uptime, because each encrypts under its own current day's key and accepts the previous and next day's, and their current days differ by at most one. The clock-skew bound is unchanged: two days apart is still rejected.

`SECURITY.md` names this mechanism where it previously asserted "daily key rotation with overlap".

## Evidence (all on the build host, foreground; logs in `oss/internal/probes/s25-1164/`)
- **RED, pure base `d444d22c6`** (stub-provider arm, compiles against the base API): `DerivedGossipKeyDayRolloverTest$StubProvider.encryptorFollowsTheProvidersCurrentKey` — `decrypt returned: Unknown gossip key ID: 102` (2 run, 1 failed).
- **RED, base + clock seam only** (seam is behaviour-preserving; the real-provider tests cannot compile without it): net/tcp `SelfSignedCertificateProviderTest` 23 run / 3 failed (`keysFollowTheClockDay_notTheBootDay`, `oneDayLater_windowSlidesByOneDay`, `dayBoundary_isTheClockZonesMidnight`); node `DerivedGossipKeyDayRolloverTest` 7 run / 3 failed — `nodeUpTwoDays_andNodeBootedTwoDaysLater_decryptEachOther`: `Unknown gossip key ID: -1186632462`; `longRunningNode_encryptsUnderTodaysKey`: `Unknown gossip key ID: -191801386`; plus the stub test. The four controls (`twoDayClockSkew_isStillRejected`, `oneDayApart_decryptEachOther`, `kvRotation_isNotUndoneByDayRollover`, `unchangedProvider_keepsTheSameDelegate`) were green on the base, as predicted.
- **GREEN** at `3d88f7034`: net/tcp 23/0, node 7/0 (targeted).
- **Mutation table** (restored by content after each; `git status --porcelain` empty):

  | Probe | Hunk reverted | Red (named) |
  |---|---|---|
  | M1 | provider never re-derives (`keysForToday` always returns the cached day) | net/tcp 3: `keysFollowTheClockDay_notTheBootDay`, `oneDayLater_windowSlidesByOneDay`, `dayBoundary_isTheClockZonesMidnight`; node 2: `RealProvider.nodeUpTwoDays_…`, `RealProvider.longRunningNode_…` |
  | M2 | encryptor snapshots (`current()` never re-reads the provider) | node 3: `RealProvider.nodeUpTwoDays_…`, `RealProvider.longRunningNode_…`, `StubProvider.encryptorFollowsTheProvidersCurrentKey`; net/tcp 23/0 green |

  Each hunk reddens a distinct named set, neither a superset of the other.
- **Module-full** (marker-filtered XML, `<testcase` elements, 0 stale): `integrations/net/tcp install` **75 / 0 / 0 / 0 skipped** (78 annotations − 3 in `ServerIT`, which has no failsafe binding, pre-existing); `aether/node test` **1624 / 0 / 0 / 1 skipped** = `ObservabilityStep0BenchTest.step0_overhead_bench` (pre-existing bench, untouched file); attribute sum 1624 agrees; all 215 `*Test` classes have a report (the three without are two `*IT` and `WireAssignmentBaselineWriter`).
- **Gates** (`jbct:check -Djbct.skip=false`, main, CI-equivalent): `aether/node` **`Running JBCT check on 159 Java file(s)` — `JBCT check passed`, 0 format / 0 lint**. `integrations/net/tcp` **`Running JBCT check on 16 Java file(s)`**: 4 format issues / 33 lint errors vs base-file control 4 / 32 — none of the 4 format issues is in my file; the +1 lint is `JBCT-EX-01` on `deriveDayKeys throws Exception`, the same rule the file's ten other helpers already trip (module style; net/tcp is CI-exempt via the inherited `jbct.skip=true`). Stated rather than hidden; happy to Result-wrap on request.

## Limits
- `[unverified: the day label is the system default zone's calendar day, not UTC as the #256 comments say — pre-existing (`LocalDate.now()`), deliberately not changed because the fix must not alter the derivation. Two hosts in different zones can disagree on the day near local midnight; the one-day window covers a one-day disagreement.]`
- `[unverified: the three provider reads in a rebuild are not one atomic snapshot; a day change landing between them yields a one-day-short window for the datagrams handled before the next read, which rebuilds it. Documented in the class doc; not locked around.]`
- No multi-node run — the defect and fix are at the encryptor boundary and the tests build both sides through the production factory.


siy:
Adversarial review at `3d88f7034` (own clone, build host key `rev1414`). **The fix is correct and
its primary axis is genuinely pinned.** Independently reproduced: the RED on the pure base in a
second clone at `d444d22c6` carrying your test file byte-identical — `2 run / 1 failed`,
`Unknown gossip key ID: 102`, with the control passing; and the compile error that forces the
two-arm RED is real (`required: byte[], found: byte[],java.time.Clock`), so that was not a
convenience. I removed the rebuild guard and `unchangedProvider_keepsTheSameDelegate` reddened, so
that instrument is real. Both `RealProvider` scenarios build each side through the production
factory rather than hand-feeding a key, so they can falsify the premise — I looked for the fixture
tell specifically and it is not there. `aether/node` gate re-run: `Running JBCT check on 159 Java
file(s)` → `JBCT check passed`, 0 format / 0 lint. Module-full re-run: **1624 / 0 / 0 / 1 skipped**,
`<testcase>` elements from surefire+failsafe under a run marker, 0 stale — matches your number
exactly. Two things to fold in before merge.

**1. The rebuilt accept window is unpinned — and it governs exactly this ticket's population.**
Every current test exercises a window built at *boot*. A node that has rolled over once — every
node up longer than a day, which is the point of #1164 — runs on the **rebuilt** delegate from
`SwimGossipEncryptors.java:123-127`. I mutated the rebuild path to drop the next-day key, and then
separately the previous-day key: **all 7 tests stayed green both times.** That silently un-does
#256's midnight overlap for long-running nodes with a green suite. The code is right; the pin is
missing. A ~15-line test that is green at head and reds on each mutation (on distinct assertions)
is in `oss/internal/probes/s26-rev1414/` as `p6_rebuiltWindowStillAcceptsPreviousAndNextDay` —
please drop it into `DerivedGossipKeyDayRolloverTest.RealProvider`.

**2. `SECURITY.md` ships the guarantee without the qualifier #1415 supplies.** `SECURITY.md:144`,
`SwimGossipEncryptors.java:31-33` and the changelog all say "nodes whose **clocks agree to within
one day** always decrypt each other". Two nodes whose clocks agree *exactly* but whose host zones
differ fail to decrypt in **both** directions — measured through the production factory:
`Unknown gossip key ID: -1186632462` / `-191801386`, with a same-harness positive control
(Warsaw/LA, one day apart) succeeding. The code behind that is #1415 and I am not re-opening it;
the ask is only that the shipped wording say **"UTC clocks"** and cite #1415, since #1164's
deliverable was to stop SECURITY.md overclaiming. Two smaller ones in the same breath: the
changelog's `[unverified:]` says zones "disagree near midnight" and that the one-day window covers
it — the disagreement reaches **two days** and the window does not (#1415 has this right, the
changelog does not); and the `[verified: twoDayClockSkew_isStillRejected]` tag sits on the
*always-decrypt* claim, while that test pins only the *rejection* bound.

Also worth a clause: the `SECURITY.md` row's subject is "SWIM gossip (UDP)", but the mechanism it
names holds only for `SelfSignedCertificateProvider`. `CloudCertificateProvider` (AWS/GCP/Azure)
serves a **static** secret-store key with `previousGossipKey() → none()` and no next key — no day
derivation and no overlap at all. Pre-existing, but this PR is the edit that makes the row
specific.

Confirmed and not a concern: a suspended VM is handled — derivation is pull-based, so a clock that
jumps five days while idle uses the new day's key on the very next datagram (no timer to miss);
#683 precedence holds by mechanism (`RotatingGossipEncryptor.rotate` does `delegate.set`, replacing
the follower wholesale); #1200 is untouched. The disclosed non-atomic triple read looks
**unreachable** rather than merely rare — the rebuild's four clock reads are monotonic and the
trigger read is the first of them, so a tear needs two day boundaries inside one rebuild; I could
not construct it single-threaded.

Issue #968, opened 2026-09-08
POST /api/v1/streams returns {"status":"created"} for streams that are never registered


siy:
## What

`POST /api/v1/streams` answers **200 `{"status":"created"}`** for a stream that does not
subsequently exist in the stream registry, on any node, indefinitely. The success response is not
true.

Measured on a live 3-node rc4 cluster on 192.168.0.71 — image `aether-node:1.0.0-rc4-rc4boot` @
`sha256:6ecd091aeb606571095e8f1772efef438e116df2f6fa67fd5facf373bfda6bdf`, build SHA
**`c4d02b3e2cbcf14300ed4c164910957b06b4dda6`**, 2026-09-08:

```sh
POST /api/v1/streams {"name":"system:batch2-probe","partitions":1}
# 200 {"name":"system:batch2-probe","partitions":1,"status":"created"}

POST /api/v1/streams {"name":"batch2probe-app","partitions":1}
# 200 {"name":"batch2probe-app","partitions":1,"status":"created"}
```

Twelve minutes later, on **all three** nodes via `GET /api/v1/streams/namespaces` — a `LOCAL`-targeted
route, so these are three independent local reads and not one forwarded answer:

```json
{"entries":[{"namespace":"system","stream":"cluster-events","version":"1.0.0",
             "registeredBy":"FRAMEWORK","registeredAtEpochMs":1788879753452,"refCount":1}]}
```

Neither probe stream is present. `GET /api/v1/streams` agrees. Addressing the created name directly
returns 500:

```sh
GET /api/v1/streams/system/batch2-probe/1.0.0    # 500
```

## This is NOT a system-namespace guard bypass — my first reading, retracted

I initially read this as the `system:*` write guard failing. **It is not**, and the controls rule it
out:

- A plain application name (`batch2probe-app`) behaves **identically** — 200 `"created"`, never
  registered. Nothing here is specific to the `system:` prefix.
- The path-identity guard demonstrably **holds**, with the same ADMIN key that reads 200 on the list:
  ```
  POST   /api/v1/streams/system/cluster-events/1.0.0/publish  -> 405 "Writes to system:* streams are not permitted over HTTP"
  DELETE /api/v1/streams/system/cluster-events/1.0.0          -> 405 "Writes to system:* streams are not permitted over HTTP"
  ```

The defect is narrower and unrelated to namespaces: **a create path that reports success and leaves
nothing behind.**

## Why it matters

[...]

Issue #1282, opened 2026-09-19
Management API can create streams under reserved prefixes (entity:, topic:, system:) with arbitrary config before the real resource exists


siy:
## Summary

Stream kind is identified by **name prefix**:
- `entity:` for entity partition arcs, named by `EntityPartitionArc.arcName`;
- `topic:` for durable topics and their DLQs, named by `DurableTopicNames`;
- `system:` for system streams.

Runtime rules key off that prefix. For example, #1233 (PR #1255) makes a dropped event **fail** the publish on any `entity:` or `topic:` stream.

The management API lets an operator **create a stream under a reserved prefix, with its own config, before the real resource exists**. Found by the #1255 adversarial re-review, by reading the code; it predates that PR.
- The real entity keyspace or durable topic then finds a stream already present with an operator-chosen config: partitions, replicas, min-sync, retention.
- Whether the provisioning path adopts it, fails, or silently uses the wrong config is not established.

A blueprint stream cannot collide, because its name starts with `groupId.artifactId`, which cannot contain `:`. Only the management/API path can.

**Nit from the same review:** `system:` streams are not covered by #1233's fail-on-drop naming rule. Decide whether they should be.

## Fix direction

- The management create-stream path refuses names with a reserved prefix (`entity:`, `topic:`, `system:`) with a typed validation error. Only the internal provisioning paths may create them.
- Decide and state the drop rule for `system:` streams.

## Acceptance

- `POST` a create-stream for `topic:foo` is refused with the named error.
- Internal durable-topic provisioning still creates `topic:foo`.
- A test pins the `system:` drop-rule decision.

Related: #1233, #1255, #386.

siy:
Closed by the 2026-09-20 backlog triage (CTO, session 25): fixed on `release-1.0.0-rc4` — fixed PR #1299 7be28329e. Merges into a non-default branch never auto-close; verified by the triage against the rc4 log and merged PR bodies (`oss/internal/sweep-state/s25-triage-merged.csv`). Reopen with a reproduction at the rc4 tip if the fix does not hold.

Issue #1224, opened 2026-09-18
POST /api/v1/streams returns 200 "created" but the stream never appears in GET /api/v1/streams, and the test asserting creation cannot fail


siy:
## What

`POST /api/v1/streams` returns **HTTP 200** with `"status":"created"`, and the stream never
appears in `GET /api/v1/streams`.

Probed directly against a live 5-node cloud cluster at `v1.0.0-rc4-candidate` (`23c9b532c`),
container runtime:

```
POST /api/v1/streams  {"name":"catalog-probe-48946","partitions":1}
  -> HTTP 200
  -> {"name":"catalog-probe-48946","partitions":1,"status":"created"}

GET /api/v1/streams        (immediately after, same cluster)
  -> 0 occurrences of "catalog-probe-48946"
```

Positive control: the same `GET` returns `system/cluster-events` and
`org.pragmatica.aether.test.test-persistence/test-events`, so the endpoint works and the
search can see entries that are present. The absence is real, not a lookup failure.

## Why it matters beyond the endpoint

Every stream operation in the integration harness resolves a name to a catalog coordinate
first (`stream_coordinate` in `aether/tests/integration/lib/cluster.sh` does
`GET /api/v1/streams` and greps for `"stream":"<name>"`). When the created stream is not in
that listing the lookup returns non-zero, and `stream_publish` / `stream_info` fail before
issuing any request.

So one catalog-visibility defect presents as nine unrelated-looking assertion failures
across two suites — all of them reporting counts of zero or empty metadata rather than
anything about the catalog:

- `04-streaming`: Publish_and_verify_count (`totalEvents (0) >= published (20)`),
  Publish_10_events (`expected '10', got '0'`), Stream_metadata, Stream_visible_on_governor,
  Read_events_from_partition, Stream_in_list_after_replication, Concurrent_publish_and_query,
  Sustained_stream_publish, Stream_info_after_load
- `08-resources`: Publish_events (`expected '25', got '0'`), Stream_info_after_publish,
  Subscriber_receives_events, Competing_consumers

Reproduced identically on two consecutive cloud rounds (1 passed / 3 failed both times).

## The test cannot fail on this, which is why it survived

`04-streaming/test-stream-replication.sh` asserts creation with:

```bash
result=$(api_post "/api/v1/streams" "$payload")
assert_contains "$result" "$STREAM_NAME" "Stream created: ${STREAM_NAME}"
```

[...]

PR #1419, opened 2026-09-21
fix(streams): POST /api/v1/streams registers the catalog entry and awaits its commit (#968)

siy:
Closes #968.

## The defect at the rc4 tip (`d444d22c6`)

`POST /api/v1/streams` (`STREAM_CREATE`, body-carried `name`) answered `200 {"status":"created"}` for a stream no read route could find. Two producers, one symptom — the ticket's hypotheses 3 and 2, both true:

1. **Disjoint stores.** `StreamRoutes.mintStream` → `StreamPartitionManager.createStream(SYNC)` materialized the rings and committed the `StreamConfigKey` (awaited, 10 s). Every read (`GET /streams`, `/streams/namespaces`, `STREAM_GET`) reads the `StreamRegistryKey` catalog via `StreamNamespacesService`, which the handler never wrote. #1229 (fixing #1224 — the same defect) added the catalog-addressed `POST /streams/{ns}/{stream}/{ver}` that registers, moved the CLI to it, and left `STREAM_CREATE` wired and lying (`ManagementRoute.java:351` said so).
2. **Log-and-drop one layer down.** `KvBackedStreamRegistry.register` fired `cluster.apply(...)`, logged the failure and returned `Result.success(entry)` unconditionally — so #1229's route also answered `"created"` for a catalog put consensus refused. Its test drove `StreamNamespacesService.inMemory()`, whose register is synchronous, and could not see it. The same fire-and-forget let the `SystemStreamRegistrar` bootstrap leg latch DONE without a commit.

## The fix (no new API surface; `StreamPartitionManager` untouched — #1287's files are not in this diff)

- `STREAM_CREATE`'s handler moves to `StreamApiRoutes` (the #742 fold pattern) and shares `createAtAddress` with `STREAMS_CREATE`: the body `name` is parsed as a catalog address `namespace:stream:version`, then materialize + register. Response keeps `{name, partitions, status}`.
- `KvBackedStreamRegistry.register` awaits the apply (`REGISTER_TIMEOUT` 10 s, matching the manager's `COMMIT_TIMEOUT`) and returns the failure.
- `ReservedStreamNames.namesSystemStream` is the shared home of the body-carried system-name guard (create + group join/leave).

**Per outcome, what the client gets now** (the contract; HTTP status is the `STREAM_CREATE` handler's; the same table is in the changelog fragment):

| Outcome | Status |
|---|---|
| created — `StreamConfigKey` **and** `StreamRegistryKey` both committed, both awaited | `200 "created"` |
| already in the catalog | `200 "exists"`, `partitions` = the existing count |
| `name` missing | `400` (was `500`) |
| bare name (`batch2probe-app`) — no catalog address, nothing could ever list/read/delete the ring | `400` naming the `namespace:stream:version` form (the #1044 CLI shape) |
| reserved prefix `system:`/`topic:`/`entity:` | `400 ReservedStreamName` (unchanged; runs before the existence check, #1282) |
| enumerated system stream, bare or catalog spelling | `405` (unchanged) |
| config commit failed / timed out | `500` with the cause; nothing registered |
| catalog commit failed / timed out | `500` with the cause; ring stays, a retry registers it (`tolerateAlreadyExists`) |

**Ruling (CTO, 2026-09-22): "bare names 400, two pins."** The bare-name `400` stands (no invented default namespace, consistent with #1224's declined option); the fixture that encoded the defect is replaced, not deleted; and two further pins were required — both in `97cd127ac`.

**Judgment call, now ruled:** bare names become `400`. Before, they minted a ring reachable by no catalog-addressed route and answered `"created"`. #1224's own trace declined to invent a default namespace; #1229 chose "caller supplies the address" for the CLI; this applies the same choice to the HTTP route. `StreamRoutesCreateSystemStreamTest.createStream_ordinaryAppStreamName_stillSucceeds` **encoded the defect**: it created the bare name `orders` and asserted only that the engine map held it — the half-write, specified as intended behaviour. It is replaced (not deleted) by `createStream_bareAppStreamName_isRefusedWith400AndNothingIsMinted` and `createStream_ordinaryAppStreamAddress_stillSucceedsAndIsCatalogued`.

**System-stream registration now fails loudly at bootstrap.** Awaiting the put in `register` reaches the `SystemStreamRegistrar` bootstrap leg: before, a refused `system:cluster-events` catalog put latched the leg DONE and it was never retried; now the leg sees the failure and retries on its backoff until the put commits. Pinned through the REAL leg (`SystemStreamBootstrap` over `KvBackedStreamRegistry`) in `SystemStreamRegistrarCatalogCommitTest` — M1 reds it (`the bootstrap leg must not latch DONE on a refused catalog put`).

**The seam every new pin uses:** the registry's `ClusterNode`, a stub whose `apply` either lands the put in a real `KVStore` or fails the promise. `StreamNamespacesService.inMemory()` (used by #1229's tests) registers synchronously and is exactly why those tests could not see the drop.

## Evidence (build host, foreground, `-pl` only; every count from `<testcase>` elements)

**RED on the unmodified base `d444d22c6`** — `StreamCreateCatalogRegistrationTest` 6/6 red, `KvBackedStreamRegistryCommitOutcomeTest` 2/3 red (the third is the accepting-cluster control), with the defect's own signature:
```
[response was Success(StreamCreateResponse[name=com.example.app:orders:1.0.0, partitions=1, status=created]) but the catalog has no entry for com.example.app:orders:1.0.0]
a bare name cannot be registered in the catalog, yet the response was: StreamCreateResponse[name=batch2probe-app, partitions=1, status=created]
catalog put was refused by consensus, yet the response was: CreateResponse[address=com.example.app:orders:1.0.0, status=created]
register must not report success for a catalog put consensus refused: StreamRegistryEntry[address=com.example.app:orders:1.0.0, ... registeredBy=OPERATOR, refCount=1]
```

**GREEN with the fix**, module-full, `-pl aether/slice,aether/node` in one reactor (node alone resolves the *base* slice jar from the host m2 and the two `catalogPutRefused` tests go red again — the stale-sibling trap, observed and then avoided):
- `aether/slice`: **832** testcases, 0 failures, 0 skipped (177 XML, all newer than the run marker; re-run at `8adf21843` after the constructor seam: 832, 0/0, gate 83 files clean)
- `aether/node` at `97cd127ac`: **1625** testcases, 0 failures, **1 skipped** = `ObservabilityStep0BenchTest` (`"manual bench — run explicitly"`, class-level, not touched here). Reconciles with the `256c9c786` run: 1623 + 1 new class (`SystemStreamRegistrarCatalogCommitTest`) + 1 added method in `StreamRoutesCreateSystemStreamTest` = 1625
- `aether/aether-management-api` (comment-only change): **74**, 0/0

**jbct gate**, separate invocations: `Running JBCT check on 83 Java file(s)` (slice), `159` (node), `7` (aether-management-api) — 0 format, 0 lint errors each.

**Mutation table** (each applied and restored by content, `git diff HEAD` empty after each; `oss/internal/probes/s25-968/mutate.py`):

| # | Hunk reverted | Red tests |
|---|---|---|
| M1 | `register` back to fire-and-forget + unconditional success | `register_consensusApplyFails_returnsTheFailureNotSuccess`, `bootstrap_consensusApplyFails_isNotReportedRegistered` (slice); `legacyCreate_catalogPutRefused_isNotReportedCreated_andRetryRegisters`, `catalogCreate_catalogPutRefused_isNotReportedCreated`, `SystemStreamRegistrarCatalogCommitTest.bootstrapLeg_catalogPutRefused_doesNotLatchDone_andRetriesUntilItCommits`, `legacyCreate_catalogPutNeverCompletes_timesOutAsFailure_ringStays_andRetryRegisters` (node) |
| M2 | bare name defaults to `system:<name>:1.0.0` instead of `400` | `legacyCreate_bareName_isRefusedWith400AndNothingIsMintedOrRegistered` |
| M3 | `STREAM_CREATE` registration removed from `StreamApiRoutes.routes()` | `StreamApiRoutesEventsTest$RouteSourceWiring.routes_enumerateEverySpecRequiredVerb_asExactNameSet` |
| M4 | missing name back to a bare cause (500) | `legacyCreate_missingName_isRefusedWith400` |
| M5 | raw-name guards (405 / reserved-prefix 400) skipped | `createStream_catalogSpellingOfASystemStream_isRejectedAndNothingIsMinted` + 4 `StreamRoutesReservedPrefixTest.createStream_*` |
| M6 | response `partitions` = requested, not the ring's | `legacyCreate_repeatedAddress_reportsExistsWithExistingPartitionsAndDoesNotReRegister` |
| M8 | `register`'s bound removed (`.await()`) | `legacyCreate_catalogPutNeverCompletes_timesOutAsFailure_ringStays_andRetryRegisters` (via `@Timeout`, 5 s) |
| M7 | legacy handler mints only, never registers (the pre-fix write set) | `legacyCreate_addressName_isReadableBackFromTheCatalog`, `legacyCreate_catalogPutRefused_…`, `legacyCreate_repeatedAddress_…`, `createStream_ordinaryAppStreamAddress_stillSucceedsAndIsCatalogued` |

**Timeout branch, pinned (`8adf21843`, ruled):** `legacyCreate_catalogPutNeverCompletes_timesOutAsFailure_ringStays_andRetryRegisters` — a `ClusterNode` stub whose `apply` returns an unresolved promise while hanging; `register` returns `CoreError.Timeout` at the bound, which is not `HttpStatusAware` and so renders `500`; the ring stays materialized, the catalog stays empty; the stub is then released and a retry answers `"created"` with the entry present. The bound is injected (200 ms) through a new constructor overload `KvBackedStreamRegistry(cluster, kvStore, TimeSpan)` — the one production hunk of the ruling round, production callers unchanged, default still 10 s — so the test runs in `time="0.22"`, no sleep. Mutations: **M8** (bound removed, `.await()`) → the test reds by its `@Timeout(5 s)` instead of hanging; **M1** (fire-and-forget) → reds with `"created"`. (`e1a6119e1`'s first version of this pin waited the real 10 s; superseded.)

`[unverified: not measured on a live cluster — the ticket's 3-node reproduction was not re-run.]`

**Stale surfaces swept:** `ManagementServer`, `ManagementRoute`, `StreamApiRoutes`, `SystemStreamWriteGateTest` docstrings that named `StreamRoutes#createFreshStream`; `management-api.md` Create Stream (per-outcome table) and Create Stream Version; `management-api-versioning-spec.md`.


siy:
**Adversarial diagnosis of this PR's `forge-tests` red: it is a real defect in this change, not a flake, not ports, not a base collision.** `forge-tests` has failed **4 of 4** runs on this branch.

**The failure**, from run 35662807779 (`--log-failed`):

```
PerSliceDecisionSnapshotProbeTest.decisionsSnapshot_twoDeployedSlices_… -- 10.04 s <<< FAILURE!
[Deployment response] Expecting actual: "{"error":"request failed"}" not to contain: ""error""
  at deployTwoSlices:101
```
66 run / 1 failed / 1 skipped. That body is the test's own `ERROR_FALLBACK` — `POST /api/v1/blueprints` hit its 10 s client timeout.

**Root cause — the suite's own hang-diagnostic thread dump names it.** `pool-91-thread-1` (psa-1's Rabia executor) is parked at:

```
PromiseImpl.await  <-  KvBackedStreamRegistry.register(:126)
  <- SystemStreamBootstrap <- SystemStreamRegistrar.runPass/activate/onLeaderChange
  <- LeaderElectionState.notifyLeaderChange <- AetherNode.handleLeaderCommit
  <- KVStore.handlePut <- RabiaEngine.commitChanges
```

**Line 126 is the line this PR adds.** It awaits a `cluster.apply` that only *that same thread* can apply — the wait can never succeed, so it burns the full 10 s and freezes consensus. The POST sat queued behind it and resolved only at shutdown: *"Blueprint publish failed: Node psa-1 is inactive"*.

**Measured 1:1:** 11 × `Stream registry register(system:cluster-events) … Promise is not resolved within specified timeout`, and 11 × `SLOW-APPLY ms=2000x queueDepth=46..96`.

**Control — green rc4 forge job 106544990713** (28,235 lines): register warns **0**; SLOW-APPLY max **10,057 ms**, zero ≥20 s; this test **PASS** in 48.15 s. Positive control for the search space: `SystemStreamRegistrar` appears 169× in that base log.

**There is a latent rc4 bug underneath.** The 10 s half is `StreamPartitionManager:795`'s sibling await, present on rc4 today. This PR doubles it to 20 s, past what forge tolerates — which is why rc4 is green and this branch is not.

**Suggested fix (described, not applied):** in `SystemStreamRegistrar.activate():120`, schedule `runPass` on the existing `RetryScheduler` seam instead of calling it inline. One line, uses machinery already present, removes **both** awaits from the consensus thread, and fixes the latent rc4 bug too. Retries already run off-thread, which is why only first passes exhibit this.

The narrower alternative — change only `register()` to return a `Promise` and await it in `StreamApiRoutes` on the HTTP thread, which is all #968 actually needs — fixes leg 2 and leaves leg 1 in place.

**Also, and please do not dismiss it on a green rerun:** this branch's earlier run 35662288866 carries `[TCP 31742 must be reclaimable within 5000 ms]` (`EmberClusterSwimStartFailureTest`) alongside the same 20 s stalls. That is now a confirmed 3-occurrence pattern with its own root cause, filed as **#1456** — a `stop()`/`start()` publish race, not port contention.

`[unverified:]` the leg-1 attribution and the TCP link are inferences from the dump and the base control, not measurements. No forge run was spent: the dump plus the green base control settle it, and a rerun would only resample a deterministic condition.

Issue #1020, opened 2026-09-11
A full cluster restart loses every cluster-minted API key, locking out an operator who followed the documented rotation path


siy:
Measured on a live 3-node cluster at rc4 `e3beb0f1d`.

> **Updated 2026-09-11** with the exact measured procedure, replacing a looser first statement.

## What was measured

1. Created `probe-cluster-key` via `POST /api/v1/cluster/keys` — returned `source=cluster`, `ACTIVE`, and authenticated successfully (200).
2. Restarted **all three** nodes.
3. The key was **absent** from `GET /api/v1/cluster/keys` and authentication returned **403**.

**No volume is mounted for node data — neither shipped compose file mounts one.**

The bootstrap credential is not regenerated to compensate, because `BootstrapAdminKeyRegistrar` runs only at **first leadership**.

## Operator consequence

Mint a key, drop the config entry (which is the documented hardening step), cold restart — and **no credential remains**. The operator is locked out of their own cluster.

The asymmetry is the dangerous part:

- a **cluster-minted** key vanishes on cold restart — fails **closed**;
- a **config-declared** key is re-seeded from its file — fails **open**.

So a revocation recorded only in the KV store does not survive a cold restart, while the credential it revoked does.

## Related mechanism, verified while investigating

There is no readiness signal indicating a node's KV state has been populated from peers. Searched `lastApplied`, `appliedIndex`, `snapshotRestored`, `stateLoaded`, `isSynced` — only `restoreSnapshot` exists, with no flag. So a node reaching `ACTIVE` does **not** imply its key registry is populated, and a restarted node can answer differently from its peers for the same credential: a **rolling** restart (quorum never lost) produced **HTTP 200 from the restarted node and 403 from both peers**, for the same key.

## Explicitly NOT verified

Whether mounting a persistent volume changes this. That is the obvious hypothesis and it has not been tested.

## Scope

This is not about the published image's built-in credential, which is handled separately. It applies to any deployment relying on cluster-minted keys as the durable credential store.

siy:
Fixed by #1432, squash-merged into `release-1.0.0-rc4` as `7ed79e73f` (CTO session 26, 2026-09-22), after adversarial review plus a fix round closing three MEDIUM findings — each pinned by a mutation that reddens a **named** test:

[...]

Issue #1013, opened 2026-09-11
A failed or absent snapshot restore signals readiness unconditionally, so metadata loss is silent


siy:
Measured at rc4 `7fa26e2ca`.

`StorageFactory.restoreAndSignalReady` is:

```java
snapshotManager.restoreFromLatest().onPresent(snapshot -> applySnapshot(name, snapshot, metadataStore));
readinessGate.snapshotLoaded();
```

There is no else branch, and `readinessGate.snapshotLoaded()` runs **unconditionally**. A snapshot that is absent, unreadable, or whose `LATEST` pointer dangles signals readiness identically to one that restored successfully.

The only difference an operator can observe is the *absence* of the `Restored snapshot for '{}'` info line — an absence, in a log where absence is not alerted on and is indistinguishable from a node that legitimately had no snapshot to restore.

This is what makes the corruption in the related snapshot-prune issue silent: a node comes up having lost its metadata, reports ready, and replays its entire WAL, with nothing in the readiness signal reflecting that anything went wrong.

**Suggested shape:** distinguish "restored", "nothing to restore", and "restore failed" as separate outcomes, and make the third loud. Readiness should state which of the three it is signalling.

siy:
**Handoff from PR #1432 (#1020), which merged as `7ed79e73f` — one behaviour it introduced is this ticket's territory and shipped UNPINNED.**

While fixing #1020, the reviewer measured both arms with a positive control on each: **a failed own-snapshot restore now wedges the node.** `restoreState`'s `.onFailure` log-and-drop skips `activate()`. At that head the node never activates; with the hunk reverted it does; the log shows the retry loop, two attempts 5 s apart.

I ruled that fix **out of scope for #1432** because whether a failed restore should still activate — and whether wedging is correct but must be explicit and bounded — is exactly this ticket's subject, and two branches authoring the same behaviour is the collision we are trying to stop having.

**So it merged without a pin.** The minimum honest fix that stays off this ticket's territory was scoped but not executed:
1. make the surviving signal honest — the per-attempt ERROR should name the consequence (does NOT activate, stays Syncing, retries every `syncRetryInterval`) and render the cause's message, which was measured rendering **empty**: `failed to restore state: Cause: `;
2. state in `activateWithoutAdoption`'s docstring that a failed `restoreSnapshot` does not activate — it currently says it "activates on this node's own state";
[...]

PR #1432, opened 2026-09-21
fix: #1020 install a node's own persisted snapshot when it activates on its own state

siy:
Closes #1020.

## What was wrong

A minted API key is a `KVCommand.Put` committed through consensus into the replicated KV store; that store is durable only through `[backup]` (`GitBackedPersistence`), in-memory otherwise. Two defects sat on the restore path even with `[backup]` enabled:

1. **#1341 (already merged):** `AetherNode::base64ToSnapshot` was handed the `# Phase: N` header `GitBackedPersistence` writes, so every production `load()` failed to `none()` and a restarted node was amnesiac whatever was on its disk. The first run of the forge pin below, against `d444d22c6`, wedged the restart: one node "restored state … phase 0", the other Syncing forever.
2. **This PR:** `RabiaEngine` read `persistence.load()` for the sync-response payload, the adoption floor and the future-history detector — never to populate its own state machine. `activateWithoutAdoption` (every response behind self's persisted phase, or n=1) activated an EMPTY store at phase 0. After a full-cluster stop, a key came back only if the first responder's persisted phase happened to EQUAL self's; a staggered graceful stop breaks the tie, and the node holding the MOST advanced snapshot came up empty and answered 403 for a key it had acknowledged.

`activateWithoutAdoption` now installs the node's own persisted snapshot through `restoreState` when its phase is ahead of the live one; a live phase at or past the persisted one (a resync from ACTIVE) still installs nothing.

## Guarantee earned, and its bounds

A key minted and acknowledged before a **graceful** all-nodes stop is accepted after restart, because it is in the consensus KV snapshot, saved by `RabiaEngine.shutdownAndReset` to `[backup] path` at stop, and restored at sync adoption — from a peer's persisted snapshot when a peer is ahead, from the node's own when every responder is behind it.

NOT earned: (a) any deployment without `[backup]` — the shipped `aether/docker/docker-compose.yml` and container `aether.toml` carry none and mount no volume, so they still lose every KV record on a full restart; the derived `bootstrap-admin` is re-registered by the fresh leader, so a REVOKED bootstrap key comes back ACTIVE `[design intent — unverified: read at BootstrapAdminKeyRegistrar/BootstrapAdminKeyLeg, not measured live]` — the ticket's "not regenerated" claim is refuted by the producer; (b) a key minted after the last lifecycle save when a node is SIGKILLed — `persistence.save` never runs on commit; (c) ~~`GitBackedPersistence` does not create `[backup] path`~~ **RETIRED at `8e5be7270`** — `writeTomlFile` now creates it, pinned by `GitBackedPersistenceTest#save_backupDirectoryDoesNotExist_createsItAndWritesTheState` (1 red of 810 when the creation is removed). The claim as originally written was also **false of one call site** even before this: `applyRestoredState` discarded its save `Result` entirely, so that save failed SILENTLY rather than at ERROR — now fixed and pinned. Deployment changes for (a) are deliberately not in this PR — awaiting a CTO ruling.

## Evidence

Base: `a5f649483` (rc4 tip with #1341). All runs on the build host, foreground, `-pl`.

**RED (unit, on `d444d22c6`):** `RabiaSyncAdoptionOwnSnapshotTest` 3/5 red — `expected: [111, 119, 110] but was: null`, `expected: Phase[value=5] but was: Phase[value=0]`; 2 controls green.

**RED (forge, on `a5f649483` with only the `RabiaEngine` hunk reverted):** `ApiKeyFullRestartForgeTest` — `forge-1020-second-minted-key presented on akr-2: expected: 200 but was: 403`, held for the full 30 s poll.

**GREEN:** `integrations/consensus` **810** `<testcase` elements at `8e5be7270` (806 at `e511aab70` + 4 added by the review follow-ups), 0 failures, 0 skipped. The `<testsuite tests="N">` attribute total for the same run reads **228** — the known `@Nested` undercount, computed only as a tripwire and never used as a total. `aether/node` 1619 `<testcase`, 0 failures, 1 skip = pre-existing `ObservabilityStep0BenchTest#step0_overhead_bench`. `ApiKeyFullRestartForgeTest` 1/1 in 70 s; node 2 logs `activating on its own persisted state (every response is behind this node's own state); persisted phase 14`, D9 future-history WARN fires detect-only, node 1 adopts node 2's state, node 3 joins live at phase 20, `BootstrapAdminKeyRegistrar` finds the restored admin key.

**Mutation table:**
| Hunk reverted | Red |
|---|---|
| `activateWithoutAdoption` own-restore (whole hunk) | exactly 5: `RabiaSyncAdoptionOwnSnapshotTest#everyResponderBehind_…`, `#singleNode_…`, `#ownSnapshotEmpty_…`, `RabiaSyncAdoptionQuorumTest#ownMoreAdvancedState_isNotRegressed_byStalerResponses`, `#futureHistory_stillDetected_…` (29 others green); forge pin red as above |
| ahead-of-live guard dropped (`filter(_ -> true)`) | exactly 1: `#resyncFromActive_ownStaleDiskSnapshotIsNotInstalledOverTheLivePhase` (41 others green) |

The two existing `RabiaSyncAdoptionQuorumTest` pins asserted "nothing installed" — they specified the defect; they now assert self's snapshot is installed and the responders' is not, which still reddens if the adoption floor is deleted (the mutation they were written against).

**Gates:** `jbct:check -Djbct.skip=false -pl integrations/consensus`: `Running JBCT check on 69 Java file(s)`, 4 format / 2 lint — IDENTICAL result with the base `RabiaEngine.java` swapped in (control), i.e. pre-existing under `integrations/` where CI never runs it; my hunk adds no finding. `-Djbct.includeTests=true` on consensus: 138 files, my new test file clear of format issues (72 pre-existing elsewhere). `forge-tests` (`-Pwith-e2e -Djbct.includeTests=true`): 53 files, 48 format / 172 lint pre-existing; my file formatted, 5 lint findings of the module's `throw new AssertionError` test idiom (neighbours: 37 and 32).

Changelog: `changelog.d/1020-own-snapshot-on-restart.md`.

## Review follow-ups (`8e5be7270`, on top of the reviewed `e511aab70`)

Three MEDIUMs from the adversarial review, each pinned by a mutation that reddens a **named** test. Space for every red set: all **810** `<testcase>` elements of `integrations/consensus`, counted as elements.

| Mutation | Red set (of 810) |
|---|---|
| Drop the `.onFailure` on the restore-path save again (swallow it) | **1** — `RabiaRestoredStateSaveFailureLogTest#restoredStateSaveFails_logsAtErrorNamingTheConsequence` |
| Remove `createDirectories(backupDir)` from `writeTomlFile` | **1** — `GitBackedPersistenceTest#save_backupDirectoryDoesNotExist_createsItAndWritesTheState` |
| Flip the own-snapshot guard `> 0` → `>= 0` | **1** — `RabiaSyncAdoptionOwnSnapshotTest#resyncFromActive_ownDiskSnapshotAtTheLivePhase_isNotInstalled` (`expected: null but was: [111, 119, 110]` — the disk snapshot resurrected) |

The third was the review's MX2, which left all 806 green; the boundary now has a test.

**Gate, run by hand because CI cannot run it here.** `integrations/` inherits `jbct.skip=true` (#1442), so CI's format/lint gate never examines this file. `mvn jbct:check -Djbct.skip=false -pl integrations/consensus` at `8e5be7270`: `Running JBCT check on 69 Java file(s)` — **4 files need formatting, 3 lint errors**, byte-identical to the same command at `e511aab70`. `[unverified: this module's gate is RED at the PR head and remains so — QuicClusterClient, RabiaEngine, TopologyObserver, BackoffConfig, plus JBCT-RET-06 ×2 and JBCT-RET-07 in RabiaEngine#stop. The three lint errors are JBCT-RET-06 in `QuicClusterClient#initiateConnection` and `QuicClusterNetwork#evictStaleConnection`, and JBCT-RET-07 in `RabiaEngine#stop` — all four files and all three lint errors are PRE-EXISTING and none is introduced here; parity with the PR head is what is claimed, not green.]`

**MEDIUM 2 — NOT IN THIS PR. Corrected after the fact.** This PR was squash-merged at head `8e5be7270`; the MEDIUM 2 commits were pushed to the branch AFTER that merge, so they are **not** in `7ed79e73f` and not on rc4 — verified by content against `release-1.0.0-rc4` (`FAILED to restore state and is NOT active` → 0 occurrences; `RabiaOwnRestoreFailureTest.java` → HTTP 404; positive control: the fetched `RabiaEngine.java` is 113,604 bytes, negative control 0). An earlier revision of this section described MEDIUM 2 as shipped here, which was wrong. The work is rebased onto rc4 tip `24a0e0130` and now ships as **PR #1464** (branch `fix/1020-own-restore-wedge-1013`, `6aa092597`) — 814 `<testcase>` elements, 0 failures, 0 skipped on that base — opened `Part of #1013`, because it pins the current behaviour without deciding it. What follows describes that branch, not this merge. A failed `restoreSnapshot` still skips `activate()`; what changed is that it is now SAID and PINNED. The docstring on `activateWithoutAdoption` states that it does not always activate despite its name; the ERROR names the CONSEQUENCE (node NOT active, serves no requests, every retry re-enters) and renders the cause's message rather than a bare object, because that line is the entire operator surface for the state — the periodic stuck-in-`Syncing` WARN is structurally suppressed here (**#1447**).

| Mutation | Red set (of 813) |
|---|---|
| Restore the pre-#1020 bare-cause log | **1** — `RabiaOwnRestoreFailureTest#ownRestoreFails_logsAtErrorNamingTheConsequence` |
| Make a failed own-restore activate anyway — **the change #1013 might make** | **1** — `RabiaOwnRestoreFailureTest#ownRestoreFails_staysInactive_untilTicket1013Decides` |

**Whether the wedge is correct, whether it should be bounded or terminal, and what readiness reports meanwhile are #1013's decisions and are deliberately not taken here.** The pin is an ENABLED tripwire rather than a `@Disabled` placeholder: it asserts current behaviour and its failure message names #1013 as the ticket that may legitimately flip it, so the change cannot be made silently. `#ownRestoreSucceeds_activates` is the control that makes its negative assertion mean "genuinely did not activate" rather than "the fixture never got there".

Gate re-checked at `7287f8b72`: `Running JBCT check on 69 Java file(s)`, **4 files / 3 lint errors — parity with `e511aab70`**. Worth recording that an intermediate version of this hunk ended the chain in a method reference, which the rc4 linter reads as `JBCT-RET-07 Discarded chain result` (3 → 4, my own regression); the lambda form restores parity, and both mutations above were **re-measured against the corrected code** rather than carried over.


siy:
Reviewed at `e511aab70`, in an independent clone, against the four real checks plus my own runs.
The fix is correct and I could not break it. Approving with two notes, both pre-existing silences
that this change newly routes traffic through.

**What I confirmed.** 806 `<testcase>` elements green, 0 skipped, matching the PR body. All ten
cited test methods exist (negative control clean). The load-bearing correctness argument — that
`persisted.phase > currentPhase` cannot be defeated by live traffic arriving mid-sync — holds, and
I checked it at the producer rather than the docstring: `handleDecision` (RabiaEngine.java:2125)
buffers unconditionally while `Syncing`, so `commitDecision → advancePhase` is unreachable and
`currentPhase` is genuinely 0 on a cold boot. `commitChanges` also runs before `advancePhase`, so
the guard errs in the safe direction.

**#1409 collision — resolved, and merge order does not matter.** Both PRs modify `RabiaEngine.java`
from the same pre-image blob but ~300 lines apart (#1409: `registerBatch`/`doHandleNewBatch`;
#1432: sync adoption). `ort` merges them with no conflict, and I built and tested the **merged**
tree: 807 testcases, 0 failures. The +1 is derived, not assumed — set-diffing the two runs gives
exactly #1409's new test added and nothing removed.

**MEDIUM 1 — the re-persist after restore is the one save nobody checks.** `applyRestoredState`
(RabiaEngine.java:1361) calls `persistence.save(...)` and **discards** the `Result<Unit>`. The other
three save sites (543, 624, 654) all attach `.onFailure(log::error)`, and `GitBackedPersistence`
contains **zero** log statements — so a failed save here is completely silent, and the next line
logs `"restored state from persistence"` at INFO regardless. This matters for #1020 specifically:
that save is what makes the restored state durable for the *next* restart. With an unwritable or
absent `[backup] path` — which this PR correctly notes `GitBackedPersistence` never creates — the
node comes up ACTIVE and correct in memory, its disk stays stale, and #1020's symptom returns one
restart later with no diagnostic. It also means limitations item (c) as written ("a missing dir
fails **every save at ERROR**") is not true of this call site. One `.onFailure(log::error)` fixes
the silence.

**MEDIUM 2 — a failed own-snapshot restore now wedges the node instead of activating it.**
`activateWithoutAdoption` previously always activated. It now routes through `restoreState`, whose
non-empty arm is `restoreSnapshot(...).onSuccessRun(this::activate).onFailure(log::error)` — so a
`restoreSnapshot` that fails skips `activate()` and the node stays `Syncing`, re-entering the same
branch on every retry tick. I measured it on both arms with a positive control on each: on this head
the node never activates; with the hunk reverted it does. The run log shows the loop directly — two
`activating on its own persisted state … persisted phase 5, live phase 0` lines five seconds apart
with the ERROR between them. This is fail-closed and arguably the right direction (it never serves
the empty store #1020 is about), and the same shape already existed on the peer-adoption path — but
the branch whose docstring says it "activates on this node's own state" may now not activate at all,
and no test covers it. Worth one docstring sentence at minimum, and preferably an explicit choice
between a fallback activation and a terminal diagnostic rather than an unbounded retry.

**MEDIUM 3 — the guard's equality boundary is unpinned, and it is what protects a reconfigure.**
Changing `> 0` to `>= 0` in the new filter leaves **all 806 tests green**. The boundary is not
arbitrary: `reconfigure` (RabiaEngine.java:624) persists the live, **non-empty** state-machine
snapshot under `Phase.ZERO` before calling `stateMachine.reset()` — the comment there calls it
"empty state", which is not what it saves; only the phase is zeroed. With `> 0` that snapshot is
correctly not reinstalled; with `>=` the previous membership's state comes back. Exposure today is
low (after a reconfigure the floor is 0, so the only reachable entry is `responses.isEmpty()`, i.e.
clusterSize 1, which is not a shipped topology) — but a load-bearing boundary with no test is one
edit away from mattering. One test on the equality case would close it.

**NIT.** "a key came back only if the first responder's persisted phase happened to EQUAL self's" —
the floor test is `candidate < floor`, so pre-fix the node recovered whenever a responder was **at
or ahead of** self, not only when equal. Same sentence in the changelog fragment.

Non-blocking context for whoever merges: `integrations/` inherits `jbct.skip=true`, so CI's
format/lint gate never examined this file — pre-existing and repo-wide, and the author already ran
the gate by hand and reported it. And `CodeRabbit` reports `success` on this PR with the description
"Review skipped: reviews are disabled for this base branch", so this is the only review it has had.

Issue #1190, opened 2026-09-14
FileOps.writeBytes does not fsync: a crash can leave an encryption marker absent while ciphertext exists


siy:
`FileOps.writeBytes` is `Files.write(CREATE, TRUNCATE_EXISTING)` — **no `fsync` on the file, and no parent-directory `fsync`.** Storage encryption markers are written through it.

**Consequence.** A crash after `commitMarker` has returned, but before the bytes reach stable storage, leaves the marker **absent while ciphertext blocks exist**. On the next boot:

- with `encrypted = true` → fail-closed as `EnablingOverExistingPlaintext`, i.e. the node refuses to start for a reason that is not the real one;
- with `encrypted = false` → **a plain tier is mounted over ciphertext.**

The second arm is the dangerous one: the guard that exists to prevent exactly this cannot see a marker that was never durable.

**Distinct from #1169 and #1118.** Those concern `rename(2)`/`ATOMIC_MOVE` durability where something is *replaced*. Here **nothing is replaced** — the marker is created, and the gap is the absence of an `fsync` on the create path plus the parent directory entry. `LocalDiskTier.writeThenRename` is untouched by this report. Filed separately so #1169 is not widened into a different mechanism.

**Also distinct from a torn write**, which was checked and is safe: presence is the signal and the content is only a diagnostic key id, and `refuseIfEncryptedWithoutKeyring` already handles `bytes.length == 0`.

Found while enumerating the failure points of the storage boot decision for #852 / PR #1145 (failure point 4 of 6 in that analysis). Pre-existing; deliberately not folded into that PR.

siy:
**Sharper distinction from #1169**, from the agent that enumerated the storage boot decision's failure points:

> #1169 is about **REPLACING** an existing file — `writeThenRename`'s `rename(2)` is not atomic in the assumed sense, so a replace can be observed torn. **This is not a replace:** the marker is only ever **CREATED**, because both guards short-circuit the moment a marker exists, so there is never an old version to tear. The defect here is **DURABILITY of a create** (no fsync), not **atomicity of a replace**, and it sits in `FileOps.writeBytes` — shared by every caller — not in `LocalDiskTier`. **Fixing #1169 does not touch it.**

Precise call path: `EncryptingStorageTier.writeMarker` → `FileOps.writeBytes` (`core/src/main/java/org/pragmatica/lang/io/FileOps.java:97`), at `b64eacc61`.

[...]

Issue #1169, opened 2026-09-14
LocalDiskTier.writeThenRename replaces the block file with a non-atomic Files.move: the previous copy is unlinked before the rename, contradicting its javadoc and the #910 ruling's 'rename(2) atomic' premise


siy:
**What.** `LocalDiskTier.writeThenRename` (integrations/storage) replaces a block file through `FileOps.moveReplace` = `Files.move(REPLACE_EXISTING)` **without** `ATOMIC_MOVE`. The JDK's `UnixFileSystem.move` unlinks the target before the rename, so there is a window where the previous copy is gone and the new one is not yet in place; a crash there loses the block, and a failed rename leaves neither. The javadoc says "previous copy never touched", and the #910 ruling recorded as `know: 2f68fded1` ("`.partial` then rename — `rename(2)` atomic, EXDEV unreachable by construction") rests on the same false premise — the partial IS a same-directory sibling, but the move that replaces it is not the atomic one.

**Found by** verify-1118 (`oss/internal/verify-1118-2026-09-14.md`, delta addendum) while reviewing #1118's identical defect in `GitBackedPersistence` (B1 there, fixed by the new `FileOps.moveAtomic` = `REPLACE_EXISTING + ATOMIC_MOVE`). `RemoteRepository.cacheArtifact` correctly keeps `moveReplace` (system-tmp source, cross-device, a cache).

**Fix.** Use `FileOps.moveAtomic` (lands with #1118); pin with the reviewer's directory-partial shape — make the partial a directory: `rename(dir, existing file)` is ENOTDIR under `ATOMIC_MOVE` and the target survives, whereas the non-atomic path unlinks first and then succeeds — run at the real call site (7/7 green under atomic, red "isDirectory=true / read empty" under `moveReplace`). Correct the javadoc; the `know:` correction is recorded separately.

siy:
Fixed by #1439, squash-merged into `release-1.0.0-rc4` as `c9b8b8406` (CTO session 26, 2026-09-22).

[...]

Issue #910, opened 2026-09-06
Write-through: cache-tier put failure fails the whole put after the durable write already succeeded


siy:
## What
In the write-through cache-promotion path, a cache-tier put failure fails the whole write after the durable write has already succeeded, and the log line describing the failure claims behavior ("skipped") that does not actually happen.

## Where
`integrations/storage/src/main/java/org/pragmatica/storage/StorageInstance.java:318-332` (`promoteToNextCacheTier`). The chain is `tier.put(id, content).onSuccess(...).onFailure(cause -> log.debug("Cache promotion to {} skipped for {}: {}", ...)).flatMap(_ -> promoteToNextCacheTier(...))` at line 328 (log) through 332 (`.flatMap`). Because `Promise.flatMap` (`core/.../Promise.java:156`) propagates failure rather than swallowing it, the failed cache-tier put fails the entire outer `put` call — the promotion was not "skipped", it took down the caller's write.

## Why it matters
The durable write already succeeded before the cache-tier put runs, so the caller receives a failure for data that is, in fact, durably stored. In the two-tier case the block's claim gets released and the bytes are orphaned on disk with no metadata record (GC there is metadata-driven, so the orphan is never collected). In the three-tier case (with a DHT `content` tier) a stale record survives `releaseClaim` and a caller retry dedups to a refcount of 2. The existing regression test for this area, `StorageIntegrationTest.java:192` (`memoryTierFull_writesToDisk_readsFromDisk`), writes 25 bytes into a 32-byte tier and never actually fills it, so it cannot observe this failure mode despite its name suggesting it covers cache-tier-full behavior.

## Evidence
Found by review-901 (NOTE-level finding, read-verified against source, not executed): `StorageInstance.java:318-332` and `StorageIntegrationTest.java:192`, both confirmed on `pragmatica` HEAD. Full writeup in `oss/internal/review-901-886-2026-09-06.md` (finding N-1). Pre-existing; not introduced by #886/#901.

## Acceptance
- A cache-tier put failure during write-through promotion, occurring after the durable write already succeeded, does not fail the caller's `put` — the caller sees success, and the cache-tier failure is logged accurately (not as "skipped" when it in fact aborted the promotion chain).
- A test that genuinely fills a cache tier (not `memoryTierFull_writesToDisk_readsFromDisk`'s current 25-into-32-byte case) demonstrates the caller receiving success despite the cache-tier put failure.

siy:
[...]

PR #1439, opened 2026-09-21
fix: atomic block replace in LocalDiskTier and durable encryption marker write (#1169, #1190)

siy:
Closes #1169
Closes #1190

Base `release-1.0.0-rc4` at `08edcacb7`. Two rc4 blocking defects on the same primitive layer, fixed with minimum code; guarantees stated per operation with the mechanism that earns them.

## What changed

- **#1169** `LocalDiskTier.writeThenRename`: `FileOps.moveReplace` → `FileOps.moveAtomic` (`REPLACE_EXISTING + ATOMIC_MOVE`, one `rename(2)`; the partial is `resolveSibling`, so EXDEV is unreachable). Guarantee: a reader concurrent with a replace sees the previous complete copy or the new one, never absence; a failed rename or a crash during it leaves the previous copy in place. Javadoc corrected; `know: 812d4de2c` already records the #910 correction.
- **#1190** new `FileOps.writeBytesDurable(path, bytes)` (`core`): one channel writes and `force(true)`s the file (data + inode), then the parent directory is opened and `force(true)`d so the directory entry is durable. `EncryptingStorageTier.writeMarker` uses it. `writeBytes` is unchanged; its javadoc now says it does not fsync (no other production caller — cache-tier partial, jar caches — needed the cost). No temp+rename for the marker: it is create-only and torn-safe, and an orphaned temp at the directory root would be read by `isBlockFile` as a plaintext block and refuse the next enable.
- **Fixture corrected, stated as such:** `CacheTierPartialWriteTest.localDiskTier_emptyDirectorySquattingTheBlockPath_isReplacedByTheBlock` (#1095) reddened under the fix. Its own docstring names what it pinned — "the JDK removes it before the rename" — i.e. the unlink step #1169 removes. The fixture was wrong, not the fix. Rewritten as `…_failsThePut_andIsLeftInPlace`: the same outcome the non-empty squatter test `CacheTierPartialWriteTest.localDiskTier_failedWrite_releasesTheReservation` pins — named, because an unnamed "the sibling test" is a citation no reader can check. That test asserted only the failure half, so it now also asserts that the squatter survives untouched, making the parity real in both halves. Reservation released, no partial left. Note only the EMPTY case discriminates the two move implementations (`rename(file, non-empty-dir)` is refused under both), which is why H1 reddens the empty one.

## Crash-window truth (#1190)

Before `commitMarker` returns the marker may be absent or torn. Torn is safe (presence is the signal; zero bytes handled). Absent is fail-safe: the guard arms a marker write only over an EMPTY directory, and `StorageFactory.createAll` hands tiers out only after every marker has committed — no ciphertext block can exist beside an absent marker. After the return, marker and directory entry are fsynced. `[unverified: power loss — what the device does with a completed fsync]` `[unverified: Windows — directory channels cannot be opened; writeBytesDurable FAILS there rather than weakening silently]`.

## Caller table (ruling 2026-09-22)

Grep space: `FileOps.writeBytes(` / `writeBytes(` / `FileOps::writeBytes` over the tree at `41c9bdaa7`, excluding `target/`, `src/test/` and the unrelated `ByteBuf.writeBytes` hits; `writeString` callers included because a `LATEST` pointer or manifest goes through it.

| Caller | Op | Durable? | Disposition |
|---|---|---|---|
| `EncryptingStorageTier.writeMarker` | `writeBytes` → `writeBytesDurable` | **y** | fixed here |
| `LocalDiskTier` default writer (`.partial`) | `writeBytes` | n | cache tier; leftover partial swept at startup, lost block = cache miss (#910) |
| `RemoteRepository.cacheArtifact` | `writeBytes` to system tmp + `moveReplace` | n | jar cache, cross-device |
| `BuiltinRepository` | `writeBytes` to tmp jar | n | transient extraction |
| `DefaultSnapshotManager.writeDurably` (snapshot, `LATEST`) | `writeString` + own `force(true)` + `moveAtomic` | already forced; rename's directory entry not fsynced | #1013/#1418 neighbourhood — named, not touched |
| `GitBackedPersistence.writeDurably` (`STATE_FILE`) | same shape | already forced; same bound | other layer — named, not touched |
| `ForgeDataDir.writeMarker` (`.forge-owner`) | `writeString` | n | absent marker REFUSES the run (`UnownedState`), fail-safe by its own design |
| jbct / aether-setup generators | `writeString` | n | build-time outputs |

Plain-JUnit control beside the JFR pin: `commitMarker_createsTheMarkerHoldingTheActiveKeyId` (exists, key id bytes); JFR threshold 0 (default 20 ms would drop every warm-disk fsync); RED on base with `[]`.

## Evidence (all on the build host, Temurin 25, ext4)

**RED on the unmodified base** (`remote-mvn … -pl integrations/storage test -Dtest=LocalDiskTierAtomicReplaceTest,EncryptionMarkerDurabilityTest`, log `20260921T224913Z`):
- `directoryPartial_renameRefused_previousCopySurvivesUntouched`: "a partial that cannot be renamed over the block must fail the put" — the non-atomic move unlinked the previous copy and renamed a DIRECTORY over the block path.
- `readerRacingReplace_seesOldOrNew_neverAbsence`: **3,526 absent reads of 49,436** across 2,000 replaces.
- `commitMarker_forcesTheMarkerFileAndItsDirectory_beforeReturning`: `jdk.FileForce` events `[]`.

**GREEN with the fix**, module-full at `41c9bdaa7` (`-pl core,integrations/storage test`, `<testcase>` elements from marker-fresh XML, 0 stale excluded): **core 934, storage 301**, 0 failures, 0 skipped. Storage base 297 + 4 new.

**Mutation table** (each hunk reverted by `sed`, landing verified by content count 1→0, restored 0→1, `git diff HEAD` empty after each; run over `FileOpsTest,LocalDiskTierAtomicReplaceTest,EncryptionMarkerDurabilityTest,CacheTierPartialWriteTest`):

| Hunk | Red |
|---|---|
| H1 `moveAtomic`→`moveReplace` (LocalDiskTier) | `directoryPartial_…`, `readerRacingReplace_…` (**2,911 absent of 42,028**), `…emptyDirectorySquatting…_failsThePut_andIsLeftInPlace` — 3 red, 39/39 core green |
| H2 marker via `writeBytes` (EncryptingStorageTier) | `commitMarker_forces…` — 1 red |
| H3 drop `file.force(true)` (FileOps) | `FileOpsTest.writeBytesDurable_forcesFileAndParentDirectory_beforeReturning`, `commitMarker_forces…` — 2 red |
| H4 drop `directory.force(true)` (FileOps) | same 2 red |
| H5 `writeMarker` returns success without writing (validates the plain control) | `commitMarker_createsTheMarkerHoldingTheActiveKeyId`, `commitMarker_forces…` — 2 red |

**jbct gate** (`jbct:check -Djbct.skip=false -pl <m>`): core `Running JBCT check on 62 Java file(s)` — 0 format, 0 lint. storage `57 Java file(s)` — 0 format, **1 pre-existing lint error** `[JBCT-RET-06] RemoteTierConfig.java:37` (untouched by this PR, present at the base; not widened into scope).

**Standalone probe** (`oss/internal/probes/s25-1169/Probe1169.java`, macOS APFS): non-atomic 203,018 absent of 343,385 reads over 20,000 replaces; atomic 0 of 183,040; JFR observes both the file and the directory force, and a plain `Files.write` control emits none.

Not changed: `RemoteRepository.cacheArtifact` keeps `moveReplace` (system-tmp source, cross-device, a cache — per #1169).

---

## Review round — s26 adversarial review (head was `41c9bdaa7`)

Two MEDIUMs were raised. Both are addressed; neither was a defect in shipped behaviour, which is unchanged.

**MEDIUM-1 — the two properties #1190 is about were UNPINNED.** Both JFR observers read only
`event.getString("path")` and asserted with an unordered `contains(file, dir)`. That pins *that* a force
was issued against two paths — never *when* relative to the write, nor *whether* metadata was included.
The reviewer ran two mutations that restore the defect and left **everything green**. Both now go RED:

| Mutation | Property | Before (at `41c9bdaa7`) | Now |
|---|---|---|---|
| force the parent directory **before** the write | fsync ORDER — the directory entry is otherwise never synced after the file is created, i.e. #1190's defect restored | **GREEN** 39/39 core + 2/2 storage | **RED in both observers** |
| `file.force(true)` → `file.force(false)` | fsync METADATA — `fdatasync` on a freshly created file need not make its inode (size, link count) durable | **GREEN** 39/39 core + 2/2 storage | **RED in both observers** |

The fix, in both observers (they are deliberate twins in different modules): the helper now returns a
`ForcedFile(path, metaData)` record instead of a bare `Path`, sorted by event start time, and the
assertion is `containsSubsequence(new ForcedFile(file, true), new ForcedFile(dir, true))`. Sorting makes
the chronology a property of the helper rather than an assumption about `RecordingFile` replay order.
Each failure names the mutated property directly — `metaData=false` in actual vs `metaData=true` expected,
and the inverted `[dir, file]` actual vs `[file, dir]` subsequence.

**MEDIUM-2 — the "sibling non-empty-squatter test" citation.** The review reported this test **does not
exist**. It does: `CacheTierPartialWriteTest.localDiskTier_failedWrite_releasesTheReservation`, enabled,
whose docstring reads *"a non-empty directory squatting on the block path makes the rename fail after the
reservation"* and whose body writes an `occupant` file into the squatting directory. The review's two
positive controls both passed while neither pattern could reach it — the test is named by OUTCOME (no
`Squatting` token) and the prose spells it hyphenated, `non-empty`, which `[Nn]onEmpty` cannot match. A
control proves the pattern, never the location. Search space for this correction: all 3,921 `*.java`
excluding `/target/`, on the token `squat`, case-insensitive, 14 hits enumerated; positive control for the
pattern used, `non-empty` → 1.

So the claim was true and the citation was merely unfalsifiable. Both are fixed above rather than deleted,
and the real gap — the sibling pinned the failure but not the "left in place" half it was cited for — is
closed.

**Also in this round:** `readerRacingReplace`'s `torn` counter was structurally unreachable in both arms
(content is written to a sibling and published by rename, so no reader can observe a partial) — removed,
with the reason recorded at the test so it is not re-added; the `reads > 0` aliveness guard, which does
real work, stays.

**Not done, and why:** the review's N-4 "mangled indentation in the new `FileOps` try-blocks" is not the
author's. `jbct format` **reproduces** that exact indentation — it is the formatter's canonical output for
a try-block inside a lambda body, and the same shape pre-exists in `FileOps.walk` and `FileOps.list`, which
this PR never touches. Tidying it by hand makes `jbct:check` FAIL. It needs a `jbct-format` ticket, not a
change here.

**Re-verified at this head:** `-pl core,integrations/storage test` → **core 934, storage 301, 0 failures,
0 skipped**, `<testcase>` elements from both report dirs, marker-fresh, 0 stale; identical to the
pre-review baseline, so nothing was lost. Gates per module: core `Running JBCT check on 62 Java file(s)` →
`JBCT check passed.`; storage `Running JBCT check on 57 Java file(s)` → the same single pre-existing lint
error `[JBCT-RET-06] RemoteTierConfig.java:37`, a file absent from this branch's 8-file diff, left alone.

### MEDIUM-2 follow-up: the assertions added above are themselves pinned

An assertion added to close a citation gap is still an instrument, and an unpinned instrument is what
produced the gap. Predictions pre-registered before either mutation ran; **6/6 correct**. Two
mutations were needed, and that is itself the finding — one would have hidden the second assertion.

| Mutation (both scoped to the failure path; both plausible "heal the squatted block id" regressions) | Reddens |
|---|---|
| failed-write cleanup also **recursively clears the block path** — i.e. silently deleting the directory | `localDiskTier_failedWrite_releasesTheReservation`, `…emptyDirectorySquatting…`, `directoryPartial_…`, `localDiskTier_failedOverwrite_keepsThePreviousCopy` — **4 red** |
| same cleanup, children only: the directory survives and only the occupant dies | `localDiskTier_failedWrite_releasesTheReservation` — **1 red** |

The first lands on the *new* assertion rather than an earlier one — the three preceding assertions
(reservation released, no partial, put failed) all still pass, so the mutation isolates exactly what
`isDirectory(squat)` pins:

```
[nothing at the block path was removed]
Expecting value to be true but was false
    at CacheTierPartialWriteTest.localDiskTier_failedWrite_releasesTheReservation(…:204)
```

The empty half reddens with the identical description, which is the parity being demonstrated rather
than asserted. The second mutation exists because the first **shadows** the occupant assertion — with
the directory gone, `isDirectory` fails first and the occupant check never runs, so it would have been
reported as pinned while never having executed. Removing only the occupant reddens it on its own.

That also produced one improvement: the occupant check first failed as a raw `NoSuchFileException`
ERROR, which reads as "the test broke" rather than "the property was violated"; it is now
`assertThat(occupant).hasContent(…)`, which fails as a described assertion.

Re-verified at this head: **core 934, storage 301, 0 failures, 0 skipped, 0 stale**; core gate
`Running JBCT check on 62 Java file(s)` → `JBCT check passed.`, storage `57 Java file(s)` with only the
pre-existing `RemoteTierConfig.java:37` error. Every symbol and `file:line` cited in this PR body was
audited against the tree at the pushed SHA, with a deliberately fake control confirming the lookup can
fail: 0 problems.


siy:
**Adversarial review at `41c9bdaa7` — APPROVE-WITH-NITS.**

Both fixes are correct and the shipped fsync order is right. I independently reproduced the green
counts (**core 934, storage 301**, 0 failures, 0 skipped, `<testcase>` elements, marker-fresh, both
report dirs) and re-ran the author's three production mutations — all three redden exactly the
named tests. The `StorageFactory.createAll` fail-safe argument checks out: `admit` commits every
marker before the setups map exists, and both `PendingSetup` sites carry `armedDisk`. The
`moveReplace` caller table is complete, EXDEV is unreachable by construction, and the marker is
genuinely create-only with content read only for a diagnostic string.

Two things to fix before merge, neither a defect in shipped behaviour:

**1. The durability fix's two load-bearing properties are unpinned.** Both JFR observers read only
`event.getString("path")` and assert with an unordered `contains`. Two mutations I ran stay
**fully green (39/39 core + 2/2 storage)**:
- moving the parent-directory force **before** the write — the directory entry is then never
  synced after the marker is created, which is #1190's defect restored;
- `file.force(true)` → `file.force(false)` — `fdatasync` on a freshly created file need not make
  the inode durable.

`jdk.FileForce` already carries the fields that catch both — verified on Temurin 25:
`[startTime, duration, eventThread, stackTrace, path, metaData]`. Assert `metaData == true` for
both forces, and swap `contains(file, dir)` for `containsSubsequence(file, dir)` (the helper
already returns a chronologically ordered list). Two lines per observer, and both mutations redden.

**2. Phantom citation.** The PR body and `CacheTierPartialWriteTest.java:238-241` both justify
rewriting the empty-squatter test by parity with "the sibling non-empty-squatter test" — **there is
no such test in the repo.** Search space: every `*.java` excluding `/target/`, `grep -E`, with two
positive controls (`SquattingTheBlockPath` → exactly the 2 known tests; `[Nn]onEmpty` → 83 live
hits). The only squatter tests are the *delete* one and the rewritten *empty* one. The parity is
almost certainly true — `rename(file, non-empty-dir)` is refused under both moves — but it is
uncheckable as written. Either add the short non-empty-squatter put test or drop the claim. The
behaviour change itself looks right to me: failing loudly beats silently deleting a directory.

Nits: `forceParentDirectory` is private while `DefaultSnapshotManager` and `GitBackedPersistence`
both need exactly it (both still carry the unsynced-directory-entry `[unverified]` — fine to leave
to #1013/#1418, worth a line there); `readerRacingReplace`'s `torn` counter is structurally
unreachable in both arms; mangled indentation in the new `FileOps` try-blocks.

Not reached: power loss, Windows (the "fails rather than weakens" claim is untested in **both**
directions), and `LocalDiskTier`'s `.partial` is still unfsynced before the rename — so the replace
is atomic, not durable, and `readBlockDefault` applies no integrity check. All three are
out of scope and the author tagged the first and third.

siy:
**Retraction — finding 2 in my previous comment is WRONG. The sibling non-empty-squatter test does exist.**

It is `CacheTierPartialWriteTest.localDiskTier_failedWrite_releasesTheReservation` (`integrations/storage/src/test/java/org/pragmatica/storage/CacheTierPartialWriteTest.java:184` at `41c9bdaa7`). Its docstring reads *"a non-empty directory squatting on the block path makes the rename fail after the reservation"*, and its body does `Files.createDirectories(squat)` + `Files.writeString(squat.resolve("occupant"), "not a block")` before asserting the put fails. The parity claim in the PR body was true. I apologise for publishing the opposite.

**How the review got it wrong, since the failure is more interesting than the finding was.** Both of its absence controls passed while neither pattern could have reached the answer at any time:

| control | hits | why it could not reach it |
|---|---|---|
| `SquattingTheBlockPath` | 0 against the method name | the test is named by outcome — `failedWrite_releasesTheReservation` — and carries no "Squatting" token |
| `[Nn]onEmpty` | 0 | the prose spells it **hyphenated**, `non-empty` — 3 hits in that one file |

A positive control proves the *pattern* works. It cannot prove the *location* was right, and that is precisely the shape that fails while showing a green light. Re-searched on the token `squat` (case-insensitive) across all 3,921 `*.java` excluding `/target/`: 13 hits in this file alone.

**What is actually left, and it is much narrower:** the sibling pins *"the put fails"* but not *"the squatter is left in place"* — its only assertions are `usedBytes()` zero and no partial files. So the asserted parity holds for one half and is unpinned for the other. The fix is to **name the test** in both citations (a bare "sibling test" reference is unfalsifiable regardless of whether it resolves) and add `isDirectory(squat)` plus an occupant-survives assertion to the sibling, so both halves are real and checkable.

**Finding 1 (the unpinned durability properties) stands unchanged** — two mutations that restore #1190's defect leave the suite fully green, and that is being fixed alongside this.

Issue #1387, opened 2026-09-20
WAL replay after restart marks unacknowledged offsets visible: visible = min(durable, acked) does not survive a restart


siy:
Measured by rev1373 (round-2 verdict `oss/internal/sweep-state/s25-rev1373-r2.md` §5, probe P5 in `oss/internal/probes/s25-rev1373/round2/`) at PR #1373 head `e82926576` (rc4 tip `866c05b3b` + the #1352 tier-read bound). Pre-existing; not caused by #1373.

**Observation (WAL-backed owner, min-sync 2, offsets 0..4 published, 0..2 sealed, the peer acked through 2).** Before restart: `visible = 2`, `fetch(0) = [0,1,2]` — honest, 3 and 4 are durable locally but not yet held by the min-sync peer. After a restart (index rebuilt from refs, same WAL dir, no acks re-received): `visible = 4`, `durable = 4`, `sealedThrough = 2`, `fetch(0) = [0,1,2,3,4]`. Recovery seeds `visible = lastSealedOffset` (`seedHead`: `visibleOffset.set(base)`) and `placeRecord` replays the WAL tail with plain `append`, which is visible at once.

**Consequence.** A restart turns "durable on the owner, unacknowledged by min-sync" into "visible" for every replayed offset. Readers then see records whose publishers were told outcome-unknown (replication timeout) or are still awaiting acks, and which no replica holds — the visibility watermark's meaning (`min(durable, acked)`) does not survive a restart. On RF=1 the two coincide and nothing changes.

Related, noted in the same probe `[unverified: not induced]`: the no-WAL branch never seeds (`recoverPartition` returns when the WAL is absent), so a node with persistent segments and no WAL dir restarts with `visible = -1` and a ring at offset −1, re-assigning offsets from 0 over the sealed ones.

Wanted: replay restores `visible` to what the acked state supports — either persist the acked watermark alongside the checkpoint/segment index and seed `visible = min(replayed durable, persisted acked)`, or replay the tail as durable-but-not-visible and let the replication manager's re-acks advance it (the #1374 barrier shape). Pin: the P5 scenario as a test — after restart `fetch(0)` must return `[0,1,2]` until the peer re-acks.

siy:
Fixed by #1453, squash-merged into `release-1.0.0-rc4` as `24a0e0130` (CTO session 26).

[...]

Issue #1352, opened 2026-09-20
DROP_OLDEST eviction seals events past the visible position: an event the owner never acknowledged reaches the durable tier


siy:
Found by rebaseB while closing rev1309's N3 (session 25, 2026-09-20). Probe: `oss/internal/probes/s25-rebaseB/ProbeN3SealPastVisibleTest.java`; finding: `oss/internal/sweep-state/s25-n3-seal-past-visible.md`.

**Verified (probe 1/1):** with min-sync 2, a peer that never acks, retention `maxCount = 2` and three publishes, `OffHeapRingBuffer.notifyAndEvict → collectEvictedEvents` hands offset 0 to the eviction listener — in production the segment sealer — with `visible = -1`: `PROBE visible=-1 tail=1 head=2 sealed=[0]`. The hand-off reads `[tail, tail + count)` raw and consults neither `visibleOffset` nor `durableOffset`. An event the owner never acknowledged is sealed to the durable tier as if it were part of the log.

**Unverified — end-to-end exposure:** `TieredStreamReader.read` / `SegmentReader.readEvents` have no visible gate and `PartitionedStreamAccess.readWithSegmentFallback` serves the tier on `CursorExpired`; in the probe `readLocal(0)` returned `[]` because `readChecked` computes `head = min(rawHead, visible) = -1` before the `fromOffset < tail` check. The path that exposes it (not run): a replica-set change — a new peer catching up from `tail = 1` acks from 1, `visible` advances past 0, a read from 0 gets `CursorExpired` and the tier serves an offset nobody acknowledged. That is the acceptance test to induce.

**Ruled direction (CTO, know on PR #1346):** the seal hand-off clamps at `visibleOffset` — an evictee the owner never acknowledged is DROPPED and counted, never sealed, and the publisher's pending await for that offset resolves as a FAILURE promptly (via the pending-ack registry; otherwise the publisher waits out the 5 s replication timeout and reports outcome-unknown for an event definitively NOT in the log — the wrong #1236 class). Do not gate the tiered read instead: data on the durable tier that must not be served is a trap for every future reader. Replica arm clamps at its own durable prefix.

siy:
[...]

PR #1445, opened 2026-09-22
fix(streams): WAL replay restores visible = min(durable, acked) after a restart (#1387)

siy:
Closes #1387

A restart promoted "durable on the owner, unacknowledged by min-sync" to "visible". `visible = min(durable, acknowledged)` did not survive recovery: `OffHeapRingBuffer.seedHead` set `visibleOffset` to the durable sealed floor, and `placeRecord` replayed the WAL tail with the plain `append`, which has no durability gate and is visible at once. Readers then saw records whose publishers were told at most that the outcome is unknown, and which no replica is known to hold.

## The fix

Two halves, both load-bearing (the mutation table below shows they redden disjoint sets):

1. **Replay appends durable-but-not-visible.** New `OffHeapRingBuffer.appendDurable`, built on the existing `appendOrdered` — which already appends without advancing visibility and without waking a consumer. No new ring mechanism was invented for this.
2. **Recovery recomputes the watermark once**, from the live acknowledgement state, with the same expression the live path uses (`StreamPartitionManager.refreshVisible`). `advanceVisible` is monotonic, so the sealed floor that was visible before the restart is **never lowered** — a restart hides nothing a reader could see before it.

The new `AcknowledgedOffsetSource` seam exists for one specific reason: `minSyncReplicasFor` reads the `streams` map, and while `StreamEntry.fromConfig` recovers a partition the entry is not in that map yet. That lookup answers `0` — "no acknowledgement required" — which would have reinstated the defect on the createStream path while the lazy-materialize path was fixed. `minSyncReplicas` therefore comes from the config being materialized. Mutation M3 pins exactly that trap.

## Evidence

Branch cut from rc4 tip `29262fbb2` (confirmed with `git ls-remote`, not `origin/*`). All runs on the shared build host, module-scoped, foreground. Counts are `<testcase>` ELEMENTS from the copied-back XML, cross-checked against the `tests="N"` attribute sum — they agree in both modules.

**RED on the unmodified base** (`93bc8c22c` = rc4 tip + the pin + a test-wiring factory), reproducing the ticket's own signature on the CURRENT tip rather than on s25's `e82926576`:

```
[a replayed but unacknowledged offset is not served to a reader]
Expecting actual:
  [0L, 1L, 2L, 3L, 4L]
to contain exactly (and in same order):
  [0L, 1L, 2L]
```

**GREEN**

| Run | Result |
|---|---|
| `-pl aether/aether-stream install` | **1219** / 0 failures / 0 skipped |
| `-pl aether/node test` | **1634** / 0 failures / **1 skipped** |
| `jbct:check -Djbct.skip=false -pl aether/aether-stream` | `Running JBCT check on 101 Java file(s)` → `JBCT check passed.` |

The single node skip is `ObservabilityStep0BenchTest.step0_overhead_bench`, a pre-existing class-level `@Disabled` bench in a file this branch does not touch — identified by walking every `<testcase>` for a `<skipped>` child, not inferred from the summary line. `aether-stream` 1219 = 1216 on base + the 3 tests added here; nothing dropped. Both modules were re-run AFTER `jbct format`, so the green is measured on the formatted source.

**Mutation table** — each row is a FULL-module run (1220 tests), each hunk reverted by CONTENT (`git diff` reads zero for a staged change) with before/after occurrence counts printed:

| # | Hunk reverted | Reds | Named |
|---|---|---|---|
| M1 | `placeRecord` uses `ring.append` again | **1** | `restart_keepsUnacknowledgedOffsetsInvisible_whenThePeerHasNotReAcked` |
| M2 | recovery no longer calls `restoreVisibleWatermark` | **15** | `restart_makesTheReplayedTailVisibleAtOnce_whenNoPeerAckIsRequired` + **14 existing** recovery tests (`StreamPartitionManagerRecoveryTest` ×8, `StreamPartitionManagerRestartAfterCompactionTest` ×4, `StreamPartitionManagerFrozenRingDropTest` ×2) |
| M3 | the seam reads the `streams` map instead of the config | **1** | `restart_keepsUnacknowledgedOffsetsInvisible_whenThePeerHasNotReAcked` |

No hunk is unpinned, and M1/M2 redden disjoint sets. M2's 15 is the measurement worth reading: **hiding the replayed tail WITHOUT the recompute breaks 14 existing recovery tests**, so the recompute is half the fix, not a refinement of it.

## Limits, in the same namespace as the claims

- `[unverified: with minSyncReplicas >= 2 and a peer that never returns, a replayed tail stays durable-but-invisible indefinitely.]` Nothing re-replicates an unacknowledged tail on owner recovery — `replicateEvent`/`replicateEvents` are reached only from the publish path. This **restores** the pre-restart verdict rather than adding a new refusal; closing it is a separate change.
- `[unverified: a prefix SEALED while still unacknowledged still reads as visible across a restart.]` `seedHead`'s floor is the sealed watermark, and eviction under pressure seals ahead of the acknowledgements (#1352's `unacknowledgedEvictee_isSealed_andReadsAsNotYetVisible` pins that it can). Same defect class, smaller offset range; only persisting the acknowledged watermark closes it. Not folded in here — it wants its own review.
- **RF=1 cannot see any of this** and is not used as a control anywhere. `replicatedThrough` answers `Long.MAX_VALUE` for `minAcks <= 0`. The min-sync-1 test present here is the opposite guard: that the fix does not over-hide.
- `[unverified: the new test file is not covered by the module gate]` — `jbct.includeTests=false`, repo-wide and pre-existing. Checked separately with the local CLI: 0 format issues, 0 lint errors.

## Separate defect found while investigating — NOT fixed here

The ticket's `[unverified: not induced]` second finding **reproduces**:

```
PROBE NO-WAL sealedBefore=2 | after restart: rebuiltSealedThrough=2 head=-1 visible=-1 nextPublishedOffset=0
```

`recoverPartition` early-returns when a partition has no WAL, so `seedHead` never runs and a node with persistent SEALED segments but no WAL directory **re-assigns offsets from 0 over already-sealed data** — the rebuilt floor is correct and simply not consulted. Production reaches this only behind the explicit `aether.allowNonDurableStreams=true` opt-in (`AetherNode.decideWalAvailability` otherwise refuses to boot), but that opt-in's warning promises "non-crash-durable (in-memory tail)", which describes losing the un-sealed tail, not two records sharing one offset. Shown pre-existing by measurement: the identical reading appears under M1+M2 (base recovery behaviour). Reported to the CTO for its own ticket.


siy:
**Closing unmerged — superseded by #1453, which is the merged result of this branch and the orphaned session-25 implementation.**

This PR is green and `MERGEABLE`, which is exactly why it needed closing deliberately rather than being left open: merging it would have shipped a measured defect.

Its `restoreVisible` applies the OWNER rule unconditionally. A replica's visible position is its **own** durability, and a replica receives no acks — it sends them — so this implementation holds a replica's replayed tail **invisible indefinitely after a restart**. Measured, not argued: replacing the role branch in #1453 with exactly this rule reddens `walReplay_onAReplica_isVisibleAtOnce` (`Expecting actual: [] to contain exactly: ["r0","r1","r2"]`) across a full 1,225-test module run. This PR's own suite has no replica test and could not have found it.

The author of this branch found that defect in their own work by cross-checking against an orphaned, never-pushed implementation recovered from a local tree, and recommended merging the two rather than picking one. #1453 carries the role-aware `restoreVisible` and its three wiring-path tests from that implementation, plus this branch's real-sealer / real-`SegmentIndex.rebuildFromRefs` / real-`fetch(0)` pin — which is the ticket's named acceptance test and which the other implementation never exercised. Neither branch alone was adequate.

`Closes #1387` moves to #1453. Nothing here is lost; the pre-merge state of the orphaned side is preserved at `archive/1387-s25-orphan-preimage`.

Issue #1184, opened 2026-09-14
[infra] version conflicts are swallowed: loadInfraIntoShared discards CompatibilityResult, first version wins permanently


siy:
`SharedDependencyLoader.loadInfraIntoShared` discards the `CompatibilityResult`, so an `[infra]` version conflict is indistinguishable from a clean no-op:

```java
return sharedLibraryLoader.checkCompatibility(dependency)
                          .fold(() -> loadInfraArtifact(...), _ -> logInfraAlreadyLoaded(dependency));
```

The `_ ->` swallows the result: **`Compatible` and `Conflict` both become a DEBUG "already loaded" no-op.** The runtime computes the correct diagnosis one layer up and then throws it away.

**`[infra]` has no equivalent of `[shared]`'s fallback.** `[shared]` routes the same condition through `handleConflictingDependency` → `loadConflictIntoSlice`, giving the conflicting slice its own loader. `[infra]` has nothing. And `SharedLibraryClassLoader.addArtifact` is keyed on `groupId:artifactId` with a `loadedArtifacts.containsKey` guard that **ignores the version outright**, so the first version to load wins permanently, for every later slice, for the lifetime of the process.

**Driven through the real production path, not simulated** — `SharedDependencyLoader.processInfraDependencies` verbatim (probe `Probe758cTest.p5_infraVersionSkew_throughTheRealSharedDependencyLoader`). Slice A declares `[infra] org.example:lib:1.0.0`, slice B declares `2.0.0`, and `NeverExisted` ships only in 2.0.0:

```
P5 after slice-A [infra] 1.0.0 => loaded={org.example:lib=Version[major=1, minor=0, patch=0]}
P5 compatibility of slice-B's 2.0.0 request => Some(Conflict[loadedVersion=1.0.0, required=Exact[2.0.0]])
P5 after slice-B [infra] 2.0.0 => loaded={org.example:lib=1.0.0} ; shared loader URLs=[.../infra-1.0.0....jar]
```

The conflict is detected, reported as `Conflict`, logged at DEBUG as "already loaded", and slice B silently runs against 1.0.0.

**Why this matters beyond the silent downgrade.** It is the *producer* of a second defect: because a version-skewed slice reaches the classloader with its classes missing, the class-load diagnostic (PR #1131, ticket #758) sees "package served, class absent" and reports

> `slice was compiled against an older runtime (references removed class X …); rebuild against this runtime version`

[...]

Issue #1344, opened 2026-09-19
A crashed surefire fork can leave Maven exiting 0: later test classes never run and the build reports success


siy:
Observed by s1234b while probing #1297 (probe P18, which reintroduces the sealer drain recursion). The recursion kills the test JVM inside `SegmentSealerBacklogTest`. **Five later classes never run, only 12 testcases are recorded, and Maven still exits 0.**

So any defect that crashes a fork — StackOverflowError, OOM, a JVM-level error — can read as a green build with a smaller test count, and nothing in the exit status says otherwise. That is the same class as every other false green recorded in CLAUDE.md, but at the harness level: it can hide arbitrarily many unrun tests on CI as well as locally.

Required:
- make a crashed fork FAIL the build (surefire `failIfNoSpecifiedTests` is not this; check `forkedProcessExitTimeoutInSeconds` and the "The forked VM terminated without properly saying goodbye" path, and whether a configuration suppresses it today);
- assert an expected minimum test count per module, or compare the executed class count against the discovered one, so a silent shortfall fails rather than passes;
- reproduce with s1234b's P18 and pin it.

[unverified: whether CI's configuration already fails on a crashed fork — measure before concluding the exposure is local-only.]

---
**Premise correction (s1344, 2026-09-20, PR #1348):** surefire DOES fail a fork that dies — `Runtime.halt(137)` mid-class gives "The forked VM terminated without properly saying goodbye" (rc=1), and a `StackOverflowError` on the test thread is reported by JUnit as an ordinary error (rc=1). P18's false green is a third path, read from the fork's own `.dump`: the recursion bottomed out inside `java.lang.StringBuffer.<clinit>`, poisoning the class for the JVM's lifetime; when the test's await later failed, surefire's listener threw `NoClassDefFoundError` while writing the failure, JUnit's attempt to log that threw the same error, the Throwable escaped the launcher, and `ForkedBooter.execute()` caught it, could not report it, and said goodbye from its `finally` — BYE, `System.exit(0)`, BUILD SUCCESS. Reproduced 2 of 3 runs on `0a244edc3` (1 of 3 fails honestly); no surefire setting covers "goodbye before the plan finished" (3.5.1 and 3.5.4 identical there). **CI is exposed too, measured:** a commit carrying only a deterministic repro of that path ran green on CI (run 35481807877) with 83 of aether-stream's 290 test classes started.

siy:
[...]

PR #1454, opened 2026-09-22
fix: decide the worker's failure disposition per phase, load PERMANENT and activation RETRY (#1184)

siy:
Closes #1184

Recovered work. Written by CTO session 25 and committed locally at 2026-09-22 **01:06:08 +0200** —
26 seconds after the commit recovered in #1443, in the same crash. It was never pushed, so PR #1416
merged (squash `9324ace76`) without it. Rebased onto `8f6045cf1`; no conflicts, and the rebased
patch is byte-identical to the original `707cf3b8c` (`git show` of both, `diff` → 0 lines).

It is **rev1416's R2-BLOCKING fix**, in the shape that review prescribed.

## Read this first: the defect is REAL, but it is NOT REACHABLE on rc4

The original framing of this recovery — mine — was that rc4 is actively rolling blueprints back on
transient worker failures. **That is wrong, and I am correcting it here rather than letting it
stand.** `WorkerDeploymentManager` is dead code on rc4. Two independent traces at `8f6045cf1`:

1. **Nobody holds the object.** Its only production mention outside its own file is
   `AetherNode.java:5800`, where it is constructed. The local `workerDeploymentManager` occurs
   **exactly once** in all of `AetherNode.java` — the assignment itself; it is never read. (Contrast
   `governorAnnouncer`, constructed twelve lines below and then `.start()`ed.) The factory returns a
   plain record and there is no `subscribe`/`register`/`addListener`/`start` anywhere in the file, nor
   any `META-INF`/ServiceLoader wiring.
2. **Nobody calls the entry point.** `onDirectivePut` — the class's only inbound method — has 1
   declaration, 1 implementation, and **5 call sites, all inside `WorkerDeploymentManagerTest`**.
   Zero production callers. Positive control: the live callback `onSubscriptionPut` *is* wired into
   `AetherNode`.

This independently reproduces RULING 2 of `s26-1390-split-plan.md` §10.0 ("an object nobody holds,
which registers nothing, cannot be invoked — the worker path was never live on rc4").

**Why the alarm was raised anyway, stated so the shape is recognisable:** rev1416 *measured* the
defect with a mock-driven unit test that constructs the manager directly. That proves the class
misbehaves; it cannot prove production reaches the class. I traced the defect to what the code
**says** and to a test that **bypasses the reachability question**, never to what **invokes** it.

So: **no operator is affected today, and no incident is in progress.** What this PR fixes is
incorrect source that ships, not running behaviour.

## What it changes

rc4 forwards ONE disposition for BOTH phases — `Unrecognised.PERMANENT` at `forwardSliceFailure` —
so an untyped *activation* failure (a resource provider that cannot reach a database for a second)
would be condemned as fatal. The FSM path decides per phase: load → `PERMANENT`
(`NodeDeploymentState.handleLoadingFailure`), activation → `RETRY` (the site whose own comment names
itself "the site that closes #923").

This commit moves the decision to the phase boundary, which is rev1416's prescription verbatim:
`loadSlice` is mapped through `classifyLoadFailure` (→ `PERMANENT`) **before** the activation legs,
and the chain's terminal `.onFailure` forwards `RETRY`. A cause typed at its raise site classifies
identically either way, so #1184's `SharedLoaderVersionConflict` case is unchanged.

## Evidence

All runs `-pl aether/node` on the shared build host, foreground. Counts are `<testcase>` **elements**
from BOTH `surefire-reports` and `failsafe-reports`, cross-checked against the `tests="N"` attribute
sum.

| Run | Elements | attr sum | Files | fail / err / skip |
|---|---|---|---|---|
| Baseline, unmodified `8f6045cf1` | **1637** | 1637 | 432 | 0 / 0 / **1** |
| GREEN, this branch `7c3246858` | **1639** | 1639 | 432 | 0 / 0 / **1** |

Element count and attribute sum agree in both. +2 is exactly the two new tests
(`WorkerDeploymentManagerTest` 5 → 7). The count rose across all five runs and never dropped, so no
run was truncated by a crashed fork (#1344). **The single skip is named**: it is the pre-existing
manual benchmark `ObservabilityStep0BenchTest.step0_overhead_bench` (`<skipped message="manual bench
— run explicitly; prints ns/op to stdout"/>`), in a file this branch does not touch. Zero report XML
predated either run's start marker.

**Gate:** `Running JBCT check on 159 Java file(s)` → 0 format issues, 0 lint errors, 1698 warnings
(non-fatal), `JBCT check passed.` N = 159 > 0.

### RED on the unmodified base

Rather than adapt anything, I applied **only the test half** of the commit to `8f6045cf1`, leaving
production untouched (verified by content: `classifyLoadFailure` = 0 occurrences, disposition still
`PERMANENT`). Result: **7 run, 1 failure.**

```
WorkerDeploymentManagerTest.untypedActivationFailure_isForwardedAsRetryable:249
  [an untyped activation failure is retryable]
  Expecting value to be false but was true
```

`fatal()` is `true` on rc4 where it must be `false` — R2-BLOCKING's exact signature, using rev1416's
own probe cause (`Causes.cause("db unreachable: connection refused")`). The other six pass,
including the new `typedFatalActivationFailure_isForwardedAsFatal`, which is a control rather than a
red: a typed Fatal is fatal on both sides.

### Mutation table — each hunk reverted BY CONTENT, module-full (1639) each

Verified landed by grep count **before** each run, restored by `git checkout --`, restore verified by
returning the marker count to baseline plus empty porcelain. Never by `git diff` (which reads zero
for a staged change) and never by exit status.

| # | Hunk reverted | marker | Red | Named test + signature |
|---|---|---|---|---|
| N1 | `.mapError(workerDeploymentManager::classifyLoadFailure)` removed | 1→0 | **1** | `untypedLoadFailure_isForwardedAsPermanent` — `Expecting value to be true but was false` |
| N2 | `Unrecognised.RETRY` → `PERMANENT` (i.e. back to rc4) | 1→0 | **1** | `untypedActivationFailure_isForwardedAsRetryable` — `Expecting value to be false but was true` |
| N3 | **control liveness**: strip the cause's type before forwarding | 1→0 | **3** | `typedFatalActivationFailure_isForwardedAsFatal`, `loadFailure_isForwardedWithItsReasonAndFatalFlag` (`a version conflict is permanent`), `untypedLoadFailure_isForwardedAsPermanent` |

**The two production hunks are INDEPENDENTLY pinned — stated as a claim about what goes RED, not
about what stays green: N1 and N2 redden DIFFERENT named tests, neither set a superset of the
other.** That is the phase split working in both directions, which is the whole point: one
disposition for two phases is wrong in both.

N3 is not a hunk revert. It exists because `typedFatalActivationFailure_isForwardedAsFatal` is green
on base *and* on the fix, so nothing so far proved it was live rather than vacuous. Stripping the
cause's type reddens it — so it genuinely pins that a typed cause passes through by its own type.

All three of rev1416's prescribed pins therefore have a named red: untyped `activateSlice` →
`fatal=false` (N2), untyped `loadSlice` → `fatal=true` (N1), typed Fatal from either phase →
`fatal=true` (N3).

## This ENLARGES the #1390 deletion precondition — from three cases to five

`WorkerDeploymentManager` and its test are deleted by #1390. `s26-1390-split-plan.md` §10.4 already
records that the delete must not land until a pin exists over whichever path the worker then takes,
because "the witness died with the subject" — a modify/delete conflict discards the test along with
the class, and after it lands nothing will ever report the absence.

**That recorded precondition names three cases. This PR adds two more.** Whoever executes the
deletion must port all five, not the three currently written down:

| # | Case | Expected | Source |
|---|---|---|---|
| 1 | typed permanent (version conflict) | `fatal=true` + reason | `9324ace76` (already recorded) |
| 2 | intermittent | `fatal=false` + reason | `9324ace76` (already recorded) |
| 3 | untyped **load** | `fatal=true` (`PERMANENT`) | `9324ace76` (already recorded) |
| 4 | **untyped activation** | **`fatal=false` (`RETRY`)** | **this PR — NEW** |
| 5 | **typed Fatal from activation** | **`fatal=true`** | **this PR — NEW** |

Cases 4 and 5 are the phase split. A port that carries only 1–3 reinstates exactly the defect
rev1416 raised, and nothing would go red.

**Consequence worth weighing before merging:** this PR modifies a file #1390 deletes, so it
guarantees a modify/delete conflict there, and its own tests die with the class. Its durable value
is the specification above, not the code. That is a real trade-off and it is the maintainer's call,
not mine — the evidence is complete either way.

## Limits

- `[unverified: no cluster or Forge run.]` The behaviour is not reachable in production on rc4, so
  there is no end-to-end path to drive; everything here is module-scope.
- `[unverified: whether #1390's replacement worker runtime reaches `NodeDeploymentState`'s FSM path.]`
  Carried forward from `s26-1390-split-plan.md` §10.3, which flags the same trace as not done.
- `[unverified: the test changes are not gated.]` `jbct.includeTests=false` repo-wide; the 159 files
  the gate examined are `src/main/java` only. Pre-existing.
- rev1416 recorded that the worker's activation cleanup gap (the FSM unpublishes routes and
  subscriptions on activation failure; the worker does not) is **pre-existing and not addressed here**.


siy:
**Closing unmerged. The work is correct; merging it would be negative value, and that is a judgement about the target, not about the change.**

All three files this PR touches — `WorkerDeploymentManager.java`, its test, and the changelog fragment — are **deleted by PR #1390**, which the owner has ruled lands. So merging this would:

- add code to a class with **zero production callers**, verified two independent ways: the local `workerDeploymentManager` occurs exactly once in all of `AetherNode.java` (the assignment, never read — control: `governorAnnouncer` twelve lines below is constructed *and* `.start()`ed), and `onDirectivePut` has five call sites, all inside its own test (control: the live `onSubscriptionPut` *is* wired into `AetherNode`);
- guarantee a modify/delete conflict against #1390;
- ship tests that die with the class they pin.

**#1184 is correctly CLOSED.** Its substantive fix — `SharedDependencyLoader` +233, `SharedLibraryClassLoader` +58, `DependencyResolver`, `SliceLoadingFailure`, plus two real test classes — merged as `9324ace76` and is reachable. Only the worker-path hunk was missing, and that path does not execute.

**Why this PR existed at all, recorded because the lesson outlives it:** rev1416 measured the defect with **mocks**, which construct the manager directly. That proves the class misbehaves; it says nothing about production reaching it. **A mock is a holder production does not supply.** The finding was traced to what the code *says* and to a test that *bypasses* reachability — never to what *invokes* it. I compounded it by reopening #1184 on that basis without checking reachability myself, and have since re-closed it with the correction recorded there.

**Nothing is lost.** The branch is preserved and pushed; the five-case failure-disposition table in this PR's body is the durable artefact, not the code. If any of those cases describe behaviour the **live** FSM path (`NodeDeploymentState.handleLoadingFailure`) ought to have and does not, that is a real ticket against reachable code — being checked now, and it will be filed on its own merits rather than smuggled in on a dead class.

